PR #3, opened 2020-09-29
Functionality to take Prometheus version (major.minor only) config params and find latest patch

justinmclark:
Based on our conversion earlier, here is the function that will search for the latest patch for the associated <major> and <minor> versions.

### DIFF
--- a/config.yaml
+++ b/config.yaml
@@ -1,18 +1,21 @@
 options:
-  prometheus-image-path:
+  prometheus-image-major-version:
     type: string
     description: |
-      The location of the image to use,
-      e.g. "registry.example.com/prometheus:v1".
-
-      This setting is required.
-    default: "prom/prometheus:latest"
+      The major version of the Prometheus image to use.
+      Patch versions will automatically be set to the latest version.
+    default: "2"
+  prometheus-image-minor-version:
+    type: string
+    description: |
+      The minor version of the Prometheus image to use.
+      Patch versions will automatically be set to the latest version.
+    default: "20"
   prometheus-image-username:
     type: string
     description: |
       The username for accessing the registry specified in
       prometheus-image-path.
-    default: ""
   prometheus-image-password:
     type: string
     description: |
--- a/config.yaml
+++ b/config.yaml
@@ -1,38 +1,11 @@
 options:
-  image-tags-url:
+  prometheus-image-version:
     type: string
     description: |
-      URL at which all the tags for the Prometheus container images can
-      be viewed and verified.
-    default: "https://registry.hub.docker.com/v1/repositories/prom/prometheus/tags"
-  image-registry-path:
-    type: string
-    description: |
-      Path from which to pull the Prometheus container image.
-    default: "prom/prometheus"
-  prometheus-image-major-version:
-    type: string
-    description: |
-      The major version of the Prometheus image to use.
+      The version of the Prometheus image to use. This should be specified
+      in the format {major}.{minor}.
       Patch versions will automatically be set to the latest version.
-    default: "2"
-  prometheus-image-minor-version:
-    type: string
-    description: |
-      The minor version of the Prometheus image to use.
-      Patch versions will automatically be set to the latest version.
-    default: "20"
-  prometheus-image-username:
-    type: string
-    description: |
-      The username for accessing the registry specified in
-      prometheus-image-path.
-  prometheus-image-password:
-    type: string
-    description: |
-      The password associated with prometheus-image-username for
-      accessing the registry specified in prometheus-image-path.
-    default: ""
+    default: "2.20"
   advertised-port:
     description: The port prometheus will be listening on
     type: int
--- a/config.yaml
+++ b/config.yaml
@@ -1,4 +1,15 @@
 options:
+  image-tags-url:
+    type: string
+    description: |
+      URL at which all the tags for the Prometheus container images can
+      be viewed and verified.
+    default: "https://registry.hub.docker.com/v1/repositories/prom/prometheus/tags"
+  image-registry-path:
+    type: string
+    description: |
+      Path from which to pull the Prometheus container image.
+    default: "prom/prometheus"
   prometheus-image-major-version:
     type: string
     description: |
--- a/config.yaml
+++ b/config.yaml
@@ -1,5 +1,5 @@
 options:
-  prometheus-image-version:
+  prometheus-version:
     type: string
     description: |
       The version of the Prometheus image to use. This should be specified
--- a/src/charm.py
+++ b/src/charm.py
@@ -311,17 +311,18 @@ class PrometheusCharm(CharmBase):
         """
         # 0. check if version has already been stored in self.store
         # 1. get valid tags from URL (or directly from input)
-        # 2. clean tags data structure and filter out release candidates "rc"
-        # 3. filter out tags that don't match our major/minor specification
+        # 2. filter out tags that don't match our major/minor specification
+        # 3. sort by version using 'packaging' library and return latest
         if self.stored.prom_version:
+            logger.debug('Using cached Prometheus version: {}'.format(self.stored.prom_version))
             return self.stored.prom_version
 
         config = self.model.config
         if not api_tags:
             api_tags_resp = requests.get(IMAGE_TAGS_URL)
             if not api_tags_resp.ok:
-                logger.error('No Prometheus image tag can be selected. Cannot access '
-                             '{} (defined in config.yaml).'.format(IMAGE_TAGS_URL))
+                logger.error('No Prometheus image tag can be selected. '
+                             'Cannot access {}.'.format(IMAGE_TAGS_URL))
                 self.unit.status = BlockedStatus('Cannot get container image path')
                 return ''
             api_tags = json.loads(api_tags_resp.text)
@@ -350,9 +351,10 @@ class PrometheusCharm(CharmBase):
             self.unit.status = BlockedStatus('Cannot verify prom v{0}.{1}'.format(major, minor))
             return ''
         else:
-            # find the latest tag by sorting by the parsed version object and get first element
+            # find the latest tag by sorting by the parsed version objects
             _, latest_tag = sorted(tags, reverse=True)[0]
-            self.stored.prom_version = latest_tag  # cache this version
+            self.stored.prom_version = latest_tag  # cache this
+            logger.debug('Using Prometheus version: {}'.format(latest_tag))
             return latest_tag
 
     def _build_pod_spec(self):
--- a/src/charm.py
+++ b/src/charm.py
@@ -5,6 +5,7 @@
 import logging
 import json
 import requests
+import semver
 import yaml
 
 from ops.charm import CharmBase
@@ -13,6 +14,8 @@ from ops.main import main
 from ops.model import ActiveStatus, MaintenanceStatus, BlockedStatus
 
 logger = logging.getLogger(__name__)
+IMAGE_REGISTRY_PATH = 'prom/prometheus'
+IMAGE_TAGS_URL = 'https://registry.hub.docker.com/v1/repositories/prom/prometheus/tags'
 
 
 class PrometheusCharm(CharmBase):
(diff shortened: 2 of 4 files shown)


config.yaml
```diff
@@ -1,18 +1,32 @@
...
-      This setting is required.
-    default: "prom/prometheus:latest"
+      URL at which all the tags for the Prometheus container images can
+      be viewed and verified.
+    default: "https://registry.hub.docker.com/v1/repositories/prom/prometheus/tags"
+  image-registry-path:
+    type: string
+    description: |
+      Path from which to pull the Prometheus container image.
+    default: "prom/prometheus"
```
jameinel, line 12: Doesn't registry-path overlap with image-tags-url?

config.yaml
```diff
@@ -1,18 +1,32 @@
...
+    description: |
+      The major version of the Prometheus image to use.
+      Patch versions will automatically be set to the latest version.
+    default: "2"
+  prometheus-image-minor-version:
+    type: string
+    description: |
+      The minor version of the Prometheus image to use.
+      Patch versions will automatically be set to the latest version.
+    default: "20"
```
jameinel, line 24: This feels really clumsy. Is there a reason to not just have "prometheus-image-version" and document that it is a Major.Minor version?

src/charm.py
```diff
@@ -301,16 +302,60 @@ def _prometheus_config(self):
...
 
+    def _get_image_tag(self, api_tags=None) -> str:
+        """Get most recent patch of major and minor Prometheus version
+        """
+        # 1. get valid tags from URL (or directly from input)
+        # 2. clean tags data structure and filter out release candidates "rc"
+        # 3. filter out tags that don't match our major/minor specification
+        config = self.model.config
+        if not api_tags:
+            api_tags_resp = requests.get(config['image-tags-url'])
```
jameinel, line 313: The concepts described here feel like something we want to encapsulate in a reusable Component, since we'll want the same behavior for Grafana, and Mattermost, and all the other things that use images.

src/charm.py
```diff
@@ -301,16 +302,60 @@ def _prometheus_config(self):
...
+        # if there is nothing in tags, we haven't found a valid set of tags
+        if not tags:
+            logger.error('Prometheus version cannot be verified: v{0}.{1}'.format(major, minor))
+            self.unit.status = BlockedStatus('Cannot verify prom v{0}.{1}'.format(major, minor))
+            return ''
+
+        # select the latest patch version for v{major}.{minor}
+        patched_tags = [tag for tag in tags if len(tag) == 3]
+        unpatched_tags = [tag for tag in tags if len(tag) == 2]
+        sorted_tags = sorted(patched_tags, key=lambda tag: tag[2], reverse=True)
```
jameinel, line 336: Doesn't this sort based on string and not based on int?
So it would do weird things with 2.2.1 vs 2.2.10 vs 2.2.2
And we also have to watch out for tags like 7.2.3-ubuntu, right?
Though I guess that is Grafana tag structure:
https://hub.docker.com/r/grafana/grafana/tags

not Prometheus.
But I do worry about str vs int sorting. (prometheus definitely has 2.20, but I don't know if they hit 2.20.10)

tests/test_charm.py
```diff
@@ -53,7 +59,9 @@ def test_password_is_required_when_username_is_set(self):
         expected = ['prometheus-image-password']
         self.assertEqual(missing, expected)
 
-    def test_alerting_config_is_updated_by_alertmanager_relation(self):
+    @patch('charm.PrometheusCharm._get_image_tag')
+    def test_alerting_config_is_updated_by_alertmanager_relation(self, _get_image_tag_mock):
+        _get_image_tag_mock.return_value = 'v2.20.1'
```
jameinel, line 64: This is definitely a sign that we are layering things in a poor way, given that this ends up impacting every test.
Is there a way to split out how this is done, so it doesn't affect the core logic?

tests/test_charm.py
```diff
@@ -322,6 +372,33 @@ def test_k8s_scrape_config_can_be_set(self):
...
+        api_tags = [{'layer': '', 'name': 'v2.20'}, {'layer': '', 'name': 'v2.21.0'}, {'layer': '', 'name': 'v2.21.0-rc.0'}, {'layer': '', 'name': 'v2.21.0-rc.1'}, {'layer': '', 'name': 'v2.3.0'}, {'layer': '', 'name': 'v2.3.1'}, {'layer': '', 'name': 'v2.3.2'}, {'layer': '', 'name': 'v2.4.0'}] # noqa
+        actual_tag = self.harness.charm._get_image_tag(api_tags=api_tags)
+        expected_tag = 'v2.20'
+        self.assertEqual(actual_tag, expected_tag)
+
+    def test_no_valid_tags(self):
+        self.harness.set_leader(True)
+        k8s_config = MINIMAL_CONFIG.copy()
+        self.harness.update_config(k8s_config)
+        api_tags = [{'layer': '', 'name': 'latest'}, {'layer': '', 'name': 'v2.2.0-rc.0'}, {'layer': '', 'name': 'v2.2.0-rc.1'}, {'layer': '', 'name': 'v2.2.1'}, {'layer': '', 'name': 'v2.20.0-rc.0'}, {'layer': '', 'name': 'v2.20.0-rc.1'}, {'layer': '', 'name': 'v2.21.0'}, {'layer': '', 'name': 'v2.21.0-rc.0'}, {'layer': '', 'name': 'v2.21.0-rc.1'}, {'layer': '', 'name': 'v2.3.0'}, {'layer': '', 'name': 'v2.3.1'}, {'layer': '', 'name': 'v2.3.2'}, {'layer': '', 'name': 'v2.4.0'}] # noqa
```
jameinel, line 397: These feel like they would be easier to read if you did them as:

```
api_tags = [
  {'layer': '', 'name': 'latest'}, 
  {'layer': '', 'name': 'v2.2.0-rc.0'},
  {'layer': '', 'name': 'v2.2.0-rc.1'},
  {'layer': '', 'name': 'v2.2.1'},
  {'layer': '', 'name': 'v2.20.0-rc.0'},
  {'layer': '', 'name': 'v2.20.0-rc.1'},
  {'layer': '', 'name': 'v2.21.0'},
...]
```

And then you can drop the #noqa as well.

config.yaml
justinmclark: Agreed.

src/charm.py
justinmclark: Agreed. I was thinking about how to do that - but it seems like every repo has it's own tagging structure which would make picking the latest patch for any repo pretty complicated.

Maybe there is a way to do this, but I think it would have to happen when we have ROCKs. I could be wrong and happy for suggestions.

src/charm.py
justinmclark: It does sort based on string, but since we do `tag.split('.')` the only string comparisons (at least for Prometheus) will be stuff like '2' vs '20' where the string '20' would be larger than '2'.

Like I mentioned in the above comment, I think this logic will only work for Prometheus tags (and repos with similar tag structures). Right now, I'm not sure how we can generalize this. I just wanted to make sure we had something working.

tests/test_charm.py
justinmclark: Again, I agree and didn't particularly like this. However, since `pod_spec_set()` requires an image path (which requires a valid tag) and the charm ends up calling it in these other tests when most (maybe all?) hooks are triggered, I wasn't sure of how to do this without completely overhauling the charm itself.

@balbirthomas , do you have thoughts on this?

tests/test_charm.py
justinmclark: Good call. I'll make the change.

src/charm.py
jameinel: I think you're missing how this works. If you have 2.20.1, 2.20.2, ... 2.20.9, 2.20.10, ,2.20.11
Then string sorting will put 2.20.2 "higher" than 2.20.11

config.yaml
justinmclark: I'm not sure on this one but I don't think there's much overlap. I'll have to do some more digging.

src/charm.py
justinmclark: Brain fart. Let me correct this.

config.yaml
jlounder: The goal here is to completely remove references to registry path, not split it into multiple configs. At this time, there's no reason to allow for a consumer to override the image location. Eventually, we'll be leveraging resources as the correct path to override. Let's not add config items which will be removed in the near future.

config.yaml
```diff
@@ -1,18 +1,32 @@
...
   prometheus-image-username:
```
jlounder, line 25: Similar to image-path and registry-path, the image we need is not protected, and will not need a username/password. Let's remove these configs now, to prevent future breaking changes.

src/charm.py
jlounder: Might be worth leveraging https://pypi.org/project/semver/ for this logic.

justinmclark: Oh that's super helpful @jlounder . I wasn't aware of that.

src/charm.py
```diff
@@ -301,18 +305,70 @@ def _prometheus_config(self):
...
+            logger.error('Prometheus version must be in the form {major}.{minor}')
+            self.unit.status = BlockedStatus('Incorrect Prometheus version format')
+            return ''
+
+        # semver doesn't like 'v' at the beginning of the version string
+        # so we'll create a map of invalid semver strings to valid semver strings
+        tags_map = {tag['name']: tag['name'][1::] if tag['name'][0] == 'v' else tag['name']
+                    for tag in api_tags}
+
+        # get list of tuples of the structure [ (original_tag, parsed_version_object), ... ]
```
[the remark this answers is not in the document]
justinmclark: I save the original tag because that's what is ultimately returned from the function and parsed_version is used for sorting.

tests/test_charm.py
```diff
@@ -322,6 +355,48 @@ def test_k8s_scrape_config_can_be_set(self):
         k8s_pod_scrape_config = scrape_config(pod_spec, 'kubernetes-pods')
         self.assertIsNotNone(k8s_pod_scrape_config, 'No k8s pods scrape config found')
 
+    def test_image_tags_with_patch_versions(self):
+        self.harness.set_leader(True)
+        k8s_config = MINIMAL_CONFIG.copy()
+        self.harness.update_config(k8s_config)
```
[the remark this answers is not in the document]
justinmclark: Config is needed because that's where the major.minor version is stored. `set_leader` is not required. I'll remove.

src/charm.py
```diff
@@ -301,18 +305,70 @@ def _prometheus_config(self):
...
+        try:
+            major, minor = config['prometheus-image-version'].split('.')
+        except ValueError:
+            logger.error('Prometheus version must be in the form {major}.{minor}')
+            self.unit.status = BlockedStatus('Incorrect Prometheus version format')
+            return ''
+
+        # semver doesn't like 'v' at the beginning of the version string
+        # so we'll create a map of invalid semver strings to valid semver strings
+        tags_map = {tag['name']: tag['name'][1::] if tag['name'][0] == 'v' else tag['name']
```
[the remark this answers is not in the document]
justinmclark: Fair enough. 1: is cleaner.

jameinel:
Speaking to this:
> Use StoredState to persist the version string after determining it once early in the life cycle.

It brought up an idea. I actually think we *do* want to persist the version that we have selected.

Here is the rationale. If you do "juju deploy prometheus --config prometheus-version=2.20", and it decides to be 2.20.11, you don't want the action of "juju relate prometheus grafana" to suddenly upgrade you to 2.20.12 because that happened to be released in the mean time.
I think StoredState is a very good use case here, and having a mechanism by which we can trigger the charm to reevaluate the version that it is using. (Probably needs to be a 'juju run-action prometheus update-patch-version' or something like that)

justinmclark:
@jameinel I see what you're saying, but my understanding of the motivation for this change is that we always want the latest patch.

jameinel:
It still shouldn't be done as a side effect of something else, IMO. It
leads to hard to debug issues (I thought the relation data was wrong, but
actually it was the application version getting updated.)

src/charm.py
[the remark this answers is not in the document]
balbirthomas: I see the problem now. The problem is the inconsistent use of the "v" prefix by Prometheus docker images. If they hand not done this it would have been easy to just reassemble the original version tag from the parsed version tag.

src/charm.py
```diff
@@ -301,18 +305,70 @@ def _prometheus_config(self):
...
+            try:
+                parsed_version = semver.VersionInfo.parse(tags_map[tag['name']])
+            except ValueError:
+                logger.warning('Error parsing possible version tag: {}'.format(tag['name']))
+                continue
+            tags.append((tag['name'], parsed_version))
+
+        # filter out only those tags with the specified major and minor versions
+        tags = [(tag, parsed_version) for tag, parsed_version in tags
+                if str(parsed_version.major) == major and str(parsed_version.minor) == minor]
```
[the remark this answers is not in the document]
justinmclark: You're totally right. I had the previous loop as a list comprehension previously but now that it's a loop this will be cleaner. Good call.

justinmclark:
The charm now caches the initial patch version. However, we still need to add an action to upgrade the patch version because right now it's locked in at whatever version is first decided. That'll be for a separate PR, though.

justinmclark:
Closing because we are moving to resources.

PR #4, opened 2020-10-13
Fix blocked state

dstathis:
Two minor fixes here. If we are setting blocked status, we need to make sure we return so it doesn't get changed back. Also, the status was set twice.


balbirthomas (approved):
Thanks.

justinmclark (approved):

PR #5, opened 2020-10-16
Alertmanager support

dstathis:
Implements the alertmanager interface

### DIFF
--- a/src/charm.py
+++ b/src/charm.py
@@ -17,44 +17,47 @@ logger = logging.getLogger(__name__)
 class PrometheusCharm(CharmBase):
     """A Juju Charm for Prometheus
     """
-    stored = StoredState()
+    _stored = StoredState()
 
     def __init__(self, *args):
         logger.debug('Initializing Charm')
 
         super().__init__(*args)
+
         self.stored.set_default(alertmanagers=[])
         self.stored.set_default(alertmanager_port='9093')
 
         self.framework.observe(self.on.config_changed, self._on_config_changed)
         self.framework.observe(self.on.stop, self._on_stop)
-        self.framework.observe(self.on['alerting'].relation_changed,
-                               self.on_alerting_changed)
-        self.framework.observe(self.on['alerting'].relation_departed,
-                               self.on_alerting_departed)
-        self.framework.observe(self.on['alerting'].relation_broken,
-                               self.on_alerting_broken)
+        self.framework.observe(self.on['alertmanager'].relation_changed,
+                               self._on_alertmanager_changed)
+        self.framework.observe(self.on['alertmanager'].relation_departed,
+                               self._on_alertmanager_departed)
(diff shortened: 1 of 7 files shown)


src/charm.py
```diff
@@ -257,8 +257,14 @@ def _alerting_config(self):
         if len(self.stored.alertmanagers) > 1:
             logger.warning('More than one altermanager found. Using first!')
 
-        manager = list(self.stored.alertmanagers.keys())[0]
-        alerting_config = self.stored.alertmanagers.get(manager, '')
+        targets = []
+        for manager in self.stored.alertmanagers.keys():
+            ip_addr = self.stored.alertmanagers[manager]['ip_addr']
+            port = self.stored.alertmanagers[manager]['port']
+            targets.append(f'{ip_addr}:{port}')
```
justinmclark, line 264: We shouldn't use f-strings since they aren't supported in Python 3.5 and charms need to support 3.5 as far as I'm aware.

(I know this because I did the same thing and Chipaca pointed it out to me).

justinmclark (changes requested):
Just one necessary change (see comment).

+1 otherwise

src/charm.py
dstathis: @chipaca Can you verify this? In what cases are charms running on python 3.5?

chipaca: anything based on ubuntu 16.04 or centos 7.

dstathis: @chipaca yes but given that this is a k8s charm, is juju ever going to deploy the operator pod as one of those OSes?

tests/test_charm.py
```diff
@@ -106,6 +114,7 @@ def test_grafana_is_provided_port_and_source(self):
         self.harness.add_relation_unit(rel_id, 'grafana/0')
         self.harness.update_relation_data(rel_id, 'grafana/0', {})
         data = self.harness.get_relation_data(rel_id, self.harness.model.unit.name)
+        print(f'HELLOG: {data}')
```
balbirthomas, line 117: Python 3.6 ! Checked with @jlounder ?

balbirthomas (approved):
Approving to progress this quickly. Will test locally after merge and push any fixes required.

src/charm.py
```diff
@@ -224,11 +230,13 @@ def _alerting_config(self):
             logger.debug('No alertmanagers available')
             return alerting_config
 
-        if len(self._stored.alertmanagers) > 1:
-            logger.warning('More than one altermanager found. Using first!')
+        targets = []
+        for manager in self.stored.alertmanagers:
+            port = self.stored.alertmanager_port
+            targets.append(f'{manager}:{port}')
```
balbirthomas, line 236: For Python 3.5 compatibility
```suggestion
            targets.append("{}:{}".format(manager, port))
```

justinmclark (approved):

balbirthomas (approved):

PR #8, opened 2020-10-29
Include expected logs in tests and avoid polluting test output with logs

mthaddon:
Include expected logs in tests and avoid polluting test output with logs


balbirthomas (approved):
Thank you very much. Will keep this good practice in mind for the future.

justinmclark (approved):

PR #13, opened 2020-11-13
Grafana-source relation interface name change

justinmclark:
This is a change required by Grafana.

### DIFF
--- a/metadata.yaml
+++ b/metadata.yaml
@@ -23,7 +23,7 @@ series:
     - kubernetes
 requires:
     grafana-source:
-        interface: grafana-dashsource
+        interface: grafana-datasource
     alertmanager:
         interface: alertmanager
 storage:


src/charm.py
```diff
@@ -24,8 +24,8 @@ def __init__(self, *args):
 
         super().__init__(*args)
 
-        self.stored.set_default(alertmanagers=[])
```
balbirthomas, line 27: Aren't these the same changes in https://github.com/canonical/prometheus-operator/pull/12 . If so, I think we should we just merge that pull request and make this pull request only about changing the Grafana interface name. I don't see the point of duplicating that work.

metadata.yaml
```diff
@@ -23,7 +23,7 @@ series:
     - kubernetes
 requires:
     grafana-source:
-        interface: grafana-dashsource
+        interface: grafana-datasource
```
balbirthomas, line 26: Would it be a good idea to run this through @jlounder ? If is ok with this I am.

balbirthomas (changes requested):

balbirthomas (approved):

dstathis (approved):

Issue #19, opened 2021-03-04
Can't configure custom Prometheus configuration


jonatasbaldin:
The Prometheus Charm doesn’t seem to have an easy way to configure any custom scrapping. The default configuration (`prometheus.yaml`) is hardcoded by the `_prometheus_config(self)` method and only scrapes itself.

With that, it’s not possible to configure the Prometheus Charm to scrape anything else, like other deployed Charms or Juju itself.

Is that assumption correct or am I missing something?

balbirthomas:
Hi @jonatasbaldin . At present there is a [branch](https://github.com/balbirthomas/prometheus-operator/tree/k8s-discovery) available that supports automatic discovery of scrape targets for Kubernetes clusters. The automatic discovery feature was removed after code review because the intention was to make configuration of scrape targets explicit by providing this information through relation data of any application that forms a relation with the Prometheus charm. This feature has not yet been implemented. 

Would it be sufficient if the Prometheus charm accepted a YAML scrape configuration from the client charm over application relation data ? Do you have any specific application charm in mind that you would like to relate with the Prometheus charm ?

ATTN: @jlounder would accepting YAML over relation data be an acceptable solution ?

jonatasbaldin:
I think a good solution would be to accept the whole `prometheus.yaml` file, like a config specifying the path to the file.

What do u think?

balbirthomas:
> I think a good solution would be to accept the whole `prometheus.yaml` file, like a config specifying the path to the file.

This is a bit risk isn't it ? The same Prometheus charm may be used by multiple client charms. If one client charm is allowed to overwrite the entire Prometheus config, it could interfere with Prometheus scrapping targets for other clients. This does this sound reasonable ?

balbirthomas:
> I think a good solution would be to accept the whole `prometheus.yaml` file, like a config specifying the path to the file.

Further to my point abovet. A config options that accepts a full `prometheus.yaml` can be implemented, if this will suit your use case. One limitation of this approach is that the administrator will need to provide and updated config file for every relation they add to the Prometheus charm.

jameinel:
I wouldn't make it a complete Prometheus.yaml, but being able to specify
scrape targets (which would then be merged with relation targets, etc)
would be useful.

balbirthomas:
[...]

PR #21, opened 2021-03-09
Add default RBAC for the Prometheus container

jonatasbaldin:
This PR adds a default RBAC to the Prometheus Pod, enabling it to scrape the Kubernetes API to configure the [Kubernetes Service Discovery](https://prometheus.io/docs/prometheus/latest/configuration/configuration/#kubernetes_sd_config).

I made a very simple PR to enable a default configuration, we can work on another PR to make it customizable via the `config.yaml`. What do you think?

Relates to #19.

### DIFF
--- a/src/charm.py
+++ b/src/charm.py
@@ -255,6 +255,52 @@ class PrometheusCharm(CharmBase):
 
         return yaml.dump(scrape_config)
 
+    def _prometheus_rbac(self):
+        """Construct Prometheus container RBAC
+        Based on https://github.com/prometheus-community/helm-charts/blob/97997ccc4bdd8699e74535dab24a861983a83b1e/charts/prometheus/templates/server/clusterrole.yaml
+        """
+
+        rbac_config = {
+            'roles': [{
+                'global': True,
+                'rules': [
+                    {
+                        'apiGroups': [''],
+                        'resources': [
+                            'nodes',
+                            'nodes/proxy',
+                            'nodes/metrics',
+                            'services',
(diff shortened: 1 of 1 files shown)


src/charm.py
```diff
@@ -301,7 +301,44 @@ def _build_pod_spec(self):
                         'content': self._prometheus_config()
                     }]
                 }]
-            }]
+            }],
+            'serviceAccount': {
+                'roles': [{
```
balbirthomas, line 306: It may be a good idea to have the value of `serviceAccount` generated and returned by a function. For example like Prometheus configuration is returned by `self._prometheus_config()`. What do you think ?

jonatasbaldin: Just did, what do u think?

balbirthomas: Looks good to me.

balbirthomas:
@jonatasbaldin I am closing this pull request because

1. This Prometheus charm has been migrated to a side car charm using [Pebble](https://ops.readthedocs.io/en/latest/#module-ops.pebble) and hence the service account settings for a "pod spec" are obsolete.
2. RBAC settings of this nature are only relevant if this Prometheus charm is deployed in the same model as the scrape target. This may not be good practice given Kubernetes failure modes.

It is on the development roadmap to support remote write to Prometheus through an ingress. Security related considerations will also be addressed in due course.

You efforts and feedback is greatly appreciated. If there are specific use cases that you still would like to see addressed please do not hesitate to create new Github issues for this repository.

PR #23, opened 2021-05-05
Charm libs

balbirthomas:
This pull request provides

1. An first version of the prometheus operator after migration to Pebble
2. An basic Prometheus provider/consumer charm library the accept scrape targets over relation data.

### DIFF
--- a/metadata.yaml
+++ b/metadata.yaml
@@ -19,8 +19,15 @@ tags:
     - monitoring
     - alerting
     - grafana
-series:
-    - kubernetes
+bases:
+    - name: ubuntu
+      channel: 20.04/stable
+containers:
+  prometheus:
+    resource: prometheus-image
+    mounts:
+      - storage: database
+        location: /var/lib/prometheus
 provides:
   monitoring:
     interface: prometheus
@@ -35,4 +42,8 @@ peers:
 storage:
   database:
     type: filesystem
-    location: /var/lib/prometheus
+resources:
+  prometheus-image:
+    type: oci-image
+    description: upstream docker image for prometheus
+    upstream-source: "prom/prometheus:latest"
--- a/metadata.yaml
+++ b/metadata.yaml
@@ -21,11 +21,17 @@ tags:
     - grafana
 series:
     - kubernetes
+provides:
+  monitoring:
+    interface: prometheus
 requires:
     grafana-source:
         interface: grafana-datasource
     alertmanager:
         interface: alertmanager
+peers:
+  prometheus:
+    interface: prometheus
 storage:
   database:
     type: filesystem
--- a/metadata.yaml
+++ b/metadata.yaml
@@ -19,9 +19,6 @@ tags:
     - monitoring
     - alerting
     - grafana
-bases:
-    - name: ubuntu
-      channel: 20.04/stable
 containers:
   prometheus:
     resource: prometheus-image
--- a/metadata.yaml
+++ b/metadata.yaml
@@ -33,9 +33,6 @@ requires:
         interface: grafana-datasource
     alertmanager:
         interface: alertmanager
-peers:
-  prometheus:
-    interface: prometheus
 storage:
   database:
     type: filesystem
--- a/src/charm.py
+++ b/src/charm.py
@@ -2,6 +2,7 @@
 # Copyright 2020 Balbir Thomas
 # See LICENSE file for licensing details.
 
+import hashlib
 import logging
 import yaml
 import json
@@ -9,10 +10,12 @@ import json
 from ops.charm import CharmBase
 from ops.framework import StoredState
 from ops.main import main
-from ops.model import ActiveStatus, MaintenanceStatus, BlockedStatus
+from ops.model import ActiveStatus, MaintenanceStatus, BlockedStatus, ModelError
+from ops.pebble import ConnectionError
 from prometheus_provider import MonitoringProvider
 from prometheus_server import Prometheus
 
+PROMETHEUS_CONFIG = "/etc/prometheus/prometheus.yml"
 logger = logging.getLogger(__name__)
 
 
@@ -29,7 +32,9 @@ class PrometheusCharm(CharmBase):
         self._stored.set_default(alertmanagers=[])
         self._stored.set_default(alertmanager_port='9093')
         self._stored.set_default(provider_ready=False)
+        self._stored.set_default(prometheus_config_hash=None)
 
+        self.framework.observe(self.on.prometheus_pebble_ready, self._setup_pebble_layers)
         self.framework.observe(self.on.config_changed, self._on_config_changed)
         self.framework.observe(self.on.start, self._on_start)
         self.framework.observe(self.on.stop, self._on_stop)
@@ -46,19 +51,37 @@ class PrometheusCharm(CharmBase):
             self.framework.observe(self.prometheus_provider.on.targets_changed,
                                    self._on_config_changed)
 
-    def _on_config_changed(self, _):
+    def _on_config_changed(self, event):
         """Set a new Juju pod specification
         """
-        self._configure_pod()
+        logger.info("Handling config changed")
+        container = self.unit.get_container("prometheus")
+
+        try:
+            service = container.get_service("prometheus")
+        except ConnectionError:
+            logger.info("Pebble API is not yet ready")
+            return
+        except ModelError:
+            logger.info("Prometheus service is not yet ready")
+            return
+
+        prometheus_config = self._prometheus_config()
+        config_hash = str(hashlib.md5(str(prometheus_config).encode('utf-8')))
+        if not self._stored.prometheus_config_hash == config_hash:
+            self._stored.prometheus_config_hash = config_hash
+            container.push(PROMETHEUS_CONFIG, prometheus_config)
+        else:
+            logger.info("No change in Prometheus configuration")
+            return
+
+        if service.is_running():
+            container.stop("prometheus")
+
+        container.start("prometheus")
+        logger.info("Restarted Prometheus service")
 
     def _on_start(self, event):
-        if not self._stored.provider_ready:
-            try:
-                _ = self.ingress_address
-            except Exception:
-                event.defer()
-                return
-
         provided = self.provides
         if provided:
             logger.debug("Prometheus provider is available")
@@ -91,7 +114,7 @@ class PrometheusCharm(CharmBase):
         self._stored.alertmanager_port = port
         self._stored.alertmanagers = addrs
 
-        self._configure_pod()
+        self._on_config_changed(event)
 
     def _on_alertmanager_broken(self, event):
         """Remove all alertmanager configuration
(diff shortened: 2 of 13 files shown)


mmanciop:
For the record, the paradigm implemented here would OTOH sound perfectly fine when implementing the Pushgateway protocol or `remote_write` :-)

balbirthomas:
> This PR implements an approach in which the application exposing a Prometheus endpoint opts in to Prometheus scraping it. The idea is cool, though the paradigm feels a little off when considering that, in Prometheus, the scraper (that is, the Prometheus server) is by design in control of what to scrape and how and how often and if and when to accept the data how it is, or rather [_relabel_ it](https://valyala.medium.com/how-to-use-relabeling-in-prometheus-and-victoriametrics-8b90fc22c4b2). With the paradigm as designed currently, we are exposing a mechanism in Juju that semantically means "Hey Prometheus do come and get these metrics, but rename them so and so using these `__meta_`". While technically can work, and it does sound like a useful feature, this inversion of control in terms of the final name of a metric feels somewhat off when keeping the standard, vanilla Prometheus experience in mind.

Intersting!. Just to clarify -- at present the only information forwarded by a prometheus client to prometheus is the IP address and the port on which the client is exposing its metrics. Though this is the only supported functionality in this PR there is no reason additional ways means of configuring Prometheus and its targets can not be done. As far as I understand regardless of what additional features are implemented each client will need to provide the prometheus server its metric endpoint information -- IP/port (this PR) and potentially path (at present defaulting to /metrics).

metadata.yaml
```diff
@@ -19,14 +19,31 @@ tags:
     - monitoring
     - alerting
     - grafana
-series:
-    - kubernetes
+bases:
+    - name: ubuntu
+      channel: 20.04/stable
```
sed-i, line 24: Is this needed? According to the [docs](https://discourse.charmhub.io/t/charm-metadata-v2/3674):
1. `series` is replaced by `containers/base`
2. `bases` is under `containers`, and "containers can include a single resource name **or** a list of bases."
2. there's a `bases` in `/manifest.yaml` (generated by charmcraft)

metadata.yaml
```diff
@@ -19,14 +19,31 @@ tags:
...
+provides:
+  monitoring:
+    interface: prometheus
 requires:
     grafana-source:
         interface: grafana-datasource
     alertmanager:
         interface: alertmanager
+peers:
+  prometheus:
```
sed-i, line 40: I like the idea of naming the peer relation "replicas". I think that brings more clarity at first glance.

metadata.yaml
```diff
@@ -19,14 +19,31 @@ tags:
...
+    interface: prometheus
 storage:
   database:
     type: filesystem
-    location: /var/lib/prometheus
+resources:
+  prometheus-image:
+    type: oci-image
+    description: upstream docker image for prometheus
+    upstream-source: "prom/prometheus:latest"
```
sed-i, line 49: the `upstream-source` property does not exist in the schema.

metadata.yaml
balbirthomas: There was a bug or actually missing feature in charmcraft because of which the charm would fail to build if bases was not listed in metadata.yaml, when I implemented this code. I can check again and remove this, if latest **stable** charmcraft does not have this issue.

metadata.yaml
balbirthomas: One needs to be careful here. Peers are not always "replicas". For example in mongodb they can be members of a sharded group. Even though this feature is not currently supported by the charm, the fact that peers need not be replicas I would be hesitant to accept a blanket prescription that peer relations should always be named "replicas".

metadata.yaml
balbirthomas: Hmm. I do remember seeing that. Will check again and  remove if it is indeed not supported.

src/charm.py
```diff
@@ -37,24 +45,97 @@ def __init__(self, *args):
...
-        """Set an alertmanager configuation
+        """Set an alertmanager configuration.
+
+        In response to any changes in relations with Alertmanager,
+        the list of currently available Alertmanagers is updated,
+        a new Prometheus configuration set and Prometheus is
+        restarted.
         """
         if not self.unit.is_leader():
             return
```
sed-i, line 141: what do you think about a `@leader_only` decorator?

src/charm.py
```diff
@@ -65,18 +146,37 @@ def _on_alertmanager_changed(self, event):
         self._stored.alertmanager_port = port
         self._stored.alertmanagers = addrs
 
-        self._configure_pod()
+        self._on_config_changed(event)
```
sed-i, line 149: not sure if it is better to just emit a ConfigChangedEvent instead

src/charm.py
balbirthomas: I hesitate to speculate :-) . However since you bring up the issue.  Would be the use case for such a decorator ? What would be the behaviour of such a decorator ? Would it immediately return if the current unit is not leader ? This can be done for the present. However the alertmanager event handler methods in this charm are very rudimentary. For example there is no logic there for handling forwarding rules from applications to prometheus to alertmanager. Even though this behaviour has not yet been discussed, agreed, and specified, it would be the next logical step to do once our charm libraries and pebble migration is complete. Would you suppose these altermanager event handlers would have no role to play if the unit is not leader once we add the logic for handling alerting rules and such ?.

sed-i: I see this exact `is_leader` preamble in many places in many repos, that's why I ask :)
 Don't know yet about alerting rules.

src/charm.py
balbirthomas: I do not see the benefit of one event handler to be raising another event, and the charm to then waiting for the next iteration of Juju invoking the same charm's code, for the so raised event to be then handled. Isn't it simpler if all that needs to done to handle the present event is done, even if this means invoking another function of the same charm, even if the other function is event handler ? Besides there is no reason to defer the event here either.

sed-i: I did not realize an emitted event is deferred

src/charm.py
balbirthomas: One can make a use case for `@leader_only` to be a utility in the Operator Framework. It would not reduce the duplication of code much if each "repo" tried to implement its own `@leader_only`. The use case may be stronger if you can think of a repertoire of such higher order functions. At present leader only seems to be the only reasonable one I have come across.

src/charm.py
balbirthomas: As far as I understand it will go into the queue of events maintained by Juju, and will be scheduled for forwarding to the charm.

metadata.yaml
rbarry82: Also true for Grafana, Alertmanager, and others. I worry that it's too application-specific to give it a blanket "replicas", and "peers" allows the charm author to decide what that should mean.

src/charm.py
rbarry82: From my POV, I'm a fan of `@leader_only`to reduce the boilerplate

src/charm.py
```diff
@@ -37,24 +45,97 @@ def __init__(self, *args):
...
+        """Provide Grafana with data source information.
+
+        Grafana needs to know the port and name of an application in order
+        to form a relation with it. Hence this information is provided here.
         """
         event.relation.data[self.unit]['port'] = str(self.model.config['port'])
         event.relation.data[self.unit]['source-type'] = 'prometheus'
+        event.relation.data[self.unit]['private-address'] = str(
+            self.model.get_binding(event.relation).network.bind_address
+        )
```
sed-i, line 130: does the use of a private address imply that all units are always deployed on the same local network?

balbirthomas: Good catch. I need to think about this one.

src/charm.py
balbirthomas: Aye. I do think it is a reasonable suggestion too. I would rather we discuss it in stand and create a ticket for `ops` where we make a list of such utilities that would be nice too have. Besides the focus of this PR is charm library and pebble migration. I am sure that there is lot more alertmanager work to be done. I will bring it in as part of one of those commits after discussion with team in standup.

src/charm.py
rbarry82: The underlying assumption from the inherited code may be that the LMA model running inside k8s will have an overlay network which routes between nodes.

It's probably a field which we should rename, and dynamically decide whether we want to use `.network.bind_address` or an ingress address if there's an ingress configured, but not all charms *should* necessarily have an ingress exposed unless it means something very different for charms than it means for k8s/"floating IPs" in openstack/etc

balbirthomas: I have created a github [issue ticket](https://github.com/canonical/prometheus-operator/issues/25) for gathering requirements and finalizing design on this issue. I am resolving this conversation and will take this issue up on a future commit, as discussed. Thank you for highlighting the issue.

metadata.yaml
balbirthomas: In view of the above. I am leaving this as is. Resolving conversation for now. Feel free to reopen if there are any new concerns.

metadata.yaml
balbirthomas: Just removed `bases` built and deployed charm. There was no error. So the issue seems to have been fixed. Pushing a commit now with bases removed.

metadata.yaml
balbirthomas: Looks like there is (or has) been an `upstream-source` property . See for instance

1. https://discourse.charmhub.io/t/writing-a-kubernetes-v1-charm/3976/3
2. https://discourse.charmhub.io/t/specify-image-or-resources-generally-via-url/2426/2

I do not know if this has been deprecated in metadata v2. If it has been it certainly has not been made explicit on the metadata v2 discourse page linked above.

src/charm.py
balbirthomas: @sed-i Let us make that github issue ticket on the Canonical Operator Framework repository, as discussed. Ping me when you are in and we can do it together. Resolving this conversation here for now.

metadata.yaml
sed-i: I did not mean as a blanket rule: if your charm has only one type of peers then plain "replicas" could do. Otherwise you could be more explicit, e.g. "prometheus-replicas". We tend to name everything with the same string literal - service, container, interface, peer, ... all named "prometheus" ... from which it is a bit difficult to draw context at first glance. Anyway, this was just a suggestion.

dstathis (changes requested):

src/charm.py
```diff
@@ -37,24 +45,97 @@ def __init__(self, *args):
...
+        if plan.services != layer["services"]:
+            container.add_layer("prometheus", layer, combine=True)
+
+            if container.get_service("prometheus").is_running():
+                container.stop("prometheus")
+
+            container.start("prometheus")
+            logger.info("Restarted prometheus container")
+
+        self.app.status = ActiveStatus()
```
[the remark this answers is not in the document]
balbirthomas: Indeed. Oversight. Thanks. Fixed.

src/charm.py
```diff
@@ -37,24 +45,99 @@ def __init__(self, *args):
...
+        layer = self._prometheus_layer()
+        plan = container.get_plan()
+        if plan.services != layer["services"]:
+            container.add_layer("prometheus", layer, combine=True)
+
+            if container.get_service("prometheus").is_running():
+                container.stop("prometheus")
+
+            container.start("prometheus")
+            logger.info("Restarted prometheus done")
```
dstathis, line 93: Minor nitpick here. This will say restarting even if it is being started for the first time.

dstathis (approved):

Abuelodelanada (approved):

src/charm.py
balbirthomas: Will reword in a future commit. Feel free to remind me, if I forget.

PR #27, opened 2021-05-19
Revert scrape config

balbirthomas:
1. Removing support for adding a Prometheus configuration using Juju command line. Also updated README accordingly.
2. Minor grammatical and spelling fixes.

### DIFF
--- a/README.md
+++ b/README.md
@@ -12,22 +12,25 @@ cluster. The alerting component of prometheus is offered through a
 separate Charm.
 
 
-## Configuration and Usage
+## Usage
 
-By default the Prometheus Operator monitors itself. There are two ways
-to provide additional scrape targets to this Prometheus charm.
+By default the Prometheus Operator monitors itself, but it also
+accepts additional scrape targets over Juju relations with charms that
+support the `prometheus` interface and preferably use the Prometheus
+charm library. This charm library provides an `add_endpoint()` method
+that creates additional scrape targets. Each scrape target is expected
+to expose a `/metrics` HTTP path that exposes its metrics in a
+Prometheus compatible format.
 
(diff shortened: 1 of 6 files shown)


rbarry82 (commented):
LGTM, but out of curiosity, why not support both?

Abuelodelanada (approved):
Looks good to me

balbirthomas:
> LGTM, but out of curiosity, why not support both?

Good Question. Just my two cents :-) below : 

The design goals of our charm ecosystem is that charms configure themselves through information exchanged over relations, which minimises the burden on system administrators. Seems like neither of us have any disagreement here. But as I see it this is the whole point of charms. So why bother also supporting the more cumbersome method of configuring Prometheus with a manually constructed YAML file. And why not discourage the practice of wanting to do so, thereby inviting more , better and refined use cases that we can add support for through relation based autoconfiguation. Besides what value does it add if users follow the same pattern of usage with charmed applications as those that are not charms.

I appreciate opinions can differ on this but @jlounder may have a reasonable basis to have requested this.

balbirthomas:
> Definitely true that it can clobber (and adding tooling to `promtool check ...`) is more than we want to do. I just wonder about users with "complex" configurations who will want to template it in advance, and whether removing a workflow without deprecation first will have side effects

Personally. This is exactly what I want :-) ---  "complains" (err.. use cases) and understanding of "complex configuations" (more use cases) so it will fuel the discussion on exactly what feature sets we want to add to put into the Prometheus charm. This PR is my tool to ensure that those discussions happen, rather than people finding ad hoc ways of solving their uses cases by slapping an an ad hoc YAML config.

sed-i (approved):

dstathis (approved):

PR #29, opened 2021-05-19
rel id in no longer stored in ConsumerBase

dstathis:


lib/charms/prometheus/v1/prometheus.py
```diff
@@ -94,8 +93,7 @@ def remove_endpoint(self, address, port=80, rel_id=None):
                 :class:`PrometheusConsumer` has been instantiated in
                 `multi` mode.
         """
-        if rel_id is None:
-            rel_id = self.relation_id
+        rel_id = self.framework.model.get_relation(self._relation_name, rel_id).id
```
sed-i, line 96: remove the 'rel_id` method argument?

lib/charms/prometheus/v1/prometheus.py
```diff
@@ -69,8 +69,7 @@ def add_endpoint(self, address, port=80, rel_id=None):
...
```
sed-i, line 72: remove the `rel_id` method argument?

lib/charms/prometheus/v1/prometheus.py
sed-i, line 96: reminds me the ren & stimpy episode where stimpy climbed into his own belly button

balbirthomas: What are you trying to achieve here ? Please explain.  I do not see a commit message or a comment. Are you trying to fix a bug for instance ? What is the bug ?

dstathis:
This was a false bug. Closing.

PR #31, opened 2021-05-21
Default http port is not special

balbirthomas:
This commit removes a redundant ternary operator that
was left over from an early development version of the
code and overlooked.

### DIFF
--- a/lib/charms/prometheus/v1/prometheus.py
+++ b/lib/charms/prometheus/v1/prometheus.py
@@ -76,7 +76,7 @@ class PrometheusConsumer(ConsumerBase):
         if address in targets:
             return
 
-        target = address if port == 80 else address + ":" + str(port)
+        target = address + ":" + str(port)
         targets.append(target)
         self._update_targets(targets, rel_id)
 


lib/charms/prometheus/v1/prometheus.py
```diff
@@ -76,7 +76,7 @@ def add_endpoint(self, address, port=80, rel_id=None):
         if address in targets:
             return
 
-        target = address if port == 80 else address + ":" + str(port)
+        target = address + ":" + str(port)
```
sed-i, line 79: may we use f-strings? `target = f"{address}:{port}"`

sed-i (approved):

lib/charms/prometheus/v1/prometheus.py
balbirthomas: Requires 3.6. But I agree good idea.

Abuelodelanada: As far as I remember, our charms have yet to be compatible with Python 3.5. 

If so we can use:

```python
target = "{}:{}".format(address, port)
```

Anyway it is a matter of taste. I prefer string interpolation over concatenation.

balbirthomas: Indeed this is a mater of taste. Here is my question to both of you. Should "matters of taste"  be blocking issues for code review ? Or should they just be comments (identified as such) left to the discretion of the code author ?

Abuelodelanada (approved):

lib/charms/prometheus/v1/prometheus.py
dstathis: +1 for f-strings but yes it doesn't really matter

String formatting of some sort, though, is typically preferred over concatenation as the language will occasionally give minor optimizations with formatting.

dstathis (approved):

lib/charms/prometheus/v1/prometheus.py
balbirthomas: I like string formatting too. But there are multiple places in this code where it needs to be done or it will be an inconsistent style. I will pick this up in a future commit. Will merge for now.

PR #42, opened 2021-06-08
Fix two bugs with other small refactors

balbirthomas:
1. Fix two timing bugs (mentioned in commit message)
2. Small refactor of names and methods to isolate configuration of Prometheus into a separate function that is not an event handler.
3. Fix import statement in charm library unit test

### DIFF
--- a/src/charm.py
+++ b/src/charm.py
@@ -33,7 +33,7 @@ class PrometheusCharm(CharmBase):
         self._stored.set_default(provider_ready=False)
         self._stored.set_default(prometheus_config_hash=None)
 
-        self.framework.observe(self.on.prometheus_pebble_ready, self._on_config_changed)
+        self.framework.observe(self.on.prometheus_pebble_ready, self._on_pebble_ready)
         self.framework.observe(self.on.config_changed, self._on_config_changed)
         self.framework.observe(self.on.stop, self._on_stop)
         self.framework.observe(self.on['alertmanager'].relation_changed,
@@ -48,9 +48,24 @@ class PrometheusCharm(CharmBase):
             self.prometheus_provider = MonitoringProvider(self,
                                                           'monitoring', 'prometheus', self.version)
             self.framework.observe(self.prometheus_provider.on.targets_changed,
-                                   self._on_config_changed)
+                                   self._on_scrape_targets_changed)
+
+    def _on_pebble_ready(self, event):
+        """Setup workload container configuration.
+        """
+        self._configure()
 
     def _on_config_changed(self, event):
+        """Handle a configuration change.
+        """
+        self._configure()
+
+    def _on_scrape_targets_changed(self, event):
+        """Handle changes in scrape targets.
+        """
+        self._configure()
+
+    def _configure(self):
         """Reconfigure and restart Prometheus.
 
         In response to any configuration change, such as a new consumer
@@ -58,7 +73,7 @@ class PrometheusCharm(CharmBase):
         Prometheus config file is regenerated, pushed to the workload
         container and Prometheus is restarted.
         """
-        logger.info("Handling config changed")
+        logger.info("Configuring Prometheus")
         container = self.unit.get_container("prometheus")
 
         # validate configuration options
@@ -74,7 +89,7 @@ class PrometheusCharm(CharmBase):
         # new config file to the workload container
         prometheus_config = self._prometheus_config()
         config_hash = str(hashlib.md5(str(prometheus_config).encode('utf-8')))
-        if not self._stored.prometheus_config_hash == config_hash:
+        if self._stored.prometheus_config_hash != config_hash:
             try:
                 container.push(PROMETHEUS_CONFIG, prometheus_config)
                 self._stored.prometheus_config_hash = config_hash
@@ -135,7 +150,7 @@ class PrometheusCharm(CharmBase):
 
         self._stored.alertmanagers = addrs
 
-        self._on_config_changed(event)
+        self._configure()
 
     def _on_alertmanager_broken(self, event):
         """Remove all alertmanager configuration.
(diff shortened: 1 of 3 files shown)


src/charm.py
```diff
@@ -44,21 +44,36 @@ def __init__(self, *args):
         self.framework.observe(self.on['grafana-source'].relation_changed,
                                self._on_grafana_changed)
 
-        if self._stored.provider_ready:
+        if self.provider_ready:
             self.prometheus_provider = MonitoringProvider(self,
                                                           'monitoring', 'prometheus', self.version)
             self.framework.observe(self.prometheus_provider.on.targets_changed,
-                                   self._on_config_changed)
+                                   self._on_scrape_targets_changed)
```
sed-i, line 51: I would prefer to avoid partial construction even if it means more sophisticated logic later.

balbirthomas: Not sure what you mean. The whole point of `self.provider_ready` is to avoid "partial construction". There is a [ticket](https://github.com/canonical/prometheus-operator/issues/41) to explore alternative options.

src/charm.py
```diff
@@ -44,21 +44,36 @@ def __init__(self, *args):
...
+    def _configure(self):
         """Reconfigure and restart Prometheus.
 
         In response to any configuration change, such as a new consumer
         relation, or a new configuration set by the administrator, the
         Prometheus config file is regenerated, pushed to the workload
         container and Prometheus is restarted.
         """
-        logger.info("Handling config changed")
+        logger.info("Configuring Prometheus")
```
sed-i, line 76: Entrance message should probably be debug. Later on (under config hash change) makes sense to have info.

src/charm.py
balbirthomas: Also not that `ProviderBase` has a `ready()` method that was perhaps intended to deal with partial construction. I am yet to explore how this can be integrated into this library.

src/charm.py
```diff
@@ -74,10 +89,14 @@ def _on_config_changed(self, event):
...
-            container.push(PROMETHEUS_CONFIG, prometheus_config)
-            logger.info("Pushed new configuration")
+        if self._stored.prometheus_config_hash != config_hash:
+            try:
+                container.push(PROMETHEUS_CONFIG, prometheus_config)
+                self._stored.prometheus_config_hash = config_hash
+                logger.info("Pushed new configuration")
+            except ConnectionError:
+                logger.info("Ignoring config changed since pebble is not ready")
+                return
```
sed-i, line 99: this is an anti pattern imho: `_stored.pebble_ready` seems like a better approach.

balbirthomas: Why do you think it is the "better approach" ? .  I do not think charms should be tracking the history of Juju events and storing this in their state.

src/charm.py
balbirthomas: Hmm... Isn't it useful for a system administrator to see that prometheus underwent a configure step when he added a relation. He can use the debug messages if he wants more details.

src/charm.py
```diff
@@ -431,6 +436,25 @@ def version(self):
...
+        with valid information, its status is recorded.
+
+        Returns:
+            True if Prometheus is ready, False otherwise
+        """
+        provided = {'prometheus': self.version}
+        if not self._stored.provider_ready and provided['prometheus']:
+            logger.debug("Prometheus provider is available")
+            logger.debug("Providing : {}".format(provided))
+            self._stored.provider_ready = True
```
sed-i, line 454: usually getters do not have side effects

balbirthomas: If you look at it from that rigid perspective. Or you can look at it as a utility to know if provider is ready or not.

tests/test_lib.py
```diff
@@ -7,7 +7,7 @@
 from ops.charm import CharmBase
 from ops.framework import StoredState
 from ops.testing import Harness
-from lib.charms.prometheus.v1.prometheus import PrometheusConsumer
+from lib.charms.prometheus_k8s.v1.prometheus import PrometheusConsumer
```
sed-i, line 10: `lib` is in the python path and should be omitted here, i.e. should just be `charms....`
```
./run_tests:10:    export PYTHONPATH="lib:src"
```

sed-i (dismissed):

tests/test_lib.py
balbirthomas: Ah thanks. I had a old version of `run_tests` and was missing this. Maybe charmcraft has been updated since then. will fix and push shortly.

src/charm.py
sed-i: I meant that the charm class is constructed conditionally: at different times it may have different composition.

src/charm.py
sed-i: So `_configure` is a common exit hook to many events. You may end up seeing this message even if effectively nothing changed. Guarded `info` messages are more precise.

src/charm.py
balbirthomas: Yes. That I will try and address in a future enhancement as it may not be trivial. I have to explore it.

src/charm.py
sed-i: With`_stored.pebble_ready` you'd have a one-liner assignment, and elsewhere you'd have a one-liner check.
The try-except block here is effectively tracking the history of juju events as well :)

src/charm.py
balbirthomas: It is true this configure message is going to appear multiple times in the lifecycle of the charm. But each one of those are significant events in the life of the charm. As opposed to say a update status.

tests/test_lib.py
balbirthomas: DONE.

src/charm.py
balbirthomas: Also I plan to explore removing stored state from this charm and have created a [ticket](https://github.com/canonical/prometheus-operator/issues/46) for this.

sed-i (approved):

rbarry82 (approved):
LGTM

PR #48, opened 2021-06-10
better interface with alertmanager

### LINKED ISSUE #47: config_hash calculated incorrectly
https://github.com/canonical/prometheus-operator/blob/56c820f5cd093846fc0fe0f642e087f5782d30d9/src/charm.py#L91

`config_hash` may be the same value even for different strings:
```
>>> prometheus_config1 = "{'global': {'scrape_interval': '1m', 'scrape_timeout': '10s', 'evaluation_interval': '1m'}, 'scrape_configs': [{'job_name': 'prometheus', 'scrape_interval': '5s', 'scrape_timeout': '5s', 'metrics_path': '/metrics', 'honor_timestamps': True, 'scheme': 'http', 'static_configs': [{'targets': ['localhost:9090']}]}], 'alerting': {'alertmanagers': [{'static_configs': [{'targets': ['10.1.157.127', '10.1.157.71', '10.1.157.74']}]}]}}"
>>> prometheus_config2 = "{'global': {'scrape_interval': '1m', 'scrape_timeout': '10s', 'evaluation_interval': '1m'}, 'scrape_configs': [{'job_name': 'prometheus', 'scrape_interval': '5s', 'scrape_timeout': '5s', 'metrics_path': '/metrics', 'honor_timestamps': True, 'scheme': 'http', 'static_configs': [{'targets': ['localhost:9090']}]}], 'alerting': {'alertmanagers': [{'static_configs': [{'targets': ['10.1.157.127', '10.1.157.71', '10.1.157.72', '10.1.157.74']}]}]}}"
>>> str(hashlib.md5(str(prometheus_config1).encode('utf-8')))
'<md5 HASH object @ 0x7f54ddcdfa10>'
>>> str(hashlib.md5(str(prometheus_config2).encode('utf-8')))
'<md5 HASH object @ 0x7f54ddcdfa10>'
```
A call to `.hexdigest()` is missing:
```
>>> hashlib.md5(str(prometheus_config1).encode('utf-8')).hexdigest()
'48e7ae9dce66f3022a3b5342e002c356'
>>> hashlib.md5(str(prometheus_config2).encode('utf-8')).hexdigest()
'6e1dcd35bee04b36d76674128782230b'
```

sed-i:
Updates to prometheus, to go with corresponding [changes in alertmanager](https://github.com/canonical/alertmanager-operator/pull/7).

### DIFF
--- a/lib/charms/alertmanager_k8s/v0/alertmanager.py
+++ b/lib/charms/alertmanager_k8s/v0/alertmanager.py
@@ -1,0 +1,132 @@
+#!/usr/bin/env python3
+# Copyright 2021 Canonical Ltd.
+# See LICENSE file for licensing details.
+
+""" # AlertmanagerConsumer library
+
+This library is design to be used by a charm consuming or providing the alertmanager-k8s relation.
+"""
+
+import ops
+from ops.framework import StoredState
+from ops.relation import ConsumerBase, ProviderBase
+from ops.charm import CharmBase
+
+from typing import List
+import logging
+
+LIBID = "abcdef1234"  # Unique ID that refers to the library forever
+LIBAPI = 0    # Must match the major version in the import path.
+LIBPATCH = 1  # The current patch version. Must be updated when changing.
+
+logger = logging.getLogger(__name__)
+
+
+class AlertmanagerConsumer(ConsumerBase):
+    """A "consumer" handler to be used by charms that relate to Alertmanager.
+    This consumer auto-registers relation events on behalf of the user and communicates information
+    directly via `_stored` TODO: have a documented contract and act on it in the "available" hook.
+
+    Arguments:
+            charm (CharmBase): consumer charm
+            relation_name (str): from consumer's metadata.yaml
+            consumes (dict): provider specifications
+            multi (bool): multiple relations flag
+
+    Attributes:
+            charm (CharmBase): consumer charm
+    """
+    _stored: StoredState
+
+    def __init__(self, charm: CharmBase, relation_name: str, consumes: dict, multi: bool = False):
+        super().__init__(charm, relation_name, consumes, multi)
+        self.charm = charm
+        self._consumer_relation_name = relation_name  # from consumer's metadata.yaml
+        # self._provider_relation_name = "alerting"  # from alertmanager's metadata.yaml
+
+        self.framework.observe(self.charm.on[self._consumer_relation_name].relation_changed,
+                               self._on_relation_changed)
+        self.framework.observe(self.charm.on[self._consumer_relation_name].relation_departed,
+                               self._on_relation_departed)
+        self.framework.observe(self.charm.on[self._consumer_relation_name].relation_broken,
+                               self._on_relation_broken)
+
+        self._stored.set_default(alertmanagers={})
+
+    def _on_relation_changed(self, event: ops.charm.RelationChangedEvent):
+        """This hook stores locally the address of the newly-joined alertmanager.
+        This is needed for consumers such as prometheus, which should be aware of all alertmanager
+        instances.
+        """
+        if event.unit:  # event.unit may be `None` in the case of app data change
+            # Save locally the public IP address of the alertmanager unit
+            if address := event.relation.data[event.unit].get("public_address"):
+                # TODO consider storing in unit data instead of StoredState
+                self._stored.alertmanagers[event.unit.name] = address
+
+                # inform consumer about the change
+                # manually emitting the "available" event until scale-up/scale-down "ready" flag is fixed
+                self.on.available.emit()
+
+    def get_cluster_info(self) -> List[str]:
+        """Returns a list of ip addresses of all the alertmanager units
+        """
+        return sorted(list(self._stored.alertmanagers.values()))
+
+    def _on_relation_departed(self, event: ops.charm.RelationDepartedEvent):
+        """This hook removes the address of the departing alertmanager from its local store.
+        This is needed for consumers such as prometheus, which should be aware of all alertmanager
+        instances.
+        """
+        if self._stored.alertmanagers.pop(event.unit.name, None):
+            # inform consumer about the change
+            # manually emitting the "available" event until scale-up/scale-down "ready" flag is fixed
+            self.on.available.emit()
+
+    def _on_relation_broken(self, event: ops.charm.RelationBrokenEvent):
+        self._stored.alertmanagers.clear()
+        # inform consumer about the change
+        # manually emitting the "available" event until scale-up/scale-down "ready" flag is fixed
+        self.on.available.emit()
+
+
+class AlertmanagerProvider(ProviderBase):
+    """A "provider" handler to be used by the Alertmanager charm for abstracting away all the
+    communication with consumers.
+    This provider auto-registers relation events on behalf of the main Alertmanager charm.
+
+    Arguments:
+            charm (CharmBase): consumer charm
+            service_name (str): a name for the provided service
+            consumes (dict): provider specifications
+            multi (bool): multiple relations flag
+
+    Attributes:
+            charm (CharmBase): the Alertmanager charm
+    """
+    _provider_relation_name = "alerting"
+    _public_api_port: int = 9093  # public port may not be the same as AlertmanagerCharm._api_port
+
+    def __init__(self, charm, service_name: str, version: str = None):
+        super().__init__(charm, self._provider_relation_name, service_name, version)
+        self.charm = charm
+        self._service_name = service_name
+
+        events = self.charm.on[self._provider_relation_name]
+        self.framework.observe(events.relation_joined, self._on_relation_joined)
+
+        # No need to observe `relation_departed` or `relation_broken`: data bags are auto-updated
+        # so both events are address on the consumer side.
+        self.framework.observe(events.relation_joined, self._on_relation_joined)
+
+    def _on_relation_joined(self, event: ops.charm.RelationJoinedEvent):
+        """This hook stores the public address of the newly-joined "alerting" relation in the
+        corresponding data bag.
+        This is needed for consumers such as prometheus, which should be aware of all alertmanager
+        instances.
+        """
+        # "ingress-address" is auto-populated incorrectly so rolling my own, "public_address"
+        event.relation.data[self.charm.unit]["public_address"] = "{}:{}".format(
+            self.model.get_binding(event.relation).network.bind_address,
+            self._public_api_port
+        )
--- a/lib/charms/alertmanager_k8s/v0/alertmanager.py
+++ b/lib/charms/alertmanager_k8s/v0/alertmanager.py
@@ -8,7 +8,7 @@ This library is design to be used by a charm consuming or providing the alertman
 """
 
 import ops
-from ops.framework import StoredState
+from ops.framework import StoredState, EventSource, EventBase
 from ops.relation import ConsumerBase, ProviderBase
 from ops.charm import CharmBase
 
@@ -16,16 +16,46 @@ from typing import List
 import logging
 
 LIBID = "abcdef1234"  # Unique ID that refers to the library forever
-LIBAPI = 0    # Must match the major version in the import path.
+LIBAPI = 0  # Must match the major version in the import path.
 LIBPATCH = 1  # The current patch version. Must be updated when changing.
 
 logger = logging.getLogger(__name__)
 
 
+class ClusterChanged(EventBase):
+    """Event raised when an alertmanager cluster is changed.
+
+    If an alertmanager unit is added to or removed from a relation,
+    then a :class:`ClusterChanged` event is raised.
+    """
+
+    def __init__(self, handle, data=None):
+        super().__init__(handle)
+        self.data = data
+
+    def snapshot(self):
+        """Save relation data."""
+        return {"data": self.data}
+
+    def restore(self, snapshot):
+        """Restore relation data."""
+        self.data = snapshot["data"]
+
+
 class AlertmanagerConsumer(ConsumerBase):
     """A "consumer" handler to be used by charms that relate to Alertmanager.
+
     This consumer auto-registers relation events on behalf of the user and communicates information
-    directly via `_stored` TODO: have a documented contract and act on it in the "available" hook.
+    directly via `_stored`
+    Every change in the alertmanager cluster emits a :class:`ClusterChanged` event that the
+    consumer charm can register and handle, for example:
+
+        self.framework.observe(self.alertmanager_lib.cluster_changed,
+                               self._on_alertmanager_cluster_changed)
+
+    The updated alertmanager cluster can then be obtained via the
+
+    This consumer library expect the consumer charm to register the `get_cluster_info` method.
 
     Arguments:
             charm (CharmBase): consumer charm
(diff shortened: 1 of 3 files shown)


src/charm.py
```diff
@@ -88,7 +95,7 @@ def _configure(self):
         # check if configuration file has changed and if so push the
         # new config file to the workload container
         prometheus_config = self._prometheus_config()
-        config_hash = str(hashlib.md5(str(prometheus_config).encode('utf-8')))
+        config_hash = hashlib.md5(str(prometheus_config).encode('utf-8')).hexdigest()
```
balbirthomas, line 98: Thanks. Although it may be more appropriate if a  separate PR was created just for this issue so that it can be merged without waiting for this draft PR and besides the two change sets are unrelated.

lib/charms/alertmanager_k8s/v0/consumer.py
```diff
@@ -0,0 +1,106 @@
...
+        self.framework.observe(self.charm.on[self._consumer_relation_name].relation_changed,
+                               self._on_relation_changed)
+        self.framework.observe(self.charm.on[self._consumer_relation_name].relation_departed,
+                               self._on_relation_departed)
+        self.framework.observe(self.charm.on[self._consumer_relation_name].relation_broken,
+                               self._on_relation_broken)
+
+        self._stored.set_default(alertmanagers={})
+
+    def _on_relation_changed(self, event: ops.charm.RelationChangedEvent):
```
balbirthomas, line 56: I am curious to understand why you choose to make the consumer (rather than the provider) responsible for aggregating "public address" of all Alertmanager peers. In the same vein I would be interested in your rationale for using stored state in the charm library to store this list rather than application data.

lib/charms/alertmanager_k8s/v0/consumer.py
```diff
@@ -0,0 +1,106 @@
...
+        instances.
+        """
+        if event.unit:  # event.unit may be `None` in the case of app data change
+            # Save locally the public IP address of the alertmanager unit
+            if address := event.relation.data[event.unit].get("public_address"):
+                # TODO consider storing in unit data instead of StoredState
+                self._stored.alertmanagers[event.unit.name] = address
+
+                # inform consumer about the change
+                self.on.available.emit()
```
balbirthomas, line 68: Doesn't this mean there will be two `available` events emitted for each joined unit (since the ConsumerBase also emits one) ? Is that the intention ?  Instead would it be possible to just ensure that `get_cluster_info()` provides the required list of addresses to Prometheus when `Prometheus.AlertmanagerConsumer` raises `available` ?

lib/charms/alertmanager_k8s/v0/consumer.py
```diff
@@ -0,0 +1,106 @@
...
+        self.charm = charm
+        self._consumer_relation_name = relation_name  # from consumer's metadata.yaml
+        # self._provider_relation_name = "alerting"  # from alertmanager's metadata.yaml
+
+        self.framework.observe(self.charm.on[self._consumer_relation_name].relation_changed,
+                               self._on_relation_changed)
+        self.framework.observe(self.charm.on[self._consumer_relation_name].relation_departed,
+                               self._on_relation_departed)
+        self.framework.observe(self.charm.on[self._consumer_relation_name].relation_broken,
+                               self._on_relation_broken)
```
balbirthomas, line 52: What should happen when alertmanager is upgraded and no longer compatible with Prometheus. Can this happen ?

src/charm.py
```diff
@@ -36,10 +37,10 @@ def __init__(self, *args):
         self.framework.observe(self.on.prometheus_pebble_ready, self._on_pebble_ready)
         self.framework.observe(self.on.config_changed, self._on_config_changed)
         self.framework.observe(self.on.stop, self._on_stop)
-        self.framework.observe(self.on['alertmanager'].relation_changed,
-                               self._on_alertmanager_changed)
-        self.framework.observe(self.on['alertmanager'].relation_broken,
-                               self._on_alertmanager_broken)
+        # self.framework.observe(self.on['alertmanager'].relation_changed,
```
balbirthomas, line 40: Why not just delete it instead of commenting it out ? Or are you planning to uncomment it in a future revision of this draft ?

src/charm.py
```diff
@@ -50,6 +51,12 @@ def __init__(self, *args):
             self.framework.observe(self.prometheus_provider.on.targets_changed,
                                    self._on_scrape_targets_changed)
 
+        self.alertmanager_lib = AlertmanagerConsumer(self,
+                                                     relation_name="alertmanager",
+                                                     consumes={'alertmanager': '>0.0.0'})
```
balbirthomas, line 56: If I remember correctly you can also pass None for version if there is no restriction on version. Please do check this if you want to do it that way. Also I presume you have indeed ascertained that any Prometheus version will work with any version of Alertmanager.

src/charm.py
```diff
@@ -50,6 +51,12 @@ def __init__(self, *args):
...
+        self.framework.observe(self.alertmanager_lib.on.available,
```
balbirthomas, line 57: Very minor style comment. Do you think of the composition here as "a Prometheus object contains annAlertmanager Library" object or "a Prometheus object contains and Altermanager Consumer" object or even just and "Alertmanager" object.

src/charm.py
```diff
@@ -135,33 +142,8 @@ def _on_grafana_changed(self, event):
...
-
-        When an Alertmanager departs it is removed from the list
-        of currently available Alertmanagers, the Prometheus configuration
-        is updated and Prometheus is restarted.
-        """
-        if not self.unit.is_leader():
-            return
-        self._stored.alertmanagers.clear()
+    def _on_alertmanager_cluster_changed(self, event):
+        self._stored.alertmanagers = self.alertmanager_lib.get_cluster_info()
```
balbirthomas, line 146: Is there any reason to store the list of alert managers both in the AlertmanagerConsumer library and the Prometheus Charm ? Why not just integrate it into `_alerting_config()` ?

tests/test_charm.py
```diff
@@ -64,14 +64,14 @@ def test_alerting_config_is_updated_by_alertmanager_relation(self, push):
...
-        # check alerting config is updated when a alertmanager joins
-        self.harness.update_relation_data(rel_id,
-                                          'alertmanager',
-                                          {
-                                              'addrs': '["192.168.0.1:9093"]'
-                                          })
-        config = push.call_args[0]
-        self.assertEqual(alerting_config(config), SAMPLE_ALERTING_CONFIG)
+        # # check alerting config is updated when a alertmanager joins
+        # self.harness.update_relation_data(rel_id,
```
balbirthomas, line 68: Same question about when not just delete rather than commenting it out. PS: I have not check test coverage since this is a draft. I trust your workmanship here.

tests/test_charm.py
```diff
@@ -83,13 +83,13 @@ def test_alerting_config_is_removed_when_alertmanager_is_broken(self, push):
...
         self.assertIsInstance(rel_id, int)
         self.harness.add_relation_unit(rel_id, 'alertmanager/0')
-        self.harness.update_relation_data(rel_id,
-                                          'alertmanager',
-                                          {
-                                              'addrs': '["192.168.0.1:9093"]'
-                                          })
-        config = push.call_args[0]
-        self.assertEqual(alerting_config(config), SAMPLE_ALERTING_CONFIG)
+        # self.harness.update_relation_data(rel_id,
```
balbirthomas, line 86: ditto.

balbirthomas (approved):

lib/charms/alertmanager_k8s/v0/consumer.py
```diff
@@ -0,0 +1,106 @@
...
+            relation_name (str): from consumer's metadata.yaml
+            consumes (dict): provider specifications
+            multi (bool): multiple relations flag
+
+    Attributes:
+            charm (CharmBase): consumer charm
+    """
+    _stored: StoredState
+
+    def __init__(self, charm: ops.charm.CharmBase, relation_name, consumes, multi=False):
```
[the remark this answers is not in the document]
rbarry82: It's odd to just have a typing hint on one of the arguments also.

I'd suggest importing `CharmBase` so this looks cleaner and/or also including types for the rest of the non-`self` arguments.

lib/charms/alertmanager_k8s/v0/consumer.py
sed-i: This reflects my current understanding, and was tested successfully on 2.9.3 and 2.9.5.
I opened an [issue](https://github.com/canonical/alertmanager-operator/issues/8) to refactor this into relation data.

src/charm.py
sed-i: #49

lib/charms/alertmanager_k8s/v0/consumer.py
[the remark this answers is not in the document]
sed-i: Changed to `CharmBase` and added remaining annotations.

lib/charms/alertmanager_k8s/v0/consumer.py
sed-i: In that case there will be a "WARNING" entry in the log about incompatibility.
Unless you meant something else?

src/charm.py
sed-i: Removed

src/charm.py
sed-i: IIRC, `None` was producing a WARNING in the log. Regardless, it's good to have it versioned.
After we merge the alertmanager PR, I can change this to correspond.

src/charm.py
sed-i: The `_lib` follows the graylog pattern. I am trying to think of it as a library for prom.

src/charm.py
sed-i: I intentionally kept changes to the prom op to a minimum, to focus on the interface only.

lib/charms/alertmanager_k8s/v0/consumer.py
balbirthomas: ok. so you are leaving that to the base class.That is fine as long as you are aware.

src/charm.py
balbirthomas: Thanks. I will look at that warning and think about it. Do paste that warning if you will. And if you care try empty string too.

src/charm.py
balbirthomas: Your call. As I said this is only a style question. I was trying to get you to reflect upon if objects in your programming model compose other objects or libraries ?

src/charm.py
balbirthomas: Can I request you to create a ticket for future refactor either in prometheus repo or alertmanager repo or both, if you agree that a refactor here may be a good idea. I think it does not make sense to store it in two places. Even more it does not make sense to store it at all. You already have a ticket to that effect (i.e. to use app data).

lib/charms/alertmanager_k8s/v0/consumer.py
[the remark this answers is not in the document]
balbirthomas: Hmmm... If you must annotate IMHO it is best sparingly done only when required, if type is not obvious, or complex (cf. [non-goals](https://www.python.org/dev/peps/pep-0526/#non-goals))

jameinel:
I think you are misreading non-goals. "Third party packages will have to be
developed to implement such functionality". The non goal is that core
python won't do it, but mypy and other tools *do*.
As do code editors that then let you do autocomplete on attributes, etc.

I believe OF went with all public methods should have type annotations.

John
=:->

lib/charms/alertmanager_k8s/v0/consumer.py
balbirthomas: Of course Prometheus will have to "pull" from consumer lib using `get_cluster_info()`. I mean `ConsumerBase` emits `ProviderAvailable` if the provider and consumer are compatible. How does this check for compatibility factor into the workflow being proposed in this PR. Can it happen that `ConsumerBase` emits `ProviderInvalid` but `AlertmanagerConsumer` emits `ProviderAvailable` in response to the same relation changed event ?

balbirthomas:
> I think you are misreading non-goals. "Third party packages will have to be developed to implement such functionality".  I believe OF went with all public methods should have type annotations. 

I would distinguish between OF, charm and charm library. While a charm has a very sparing "public interface" a charm library does indeed have some. Providing annotations in public interfaces (particularly for complex types) would be reasonable. But mandating it for every string, int etc. I fear will unnecessarily clutter code.

lib/charms/alertmanager_k8s/v0/consumer.py
sed-i: That is a very good point.
That seems to suggest we should consider a custom event after all?

balbirthomas: Custom event would certainly be one way. I am always looking for the "simplest" way. I have not dug deep into this as I trust you are in a better position to do that and trust your judgement. But if you would like to keep things simple kindly do consider if it is possible that the `AlertmanagerConsumer` charm library exposes just one function `get_cluster_info()` and nothing else (no event handler). Since Prometheus just needs the list of alertmangers any time one of the `ConsumerBase` events fire.

tests/test_charm.py
sed-i: removed alertmanager test

tests/test_charm.py
sed-i: removed alertmanager test

src/charm.py
sed-i: removed `_stored.alertmanagers`

lib/charms/alertmanager_k8s/v0/consumer.py
sed-i: I tested without manually emitting, and occasionally there are misses. I'll keep it in for now until the `ready` flag on scale-up/down is resolved.

lib/charms/alertmanager_k8s/v0/alertmanager.py
```diff
@@ -0,0 +1,132 @@
...
+    _provider_relation_name = "alerting"
+    _public_api_port: int = 9093  # public port may not be the same as AlertmanagerCharm._api_port
+
+    def __init__(self, charm, service_name: str, version: str = None):
+        super().__init__(charm, self._provider_relation_name, service_name, version)
+        self.charm = charm
+        self._service_name = service_name
+
+        events = self.charm.on[self._provider_relation_name]
+        self.framework.observe(events.relation_joined, self._on_relation_joined)
```
balbirthomas, line 116: Isn't like 116 a duplicate of line 120 or am I missing something ?

lib/charms/alertmanager_k8s/v0/consumer.py
balbirthomas: I do not see how `ready` flag will solve the issue being raised here.

lib/charms/alertmanager_k8s/v0/alertmanager.py
sed-i: Removed.

lib/charms/alertmanager_k8s/v0/consumer.py
dstathis: Resolving conversation as a custom event is now used.

src/charm.py
dstathis: changed

sed-i:
The changes in this PR are already covered by #65. Closing.

sed-i:
taking over https://github.com/canonical/prometheus-operator/pull/65

lib/charms/alertmanager_k8s/v0/alertmanager.py
```diff
@@ -0,0 +1,175 @@
...
+
+import ops
+from ops.framework import StoredState, EventSource, EventBase
+from ops.relation import ConsumerBase, ProviderBase
+from ops.charm import CharmBase
+
+from typing import List
+import logging
+
+LIBID = "abcdef1234"  # Unique ID that refers to the library forever
```
balbirthomas, line 18: You will need to make another pull request when you finally set the correct LIBID. Since you are in the process of making a bundle. It may be a good idea to finalize your alertmanager library and then merge this PR.

balbirthomas (approved):

PR #51, opened 2021-06-10
Added github unit test running

balbirthomas:
Courtsey @dstathis . Thanks.

### DIFF
--- a/.github/workflows/run_tests.yaml
+++ b/.github/workflows/run_tests.yaml
@@ -1,0 +1,25 @@
+name: Run-Tests
+
+on: [pull_request]
+
+jobs:
+  run-tests:
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        python-version: [3.8]
+
+    steps:
+      - uses: actions/checkout@v2
+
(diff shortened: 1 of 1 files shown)


.github/workflows/run_tests.yaml
```diff
@@ -0,0 +1,25 @@
+name: Run-Tests
+
+on: [pull_request]
+
+jobs:
+  run-tests:
+    runs-on: ubuntu-latest
```
sed-i, line 7: consider using `ubuntu-20.04`
https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#jobsjob_idruns-on

sed-i (approved):

dstathis (approved):

.github/workflows/run_tests.yaml
dstathis: I would argue that if ubuntu version matters we have a bug and should fix it

balbirthomas: We can continue this conversation on MM. Resolving this conversation here for now so I can merge this. Will be happy to make any changes when there is a consensus.

PR #52, opened 2021-06-11
Add coverage to tests, add display-name for nicer display on Charmhub

mthaddon:
The charmcraft init template now includes coverage, update this charm to include that.

Also add a "display-name" in metadata.yaml so the name of the charm is correctly displayed on https://charmhub.io

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,6 +1,7 @@
 *~
 *swp
 *.charm
+.coverage
 __pycache__
 build
 venv
--- a/run_tests
+++ b/run_tests
@@ -13,5 +13,5 @@ else
 fi
 
 black --diff
-coverage run --branch --source=lib,src -m unittest -v "$@"
+coverage run --branch --source=src,lib/charms/prometheus_k8s -m unittest -v "$@"
 coverage report -m
--- a/run_tests
+++ b/run_tests
@@ -13,4 +13,5 @@ else
 fi
 
 black --diff
-python3 -m unittest -v "$@"
+coverage run --branch --source=lib,src -m unittest -v "$@"
+coverage report -m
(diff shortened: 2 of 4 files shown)


balbirthomas (dismissed):
Thanks

run_tests
```diff
@@ -13,4 +13,5 @@ else
 fi
 
 black --diff
-python3 -m unittest -v "$@"
+coverage run --branch --source=lib,src -m unittest -v "$@"
```
sed-i, line 16:
```suggestion
coverage run --branch --source=src,lib/charms/prometheus_k8s -m unittest -v "$@"
```
Should coverage exclude external libs?

balbirthomas: Sounds like a good idea. Particularly so, since unit tests for the charm library already exist :-) .

mthaddon: I think Leon's suggestion makes sense though. `lib/charms/prometheus_k8s` ensures you're only running tests for your own libraries, not for others you may import in the future. I'll commit that.

.gitignore
```diff
@@ -1,6 +1,7 @@
 *~
 *swp
 *.charm
+.coverage
```
dstathis, line 4: I think we want to have everyone using the same coverage settings so we should probably commit this file

dstathis (dismissed):

.gitignore
balbirthomas: This is an autogenerated coverage report. I do not see why it should be committed.

balbirthomas (approved):

balbirthomas (approved):

Abuelodelanada (approved):
Looks good to me

dstathis (approved):

.gitignore
dstathis: Ah I see. I thought this was .coveragerc

It's a little bit weird they would start the report file with "."

PR #61, opened 2021-06-23
Make the prometheus charm use a better Grafana interface

rbarry82:
Update the prometheus charm to use smarter data bags instead of dumping a raw dict into the relation

### DIFF
--- a/src/charm.py
+++ b/src/charm.py
@@ -128,7 +128,7 @@ class PrometheusCharm(CharmBase):
                 self.model.get_binding(event.relation).network.bind_address
             ),
             "port": str(self.model.config["port"]),
-            "source-type": "prometheus"
+            "source-type": "prometheus",
         }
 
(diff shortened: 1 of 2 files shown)


balbirthomas (approved):
LGTM. May I request you to add a line or two to INTEGRATION.md once it is merged.

balbirthomas (approved):

src/charm.py
```diff
@@ -123,11 +123,15 @@ def _on_grafana_changed(self, event):
...
-        )
+        source_data = {
+            "private-address": str(
+                self.model.get_binding(event.relation).network.bind_address
+            ),
+            "port": str(self.model.config["port"]),
+            "source-type": "prometheus",
+        }
+
+        event.relation.data[self.unit]["sources"] = json.dumps(source_data)
```
dstathis, line 134: I would consider "source" instead of "sources" since it is only one.

dstathis (approved):

PR #62, opened 2021-06-24
Add a property to get the current endpoints on the relation

dstathis:


rbarry82 (approved):
Looks ok to me. Where is this going to be used?

tests/test_consumer.py
```diff
@@ -56,3 +60,12 @@ def test_consumer_can_remove_an_endpoint(self):
         data = self.harness.get_relation_data(rel_id, self.harness.model.app.name)
         target = json.loads(data["targets"])
         self.assertFalse(target)
+
+    def test_consumer_can_get_endpoints(self):
```
balbirthomas, line 64: Hmm .... :-) In your docstring you do recognize that the endpoints method is useful "only in" multi mode but I do not see that the test specifically targets multi mode, unless I am being daft today. Oh well. I will accept this sanity test for now.

balbirthomas (approved):

tests/test_consumer.py
dstathis: It is not only useful in multi mode. I think you were reading the "rel_id" arg

PR #65, opened 2021-06-29
Feature/better alertmanager interface

dstathis:
Reopening Leon's pull request from my branch so I can make changes as necessary. Also added one commit to make it work right.

### DIFF
--- a/src/charm.py
+++ b/src/charm.py
@@ -10,10 +10,11 @@ import json
 from ops.charm import CharmBase
 from ops.framework import StoredState
 from ops.main import main
-from ops.model import ActiveStatus, MaintenanceStatus
+from ops.model import ActiveStatus, MaintenanceStatus, BlockedStatus
 from ops.pebble import ConnectionError
 from prometheus_server import Prometheus
 from charms.prometheus_k8s.v1.prometheus import PrometheusProvider
+from charms.alertmanager_k8s.v0.alertmanager import AlertmanagerConsumer
 
 PROMETHEUS_CONFIG = "/etc/prometheus/prometheus.yml"
 logger = logging.getLogger(__name__)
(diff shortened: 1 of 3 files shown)


src/charm.py
```diff
@@ -133,33 +134,8 @@ def _on_grafana_changed(self, event):
...
-
-        When an Alertmanager departs it is removed from the list
-        of currently available Alertmanagers, the Prometheus configuration
-        is updated and Prometheus is restarted.
-        """
-        if not self.unit.is_leader():
-            return
-        self._stored.alertmanagers.clear()
+    def _on_alertmanager_cluster_changed(self, event):
+        self._stored.alertmanagers = self.alertmanager_lib.get_cluster_info()
```
balbirthomas, line 138: Now that you have `get_cluster_info()` it is no longer necessary to store this in two places that is in the Prometheus charm and the `AlertmanagerConsumer`. It should be possible to just invoke `get_cluster_info()` from within `_alerting_config()` method of the Prometheus charm.

sed-i:
This change works well together with https://github.com/canonical/alertmanager-operator/pull/13

rbarry82:
@sed-i Development paradigms have changed a lot in the last 3 weeks. Are we sure we want the consumer emitting events instead of updating the databag and relying on `relation_changed`?

sed-i:
> @sed-i Development paradigms have changed a lot in the last 3 weeks. Are we sure we want the consumer emitting events instead of updating the databag and relying on `relation_changed`?

The use of custom events and StoredState is the most simple and concise solution I found to the challenges encountered. 
Relying on `relation_changed` being handled by the consumer charm means that implementation detail of alertmanager-operator side would be implicitly coded in prometheus. I would rather abstract it away with interfaces using the alertmanager library.

Can we take a deeper look into this via issues against alertmanager?

sed-i:
moving back to https://github.com/canonical/prometheus-operator/pull/48 and closing this one

PR #67, opened 2021-07-01
Use a Grafana consumer library instead of raw relation data 

rbarry82:
Add the Grafana consumer library and use that to set data on the data bucket instead of doing it directly. Update the tests to match

### DIFF
--- a/lib/charms/grafana_k8s/v0/grafana_source.py
+++ b/lib/charms/grafana_k8s/v0/grafana_source.py
@@ -43,11 +43,11 @@ class GrafanaSourcesChanged(EventBase):
         super().__init__(handle)
         self.data = data
 
-    def snapshot(self):
+    def snapshot(self) -> Dict:
         """Save grafana source information"""
         return {"data": self.data}
 
-    def restore(self, snapshot):
+    def restore(self, snapshot) -> None:
         """Restore grafana source information"""
         self.data = snapshot["data"]
 
@@ -69,12 +69,12 @@ def _validate(self, source: SourceData) -> dict:
         source: A :dict: representing the source data
     """
 
-    def source_name_in_use(source_name):
+    def source_name_in_use(source_name) -> bool:
         return any(
             [s["source-name"] == source_name for s in self._stored.sources.values()]
         )
 
-    def check_required_fields(source_data: dict):
+    def check_required_fields(source_data: Dict) -> Dict:
         # dictionary of all the required/optional datasource field values
         # using this as a more generic way of getting data source fields
         validated_source = {
@@ -113,7 +113,7 @@ def _validate(self, source: SourceData) -> dict:
 
         return validated_source
 
-    def set_defaults(validated_source):
+    def set_defaults(validated_source) -> Dict:
         # specifically handle optional fields if necessary
         # check if source-name was not passed or if we have already saved the provided name
         if validated_source.get("source-name") is None or source_name_in_use(
@@ -153,15 +153,17 @@ class GrafanaSourceConsumer(ConsumerBase):
     ) -> None:
         """Construct a Grafana charm client.
 
-        The :class:`GrafanaConsumer` object provides an interface
+        The :class:`GrafanaSourceConsumer` object provides an interface
         to Grafana. This interface supports providing additional
         sources for Grafana to monitor. For example, if a charm
         exposes some metrics which are consumable by a dashboard
         (such as Prometheus), then an additional source can be added
-        by instantiating a :class:`GrafanaConsumer` object and
+        by instantiating a :class:`GrafanaSourceConsumer` object and
         adding its datasources as follows:
 
-            self.grafana = GrafanaConsumer(self, "grafana-source", {"grafana-source"}: ">=2.0"})
+            self.grafana = GrafanaSourceConsumer(
+                self, "grafana-source", {"grafana-source"}: ">=2.0"}
+            )
             self.grafana.add_source({
                 "source-type": <source-type>,
                 "address": <address>,
@@ -171,7 +173,7 @@ class GrafanaSourceConsumer(ConsumerBase):
         Args:
 
             charm: a :class:`CharmBase` object which manages this
-                :class:`GrafanaConsumer` object. Generally this is
+                :class:`GrafanaSourceConsumer` object. Generally this is
                 `self` in the instantiating class.
             name: a :string: name of the relation between `charm`
                 the Grafana charmed service.
@@ -209,7 +211,7 @@ class GrafanaSourceConsumer(ConsumerBase):
                 }
             rel_id: an optional integer specifying the relation ID
                 for the grafana source service, only required if the
-                :class:`GrafanaConsumer` has been instantiated in
+                :class:`GrafanaSourceConsumer` has been instantiated in
                 `multi` mode.
 
         """
@@ -239,7 +241,7 @@ class GrafanaSourceConsumer(ConsumerBase):
         Args:
             rel_id: an optional integer specifying the relation ID
                 for the grafana-k8s source service, only required if the
-                :class:`GrafanaConsumer` has been instantiated in
+                :class:`GrafanaSourceConsumer` has been instantiated in
                 `multi` mode.
         """
         rel = self.framework.model.get_relation(self.name, rel_id)
@@ -347,11 +349,13 @@ class GrafanaSourceProvider(ProviderBase):
             else None
         )
         if not data:
-            logger.debug("No data in grafana-source relation")
+            logger.info("NO DATA")
             return
 
-        # This is due to a double fire on relation_joined
-        # calling the consumer, which calls relation_changed
+        # This is due to a double fire on relation_joined calling
+        # the consumer, which calls relation_changed, and the actual data may be
+        # the same, but we'll see the source name in use later, try to set it again,
+        # and get double relation names
         if (
             self._stored.sources.get(rel.id, None)
             and len(set(data.items()) ^ set(self._stored.sources[rel.id].items())) == 0
@@ -367,6 +371,8 @@ class GrafanaSourceProvider(ProviderBase):
                 f"Missing data on added grafana-k8s source {e}", exc_info=True
             )
             return
+
+        logger.info("ADDING SOURCE: {}".format(data))
 
         self._stored.sources[rel.id] = data
         self.on.sources_changed.emit()
@@ -391,7 +397,7 @@ class GrafanaSourceProvider(ProviderBase):
         broken.
         """
 
-        logger.info("Removing all grafana-source data for relation: {}".format(rel_id))
+        logger.info("Removing all data for relation: {}".format(rel_id))
 
         try:
             removed_source = self._stored.sources.pop(rel_id, None)
--- a/lib/charms/grafana_k8s/v0/grafana_source.py
+++ b/lib/charms/grafana_k8s/v0/grafana_source.py
@@ -1,0 +1,427 @@
+import json
+import logging
+
+from collections import namedtuple
+from ops.charm import (
+    CharmBase,
+    CharmEvents,
+    RelationBrokenEvent,
+    RelationChangedEvent,
+    RelationJoinedEvent,
+)
+from ops.framework import EventBase, EventSource, StoredState
+from ops.relation import ConsumerBase, ProviderBase
+from typing import Dict, List
+
+
+LIBID = "987654321"
+LIBAPI = 1
+LIBPATCH = 0
+
+logger = logging.getLogger(__name__)
+
+SourceData = namedtuple("SourceData", "name unit app rel_id data")
+REQUIRED_DATASOURCE_FIELDS = {
+    "address",  # the hostname/IP of the data source server
+    "port",  # the port of the data source server
+    "source-type",  # the data source type (e.g. prometheus)
+}
+
+OPTIONAL_DATASOURCE_FIELDS = {
+    "source-name",  # a human-readable name of the source
+}
+
+
+class SourceFieldsMissingError(Exception):
+    pass
+
+
+class GrafanaSourcesChanged(EventBase):
+    """Event emitted when Grafana sources change"""
+
+    def __init__(self, handle, data=None):
+        super().__init__(handle)
+        self.data = data
+
+    def snapshot(self):
+        """Save grafana source information"""
+        return {"data": self.data}
+
+    def restore(self, snapshot):
+        """Restore grafana source information"""
+        self.data = snapshot["data"]
+
+
+class GrafanaSourceEvents(CharmEvents):
+    """Events raised by :class:`GrafanaSourceEvents`"""
+
+    sources_changed = EventSource(GrafanaSourcesChanged)
+    sources_to_delete_changed = EventSource(GrafanaSourcesChanged)
+
+
+def _validate(self, source: SourceData) -> dict:
+    """Check whether given source data is valid. If it is missing optional
+    fields only, those are set from defaults. If it is missing required fields,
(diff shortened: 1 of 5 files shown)


lib/charms/grafana_k8s/v0/grafana_source.py
```diff
@@ -0,0 +1,433 @@
...
+        data = (
+            json.loads(event.relation.data[event.app].get("sources", {}))
+            if event.relation.data[event.app].get("sources", {})
+            else None
+        )
+        if not data:
+            logger.info("NO DATA")
+            return
+
+        # This is due to a double fire on relation_joined calling
```
balbirthomas, line 355: I am surprised. I was given to understand that if relation data is unchanged then Juju does not fire a relation changed event. Is this wrong ? If so could it be a Juju bug ? If it may be a bug, then a bug ticket link here would help in future maintenance of this lines.

lib/charms/grafana_k8s/v0/grafana_source.py
```diff
@@ -0,0 +1,433 @@
...
+
+        self._stored.sources[rel.id] = data
+        self.on.sources_changed.emit()
+
+    def on_grafana_source_relation_broken(self, event: RelationBrokenEvent) -> None:
+        """Update job config when consumers depart.
+
+        When a Grafana source consumer departs, the configuration
+        for that consumer is removed from the list of sources jobs,
+        added to a list of sources to remove, and other consumers
```
balbirthomas, line 385: Since this is the Provider class, shouldn't this read "and the provider is informed through a" rather than "and other consumers are informed through a" ?

lib/charms/grafana_k8s/v0/grafana_source.py
```diff
@@ -0,0 +1,433 @@
...
+    def list_sources(self) -> List[dict]:
+        """Returns an array of currently valid sources"""
+        sources = []
+        for source in self._stored.sources.values():
+            sources.append(source)
+
+        return sources
+
+    @property
+    def removed_source_names(self) -> List[str]:
```
balbirthomas, line 266: It is not obvious how and why this method would be used by the Consumer charm ?

balbirthomas (dismissed):

lib/charms/grafana_k8s/v0/grafana_source.py
rbarry82: I may not be reading their repo right, but if I am, there is no DeepEquals against the map[string]interface{} parts before updating. I'll ask in discourse

lib/charms/grafana_k8s/v0/grafana_source.py
rbarry82: Typo. I'll resolve.

lib/charms/grafana_k8s/v0/grafana_source.py
rbarry82: Grafana's method of removal is exceptionally stupid. See https://grafana.com/docs/grafana/latest/administration/provisioning/ and look for "deleteDatasources". Any consumer which may want to programatically update their own list or check to see if it's there (Grafana Dashboard being a good example, but there are others) may want to know

lib/charms/grafana_k8s/v0/grafana_source.py
```diff
@@ -0,0 +1,433 @@
...
+    RelationChangedEvent,
+    RelationJoinedEvent,
+)
+from ops.framework import EventBase, EventSource, StoredState
+from ops.relation import ConsumerBase, ProviderBase
+from typing import Dict, List
+
+
+LIBID = "987654321"
+LIBAPI = 1
```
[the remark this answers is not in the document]
rbarry82: Didn't we agree on 1? This is also what the Prometheus charm itself is using:
https://github.com/canonical/prometheus-operator/blob/76f0d6cc56b78ef70b32b4dc573ff94f330dc08e/lib/charms/prometheus_k8s/v1/prometheus.py#L162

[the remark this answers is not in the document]
balbirthomas: I do not know the right answer here. The Prometheus charm may also need to be corrected.

balbirthomas (dismissed):

lib/charms/grafana_k8s/v1/grafana_source.py
```diff
@@ -0,0 +1,336 @@
...
+                this object should support interacting with multiple
+                service providers.
+
+        """
+        super().__init__(charm, name, consumes, multi)
+
+        self.charm = charm
+
+        self._stored.set_default(sources=dict())  # available data sources
+        self._stored.set_default(sources_to_delete=set())
```
sed-i, line 102: set_default takes `**kwargs`

lib/charms/grafana_k8s/v1/grafana_source.py
```diff
@@ -0,0 +1,336 @@
...
+            self._stored.sources[rel_id]
+        )
+
+    def _update_sources(self, data: Dict, rel_id: int) -> None:
+        """Updates the event data bucket with new data"""
+        self._stored.sources[rel_id] = data
+        rel = self.framework.model.get_relation(self.name, rel_id)
+
+        logger.debug("Updating Grafana source data to {}".format(data))
+        rel.data[self.charm.app]["sources"] = json.dumps(data)
```
sed-i, line 195: I wonder why this is stored in app rather than unit data bag.
On leadership change, there is a considerable time without a leader so sources may be lost if updated during that time.

lib/charms/grafana_k8s/v1/grafana_source.py
rbarry82: Sure, but I'm not sure it offers any clarity/advantages for two variables.

lib/charms/grafana_k8s/v1/grafana_source.py
rbarry82: So it's available to all Grafana instances, which are "shared nothing". In the case of a leadership change when we are "really" HA, the same concern is possible, because it will require a database backend, and only the leader should be updating the database.

lib/charms/grafana_k8s/v1/grafana_source.py
```diff
@@ -0,0 +1,336 @@
...
+            source_type: a :str: indicating the source type. Defaults to "prometheus"
+            source_name: a :str: specifying the source name. Otherwise this is calculated
+            rel_id: an optional integer specifying the relation ID
+                for the grafana source service, only required if the
+                :class:`GrafanaSourceConsumer` has been instantiated in
+                `multi` mode.
+
+        """
+        rel = self.framework.model.get_relation(self.name, rel_id)
+        rel_id = rel_id if rel_id is not None else rel.id
```
sed-i, line 123: If the caller did not provide `rel_id`, this will work only if there is only one relation in place.
Instead, shouldn't this iterate over `self.model.relations`?

lib/charms/grafana_k8s/v1/grafana_source.py
```diff
@@ -0,0 +1,336 @@
...
+        self._stored.sources[rel_id] = data
+        rel = self.framework.model.get_relation(self.name, rel_id)
+
+        logger.debug("Updating Grafana source data to {}".format(data))
+        rel.data[self.charm.app]["sources"] = json.dumps(data)
+
+    def _source_name_in_use(self, source_name) -> bool:
+        return any(
+            [s["source-name"] == source_name for s in self._stored.sources.values()]
+        )
```
sed-i, line 200: iiuc, in this notation the evaluation would not be lazy.
maybe get rid of the linear search altogether :)

lib/charms/grafana_k8s/v1/grafana_source.py
```diff
@@ -0,0 +1,336 @@
...
+    def remove_source(self, rel_id=None) -> None:
+        """Removes a source relation.
+
+        Args:
+            rel_id: an optional integer specifying the relation ID
+                for the grafana-k8s source service, only required if the
+                :class:`GrafanaSourceConsumer` has been instantiated in
+                `multi` mode.
+        """
+        rel = self.framework.model.get_relation(self.name, rel_id)
```
sed-i, line 155: will be None if no relations in place

lib/charms/grafana_k8s/v1/grafana_source.py
```diff
@@ -0,0 +1,336 @@
...
+
+        if rel_id is None:
+            rel_id = rel.id
+
+        source = self._stored.sources.pop(rel_id)
```
sed-i, line 160: should probably be `.pop(rel_id, None)`

lib/charms/grafana_k8s/v1/grafana_source.py
```diff
@@ -0,0 +1,336 @@
...
+        if source is not None:
+            self._stored.sources_to_delete.add(source["source-name"])
+
+        self._update_sources({}, rel_id)
+
+    def list_sources(self) -> List[dict]:
+        """Returns an array of currently valid sources"""
+        sources = []
+        for source in self._stored.sources.values():
+            sources.append(source)
```
sed-i, line 171: list comprehension?

lib/charms/grafana_k8s/v1/grafana_source.py
```diff
@@ -0,0 +1,336 @@
...
+        and update the relation data bag if we know about anything
+        """
+        rel_id = event.relation.id
+        if not self._stored.sources.get(rel_id, {}):
+            return
+
+        logger.debug("Setting Grafana data sources: %s", self._stored.sources[rel_id])
+        event.relation.data[self.charm.app]["sources"] = json.dumps(
+            self._stored.sources[rel_id]
+        )
```
sed-i, line 187: `is_leader` guard missing

lib/charms/grafana_k8s/v1/grafana_source.py
sed-i: `is_leader` guard missing

lib/charms/grafana_k8s/v1/grafana_source.py
```diff
@@ -0,0 +1,336 @@
...
+            [s["source-name"] == source_name for s in self._stored.sources.values()]
+        )
+
+
+class GrafanaSourceProvider(ProviderBase):
+    on = GrafanaSourceEvents()
+    _stored = StoredState()
+
+    def __init__(self, charm: CharmBase, name: str, service: str, version=None) -> None:
+        """A Grafana based Monitoring service consumer
```
sed-i, line 208: typo: consumer -> provider

lib/charms/grafana_k8s/v1/grafana_source.py
```diff
@@ -0,0 +1,336 @@
...
+            version: a string providing the semantic version of the Grafana
+                source being provided.
+
+        """
+        super().__init__(charm, name, service, version)
+        self.charm = charm
+        events = self.charm.on[name]
+
+        self._stored.set_default(sources=dict())  # available data sources
+        self._stored.set_default(sources_to_delete=set())
```
sed-i, line 229: set_default takes `**kwargs`

lib/charms/grafana_k8s/v1/grafana_source.py
```diff
@@ -0,0 +1,336 @@
...
+
+        rel = event.relation
+        data = (
+            json.loads(event.relation.data[event.app].get("sources", {}))
+            if event.relation.data[event.app].get("sources", {})
+            else None
+        )
+        if not data:
+            logger.debug("No grafana source data in relation data bag")
+            return
```
sed-i, line 259:
```suggestion
        if as_json := event.relation.data[event.app].get("sources", {}):
            data = json.loads(as_json)
        else:
            logger.debug("No grafana source data in relation data bag")
            return
```

lib/charms/grafana_k8s/v1/grafana_source.py
```diff
@@ -0,0 +1,336 @@
...
+
+        # This is due to a double fire on relation_joined calling
+        # the consumer, which calls relation_changed, and the actual data may be
+        # the same, but we'll see the source name in use later, try to set it again,
+        # and get double relation names
+        if (
+            self._stored.sources.get(rel.id, None)
+            and len(set(data.items()) ^ set(self._stored.sources[rel.id].items())) == 0
```
sed-i, line 267:
```suggestion
            and set(data.items()) ^ set(self._stored.sources[rel.id].items()) == set()
```

lib/charms/grafana_k8s/v1/grafana_source.py
```diff
@@ -0,0 +1,336 @@
...
+        """Remove the grafana-source from the datastore. and add the
+        name to the list of sources to remove when a relation is
+        broken.
+        """
+
+        logger.debug("Removing all data for relation: {}".format(rel_id))
+
+        try:
+            removed_source = self._stored.sources.pop(rel_id, None)
+            if removed_source:
```
sed-i, line 306: walrus?

lib/charms/grafana_k8s/v1/grafana_source.py
rbarry82: This will throw an exception from the framework if possible. But this is a consumer library. How many `grafana-source` relations do you think it's going to have?

If it needs more than one, that will be in the Charm, and the author will see exceptions/know to pass the ID specific to that relation.

lib/charms/grafana_k8s/v1/grafana_source.py
rbarry82: Correct, it's not lazy, but converting something which is extremely Pythonic to a generator expression over a very small problem space is a premature optimization which harms readability badly

lib/charms/grafana_k8s/v1/grafana_source.py
rbarry82: In which case the consumer will never be ready.

lib/charms/grafana_k8s/v1/grafana_source.py
sed-i: an innocent user experimenting with two grafanas at the same time.
nobody wants to touch relation ids, even with a pole.

lib/charms/grafana_k8s/v1/grafana_source.py
sed-i: how about after remove-relation?
you mean you have `.ready` and `.unready` guards?

rbarry82: This code has already been merged into `grafana-operator`. With approvals. Including you.

There is no reason we would expect the user of a consumer library to try to invoke it after the relationship is broken. And if they do, an exception is _expected_ and good practice Python. Exceptions are not nearly as forbidden in Python as other languages

sed-i: Apologies for entering into nitpick land here, but:
The exception would be coming from your new code, not library code, i.e. you'd be introducing a new exception which is not related to the problem: `AttributeError: 'NoneType' object has no attribute 'id'`, which seems like a coding mistake rather than exception propagation.

"There is no reason we would expect the user [...] to try to invoke it" is acceptable though.

sed-i (approved):

lib/charms/grafana_k8s/v1/grafana_source.py
rbarry82: Correct (except this is a consumer library, which is why I called it library bode). "Something very bad happened and your expected workflow did not complete. You need to take a look." is the role of exceptions being thrown. It is the `5xx` return code of "you did something wrong" in Python if it were HTTP, rather than the `4xx` `logging.error` of "you did everything right, but something bad happened when we tried to talk to that webserver/synchronize data".

balbirthomas (approved):

PR #69, opened 2021-07-05
namespace interfaces with underscore

### LINKED ISSUE #44: Reconsider alertmanager interface name
`alertmanager` is not a general relation name, it is more the `interface` of the relation. We need to find a more general name and use it as standard name for relations like "we push alerts".

sed-i:
Resolves #44

### DIFF
--- a/metadata.yaml
+++ b/metadata.yaml
@@ -33,7 +33,7 @@ requires:
     grafana-source:
         interface: grafana_datasource
     alertmanager:
-        interface: alertmanager
+        interface: alertmanager_dispatch
 storage:
   database:
     type: filesystem


metadata.yaml
```diff
@@ -33,7 +33,7 @@ requires:
     grafana-source:
         interface: grafana_datasource
     alertmanager:
-        interface: alertmanager
+        interface: alertmanager_dispatch
```
balbirthomas, line 36: It will be very helpful if you can add a commit message with the rationale underlying this choice of name.

balbirthomas (approved):

metadata.yaml
sed-i: You mean in addition to the discussion in the linked issue #44?

balbirthomas: If you prefer you can just link to the issue. But a few lines copied and pasted  from there would also be be nice if you think git commit history is a valuable resource for future developers.

sed-i: Done:
https://github.com/canonical/alertmanager-operator/pull/14/commits/29514ca066294ef62cfcb5b28f58240a6d3fecf6

mmanciop (approved):

mmanciop (approved):

rbarry82 (approved):

PR #70, opened 2021-07-07
Remove spurious quotes when relations are broken

rbarry82:
There is no key named "rel_id", and this will always result in a KeyError. Remove the quotes so lookups work as expected


balbirthomas (dismissed):
Approving. But please be aware all this code is going to go away with work in progress "juju-topology" personal branch in order to set unit specific labels. This is a breaking change in the as yet unreleased Prometheus charm library.

sed-i (approved):

sed-i (approved):

dstathis (approved):

PR #71, opened 2021-07-12
Juju topology

balbirthomas:
This PR fundamentally changes the structure of the Prometheus charm library. This is a breaking change because the two methods `add_endpoint()` and `remove_endpoint()` have been removed. Instead the `PrometheusProvider` now scrapes each unit of the `PrometheusConsumer` without being told explicitly which are the scrape endpoints. However it is still possible to configure the metrics path and port for each consumer charm. Future PR's will elaborate further on this configurability. 

These changes were necessitated because of requirement to provide per unit scrape labels. In particular the requirement to provide a different unit name label for each consumer unit.

### DIFF
--- a/lib/charms/prometheus_k8s/v1/prometheus.py
+++ b/lib/charms/prometheus_k8s/v1/prometheus.py
@@ -249,6 +249,9 @@ class PrometheusProvider(ProviderBase):
         Returns:
             A static scrape configuration for a specific relation.
         """
+        if len(relation.units) == 0:
+            return {}
+
         scrape_metadata = json.loads(
             relation.data[relation.app].get("prometheus_scrape_metadata")
         )
--- a/lib/charms/prometheus_k8s/v1/prometheus.py
+++ b/lib/charms/prometheus_k8s/v1/prometheus.py
@@ -377,27 +377,27 @@ class PrometheusProvider(ProviderBase):
         if len(relation.units) == 0:
             return []
 
-        scrape_jobs = json.loads(
-            relation.data[relation.app].get("scrape_jobs")
-        )
+        scrape_jobs = json.loads(relation.data[relation.app].get("scrape_jobs"))
 
         if not scrape_jobs:
             return []
 
-        scrape_metadata = json.loads(
-            relation.data[relation.app].get("scrape_metadata")
-        )
+        scrape_metadata = json.loads(relation.data[relation.app].get("scrape_metadata"))
 
         job_name_prefix = "juju_{}_{}_{}_prometheus_{}_scrape".format(
-            scrape_metadata["model"], scrape_metadata["model_uuid"][:7],
-            scrape_metadata["application"], relation.id
+            scrape_metadata["model"],
+            scrape_metadata["model_uuid"][:7],
+            scrape_metadata["application"],
+            relation.id,
         )
 
         hosts = self._relation_hosts(relation)
 
         labeled_job_configs = []
         for job in scrape_jobs:
-            config = self._labeled_static_job_config(job, job_name_prefix, hosts, scrape_metadata)
+            config = self._labeled_static_job_config(
+                job, job_name_prefix, hosts, scrape_metadata
+            )
             labeled_job_configs.append(config)
 
         return labeled_job_configs
@@ -467,13 +467,15 @@ class PrometheusProvider(ProviderBase):
                     unitless_targets.append(target)
 
             if unitless_targets:
-                unitless_config = self._labeled_unitless_config(unitless_targets,
-                                                                labels, scrape_metadata)
+                unitless_config = self._labeled_unitless_config(
+                    unitless_targets, labels, scrape_metadata
+                )
                 config["static_configs"].append(unitless_config)
 
             for host_name, host_address in hosts.items():
-                static_config = self._labeled_unit_config(host_name, host_address,
-                                                          ports, labels, scrape_metadata)
+                static_config = self._labeled_unit_config(
+                    host_name, host_address, ports, labels, scrape_metadata
+                )
                 config["static_configs"].append(static_config)
 
         return config
@@ -520,13 +522,12 @@ class PrometheusProvider(ProviderBase):
             for a list of fully qualified hosts.
         """
         juju_labels = self._set_juju_labels(labels, scrape_metadata)
-        unitless_config = {
-            "targets": targets,
-            "labels": juju_labels
-        }
+        unitless_config = {"targets": targets, "labels": juju_labels}
         return unitless_config
 
-    def _labeled_unit_config(self, host_name, host_address, ports, labels, scrape_metadata):
+    def _labeled_unit_config(
+        self, host_name, host_address, ports, labels, scrape_metadata
+    ):
         """Static scrape configuration for a wildcard host.
 
         Wildcard hosts are those scrape targets whose address is
@@ -607,9 +608,7 @@ class PrometheusConsumer(ConsumerBase):
 
         events = self._charm.on[self._relation_name]
         self.framework.observe(events.relation_joined, self._set_scrape_metadata)
-        self.framework.observe(
-            self._service_event, self._set_unit_ip
-        )
+        self.framework.observe(self._service_event, self._set_unit_ip)
 
     def _set_scrape_metadata(self, event):
         """Ensure scrape targets metadata is made available to Prometheus.
@@ -654,11 +653,7 @@ class PrometheusConsumer(ConsumerBase):
            A list of dictionaries, where each dictionary specifies a
            single scrape job for Prometheus.
         """
-        default_job = [
-            {
-                "metrics_path": "/metrics"
-            }
-        ]
+        default_job = [{"metrics_path": "/metrics"}]
         return self._jobs if self._jobs else default_job
 
     @property
--- a/lib/charms/prometheus_k8s/v1/prometheus.py
+++ b/lib/charms/prometheus_k8s/v1/prometheus.py
@@ -246,15 +246,14 @@ class PrometheusProvider(ProviderBase):
             A static scrape configuration for a specific relation.
         """
         scrape_metadata = json.loads(
-            relation.data[relation.app].get("prometheus_scrape_metadata"))
+            relation.data[relation.app].get("prometheus_scrape_metadata")
+        )
 
         if not scrape_metadata:
             return None
 
         job_name = "juju_prometheus_scrape_{}_{}_{}".format(
-            scrape_metadata["model"],
-            scrape_metadata["application"],
-            relation.id
+            scrape_metadata["model"], scrape_metadata["application"], relation.id
         )
 
         hosts = {}
@@ -267,20 +266,21 @@ class PrometheusProvider(ProviderBase):
         scrape_config = {
             "job_name": job_name,
             "metrics_path": scrape_metadata["static_scrape_path"],
-            "static_configs": []
+            "static_configs": [],
         }
 
         for host_name, host_address in hosts.items():
-            metrics_url = "{}:{}".format(host_address,
-                                         scrape_metadata["static_scrape_port"])
+            metrics_url = "{}:{}".format(
+                host_address, scrape_metadata["static_scrape_port"]
+            )
 
             config = {
                 "targets": [metrics_url],
                 "labels": {
                     "juju_model": "{}".format(scrape_metadata["model"]),
                     "juju_application": "{}".format(scrape_metadata["application"]),
-                    "juju_unit": "{}".format(host_name)
-                }
+                    "juju_unit": "{}".format(host_name),
+                },
             }
             scrape_config["static_configs"].append(config)
 
@@ -288,7 +288,6 @@ class PrometheusProvider(ProviderBase):
 
 
 class PrometheusConsumer(ConsumerBase):
-
     def __init__(self, charm, name, consumes, service, config={}, multi=False):
         """Construct a Prometheus charm client.
 
@@ -329,8 +328,9 @@ class PrometheusConsumer(ConsumerBase):
 
         events = self._charm.on[self._relation_name]
         self.framework.observe(events.relation_joined, self._set_scrape_metadata)
-        self.framework.observe(self._charm.on[self._service].pebble_ready,
-                               self._set_unit_ip)
+        self.framework.observe(
+            self._charm.on[self._service].pebble_ready, self._set_unit_ip
+        )
 
     def _set_scrape_metadata(self, event):
         """Ensure scrape targets metadata is made available to Prometheus.
@@ -342,13 +342,15 @@ class PrometheusConsumer(ConsumerBase):
         host address in Juju unit relation data.
         """
         event.relation.data[self._charm.unit]["prometheus_scrape_host"] = str(
-            self._charm.model.get_binding(event.relation).network.bind_address)
+            self._charm.model.get_binding(event.relation).network.bind_address
+        )
 
         if not self._charm.unit.is_leader():
             return
 
-        event.relation.data[self._charm.app][
-            "prometheus_scrape_metadata"] = json.dumps(self._scrape_metadata)
+        event.relation.data[self._charm.app]["prometheus_scrape_metadata"] = json.dumps(
+            self._scrape_metadata
+        )
 
     def _set_unit_ip(self, event):
         """Set unit host address
@@ -358,7 +360,8 @@ class PrometheusConsumer(ConsumerBase):
         """
         for relation in self._charm.model.relations[self._relation_name]:
             relation.data[self._charm.unit]["prometheus_scrape_host"] = str(
-                self._charm.model.get_binding(relation).network.bind_address)
+                self._charm.model.get_binding(relation).network.bind_address
+            )
 
     @property
     def _scrape_metadata(self):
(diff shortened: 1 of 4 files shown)


lib/charms/prometheus_k8s/v1/prometheus.py
```diff
@@ -317,99 +314,71 @@ def __init__(self, charm, name, consumes, multi=False):
...
-                provider. This is only necessary if the
-                `PrometheusConsumer` has been instantiated in
-                `multi` mode.
+        Each time a consumer charm container is restarted it updates its own
+        host address in the unit relation data for the Prometheus provider.
         """
+        for relation in self._charm.model.relations[self._relation_name]:
+            relation.data[self._charm.unit]["prometheus_scrape_host"] = str(
+                self._charm.model.get_binding(relation).network.bind_address
+            )
```
johnsca, line 369: It might be worth reaching out to someone from the Juju team familiar with networking but I worry a bit that, particularly for CMR, there's a chance that the address that the Prometheus side needs to connect to could be different than the address the target side should bind to. My understanding is that this edge case is why `ingress-address` and `bind-address` are two separate fields.

lib/charms/prometheus_k8s/v1/prometheus.py
```diff
@@ -317,99 +314,71 @@ def __init__(self, charm, name, consumes, multi=False):
...
+        """Generate scrape metadata.
 
-        logger.debug("Updating scrape targets to : %s", targets)
-        rel.data[self._charm.app]["targets"] = json.dumps(list(targets))
+        Returns:
+            Scrape configutation metadata for this Prometheus consumer charm.
+        """
+        metadata = {
+            "model": "{}".format(self._charm.model.name),
+            "application": "{}".format(self._charm.model.app.name),
```
johnsca, line 380: Just be aware that, with CMR, the model and application name could be the same for multiple related applications, but that should be fine with how it's handled because the relation ID is still guaranteed to be unique.

lib/charms/prometheus_k8s/v1/prometheus.py
```diff
@@ -317,99 +314,71 @@ def __init__(self, charm, name, consumes, multi=False):
...
+        self._service = service
         self._relation_name = name
-        self._stored.set_default(targets={})
+        self._static_scrape_port = config.get("static_scrape_port", 80)
+        self._static_scrape_path = config.get("static_scrape_path", "/metrics")
+        self._multi_mode = multi
+
         events = self._charm.on[self._relation_name]
-        self.framework.observe(events.relation_joined, self._set_targets)
+        self.framework.observe(events.relation_joined, self._set_scrape_metadata)
```
johnsca, line 335: I'm pretty sure this could be done on `events.relation_created` which would get the info over to the other side a little bit sooner.  Not a big deal, though.

johnsca (commented):
A few comments, mostly just things to keep in mind / be aware of, but overall seems fine to me.

lib/charms/prometheus_k8s/v1/prometheus.py
```diff
@@ -317,99 +314,71 @@ def __init__(self, charm, name, consumes, multi=False):
...
-
         """
         super().__init__(charm, name, consumes, multi)
+
         self._charm = charm
+        self._service = service
         self._relation_name = name
-        self._stored.set_default(targets={})
+        self._static_scrape_port = config.get("static_scrape_port", 80)
+        self._static_scrape_path = config.get("static_scrape_path", "/metrics")
```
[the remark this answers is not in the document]
balbirthomas: Thanks for pointing this out. Give me some time to digest this. Kube state metrics operator is an important one so I will study it and reach out to you, so we can explore what our options are and what works best. If required I will consider further refactors in subsequent commits.

lib/charms/prometheus_k8s/v1/prometheus.py
balbirthomas: My understanding was `ingress-address` was going to be deprecated. Thanks again for pointing this out. I will get it clarified by Juju/Operator teams.

lib/charms/prometheus_k8s/v1/prometheus.py
balbirthomas: There is actually a commit in this PR that intentionally removes relation ID from the list of scrape target labels. This was done on request of @mmanciop because there was concern here that Alertmanager will not be able to dedupe alerts in the presence of relation ID. I think this is a valid concern. A future PR will add model UUID once https://github.com/canonical/operator/pull/563 gets merged.

lib/charms/prometheus_k8s/v1/prometheus.py
balbirthomas: Thanks. Interesting. Had not thought of that. Will investigate. And perhaps add it into next PR or this one.

lib/charms/prometheus_k8s/v1/prometheus.py
rbarry82: It's merged now.

lib/charms/prometheus_k8s/v1/prometheus.py
rbarry82: Other way around -- `private-address` is deprecated, leaving `ingress-address` only. See https://github.com/juju/juju/blob/4ae860b247b4f3e821aa333ff4bf85af1d20b9d0/apiserver/facades/agent/uniter/uniter.go#L1389.

In any case, my hope for the concern raised by @johnsca is that Juju/OF will be "smart" enough to give you the right address in context for the relation

lib/charms/prometheus_k8s/v1/prometheus.py
```diff
@@ -208,66 +189,37 @@ def __init__(self, charm, name, service, version=None):
...
-        if not self._charm.unit.is_leader():
-            return
-
         rel_id = event.relation.id
-        try:
-            del self._stored.jobs[rel_id]
-            self.on.targets_changed.emit()
-        except KeyError:
-            pass
+        self.on.targets_changed.emit(relation_id=rel_id)
```
rbarry82, line 222: This is risky, for the same reason as the Karma charm from @sed-i. Even after `RelationDeparted` or `RelationBroken`, the data bag still exists. I would suspect that, even after `RelationDeparted` triggers this event, the list of jobs populated when the charm calls `provider.jobs()` will include this despite it being a broken relation (which has not finished departing yet)

rbarry82 (commented):
Looks great in general, but I do have a concern around `RelationDeparted`

lib/charms/prometheus_k8s/v1/prometheus.py
balbirthomas: Thanks. I am not seeing this behaviour. When a unit is removed Prometheus dashboard does reflect this change. Will test again.

lib/charms/prometheus_k8s/v1/prometheus.py
balbirthomas: Thanks. Added commit with model uuid into this PR.

lib/charms/prometheus_k8s/v1/prometheus.py
balbirthomas: On further thought I actually need the relation joined because I am setting some unit data also. Leaving as is for now, but will keep this in mind for future.

lib/charms/prometheus_k8s/v1/prometheus.py
balbirthomas: So what is the conclusion here ? Should I switch the bind address to ingress address ? . I have created a ticket for this here, https://github.com/canonical/prometheus-operator/issues/72 just in case this question does not have a definitive answer as yet.

lib/charms/prometheus_k8s/v1/prometheus.py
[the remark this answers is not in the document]
balbirthomas: From discussion with @johnsca it turns out the Kube State Metrics use case is that each unit of that charm exposes two endpoints (paths+ports). The units are homogenous. The next PR will support multiple static scrape config metadata blocks. The unit IPs will still be gathered directly from each unit. Each of these gathered IPs will be used to populate each of the multiple [static scrape config](https://prometheus.io/docs/prometheus/latest/configuration/configuration/#static_config) metadata blocks. This will support @johnsca use case and many more.

lib/charms/prometheus_k8s/v1/prometheus.py
balbirthomas: Tested again. Units do get removed from Prometheus scrape configuration when they depart. This was tested using the [prometheus-tester](git@github.com:balbirthomas/prometheus-tester.git) charm. Do let me know if there any other way you would like me to test or if there is anything else I can try.

If a unit departs and its *unit data bag* still persists or if and application is removed or its relation removed and the *application data bag* still persists, I would presume this would be considered a bug in Juju or the Operator Framework. Even though I am not seeing this, do let me know if I am mistaken in this presumption.

Leaving as is for now.

lib/charms/prometheus_k8s/v1/prometheus.py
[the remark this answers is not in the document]
balbirthomas: My previous post was not clear. Let me try and clarify. Prometheus [scrape configuration](https://prometheus.io/docs/prometheus/latest/configuration/configuration/#scrape_config) allows for a list of `<scrape_config>` blocks, each representing one "job". Each job has multiple metadata fields including `metrics_path`.  This implies each job can only have the same metrics path. Since @johnsca use case does conform to a single metrics path, the next PR will support only this, but it may be extended to multiple metrics paths using multiple jobs in the future. This forthcoming PR will allow units set one or more "metric ports" by a mechanism as yet not finalized.

lib/charms/prometheus_k8s/v1/prometheus.py
rbarry82: We should ask the Juju team what their plans for this are. My guess is that this is more or less deferred while they figure out cross-model relations and which address should be selected. I'd leave it as-is, if it were me.

lib/charms/prometheus_k8s/v1/prometheus.py
rbarry82: So the basic workflow @sed-i has is the same. The events are not actually asynchronous (except where they're coming from OF). The data bags still exist, at least for applications, until `RelationDeparted` or `RelationBroken` finishes. At least walking the Juju code (which I'm not that familiar with yet), and from @sed-i testing.

With a call stack more or less like:

```
ENTER relation_departed
  ENTER _on_scrape_relation_departed
  EMIT targets_changed
    ENTER charm.on_targets_changed
      ENTER charm._configure
        ENTER provder.jobs()
        # databag still exists, relation_departed has not finished
      EXIT
    EXIT
  EXIT
EXIT
```

Units do get removed from the scrape configuration. From your patch, I wonder if _applications_ do. If an arbitrary, say, Cassandra/MySQL enters a scrape relation, then the application leaves entirely, we would see `provder.jobs()` "successfully" iterate over the relations (including that one), "successfully" get the scrape targets from the databag, and set it in the config.

I did not test @sed-i 's charm myself, but now that it's a similar conversation two days in a row, I'll build from this branch and test for sanity.

balbirthomas: Applications also do get removed from the data bag. AFAIU the workflow is 

ENTER relation_departed
  ENTER _on_scrape_relation_departed
  EMIT targets_changed
  EXIT
EXIT
# Juju fires targets_changed

balbirthomas: Tested again. Both `remove-relation` and `remove-application` results in scrape target being removed from Prometheus dashboard. Let me know if there is any other test you would like me to try.

lib/charms/prometheus_k8s/v1/prometheus.py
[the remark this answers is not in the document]
balbirthomas: I have created a [ticket](https://github.com/canonical/prometheus-operator/issues/73) to pick up this feature in the next PR. Resolving this conversation here form. We can continue requirements gathering on the cited ticket.

lib/charms/prometheus_k8s/v1/prometheus.py
balbirthomas: I can confirm metadata was remaining in `prometheus.yml`. I have pushed a [commit](https://github.com/canonical/prometheus-operator/pull/71/commits/6ed46985c2ab2ac6eb382b08651413ddc527959a) fixing this. I did check this seems to have fixed the issue. Thanks.

lib/charms/prometheus_k8s/v1/prometheus.py
```diff
@@ -279,30 +231,79 @@ def jobs(self):
...
+            relation: an `ops.model.Relation` object whose static
+                scrape configuration is required.
 
-    def __init__(self, charm, name, consumes, multi=False):
-        """Construct a Prometheus charm client.
+        Returns:
+            A static scrape configuration for a specific relation.
+        """
+        if len(relation.units) == 0:
+            return {}
```
sed-i, line 253: just wondering if an empty dict is consistent with `return None` and `return scrape_config` below

lib/charms/prometheus_k8s/v1/prometheus.py
```diff
@@ -279,30 +231,79 @@ def jobs(self):
...
+
+        for relation in self._charm.model.relations[self._relation_name]:
+            job = self._static_scrape_config(relation)
+            if job:
+                scrape_jobs.append(job)
 
         return scrape_jobs
 
+    def _static_scrape_config(self, relation):
+        """Generate the static scrape configuration for a relation.
```
sed-i, line 243: consider documenting how the full scrape_config looks like ("pseudo schema"?)

lib/charms/prometheus_k8s/v1/prometheus.py
```diff
@@ -317,99 +318,72 @@ def __init__(self, charm, name, consumes, multi=False):
...
-        Args:
-            rel_id: an optional integer providing the relation ID for
-                the related Prometheus monitoring service
-                provider. This is only necessary if the
-                `PrometheusConsumer` has been instantiated in
-                `multi` mode.
+        Each time a consumer charm container is restarted it updates its own
+        host address in the unit relation data for the Prometheus provider.
         """
+        for relation in self._charm.model.relations[self._relation_name]:
```
sed-i, line 370: is it guaranteed in this context that `relations[self._relation_name]` never raises a KeyError?

src/charm.py
```diff
@@ -82,15 +80,12 @@ def _configure(self):
...
-        if self._stored.prometheus_config_hash != config_hash:
-            try:
-                container.push(PROMETHEUS_CONFIG, prometheus_config)
-                self._stored.prometheus_config_hash = config_hash
-                logger.info("Pushed new configuration")
-            except ConnectionError:
-                logger.info("Ignoring config changed since pebble is not ready")
-                return
+        try:
+            container.push(PROMETHEUS_CONFIG, prometheus_config)
```
sed-i, line 84: comment is outdated: no check anymore
but I don't understand:
1. why push in the constructor
2. how will you know to restart/hot-reload?

lib/charms/prometheus_k8s/v1/prometheus.py
rbarry82: In the sense that they both evaluate to `False`. It would be nice to see type hinting for the return type, but eh

lib/charms/prometheus_k8s/v1/prometheus.py
rbarry82: Yes. See [here](https://github.com/canonical/operator/blob/92c7922e451b14184ceb9e7170d1564306fa1046/ops/model.py#L414). `meta.relations` (which ultimately goes to `RelationMapping` as `relation_meta` and is iterated over to set the keys) comes from `metadata.yaml`.

src/charm.py
rbarry82: Hot reloading is moot point anyway. Prometheus requires a `HUP`, and Pebble does not (yet) provide a way to run arbitrary commands inside the sidecar

lib/charms/prometheus_k8s/v1/prometheus.py
```diff
@@ -279,30 +353,228 @@ def jobs(self):
...
+        scrape_jobs = json.loads(relation.data[relation.app].get("scrape_jobs"))
+
+        if not scrape_jobs:
+            return []
+
+        scrape_metadata = json.loads(relation.data[relation.app].get("scrape_metadata"))
+
+        job_name_prefix = "juju_{}_{}_{}_prometheus_{}_scrape".format(
+            scrape_metadata["model"],
+            scrape_metadata["model_uuid"][:7],
```
dstathis, line 389: Is model name not unique? If it isn't, we could maybe just check for dups rather than appending a string that is not human readable.

src/charm.py
dstathis: If you expand a bit, you can see this is not in the constructor

dstathis (approved):

lib/charms/prometheus_k8s/v1/prometheus.py
balbirthomas: Done.

lib/charms/prometheus_k8s/v1/prometheus.py
balbirthomas: The structure of the model name necessarily implies it will be unique for a single Juju controller. It is simpler to ensure that names are unique. This was deemed good enough as far as product requirements go.

Abuelodelanada (commented):
LGTM

PR #75, opened 2021-07-20
Unit tests

balbirthomas:
Added unit tests for recent enhancements to Prometheus charm library and fixes a couple of bugs with those enhancements.

### DIFF
--- a/run_tests
+++ b/run_tests
@@ -5,7 +5,7 @@
 USAGE="usage: $0
        -h    show help
        -f    fix linting errors
-       -u    run using Ubunut host native packages"
+       -u    run using Ubuntu host native packages"
 
 
 # Parse command line options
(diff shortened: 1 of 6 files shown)


run_tests
```diff
@@ -4,19 +4,24 @@
 
 USAGE="usage: $0
        -h    show help
-       -f    fix linting errors"
+       -f    fix linting errors
+       -u    run using Ubunut host native packages"
```
dstathis, line 8: Ubunut -> Ubuntu

dstathis (approved):

run_tests
balbirthomas: Ops. @dstathis may I request you to just "suggest an edit" because if I make another push to fix this, you may need to re-review and re-approve. I think if you "suggest an edit" and I accept this problem does not occur.

dstathis (dismissed):

sed-i (approved):

PR #77, opened 2021-07-21
Move to Grafana Juju topology, relabel `instance` to something matching Juju topology

rbarry82:
Relabel instance so it's all consistent when pods restart instead of using an IP address.

As part of the rebase, also pull in the new Grafana consumer library which uses Juju topology.


src/charm.py
```diff
@@ -353,7 +340,7 @@ def _prometheus_config(self):
 
         logger.debug("Prometheus config : {}".format(scrape_config))
 
-        return yaml.dump(scrape_config)
+        return yaml.dump(scrape_config, default_flow_style=True)
```
balbirthomas: I am surprised. I presume you really want to enforce "non block" style. I would have assumed this was the default.

balbirthomas (approved):

src/charm.py
rbarry82: It's really because:

```
yaml.dump({
  "key": "[__value__]"
  })
```

Will spit out:
```
key: '[__value__]'
```

Which Prometheus won't parse. To match exactly as:
```
source_labels: [__address__]
```
Or:
```
source_labels: [juju_model, juju_model_uuid, ...]
```

"non block" style was required, since pyyaml considers '[' dangerous in YAML. After a little more testing once I knew it worked, I removed the flow. Prometheus parses block-style YAML arrays just fine, despite it not being in any docs. The last patch changed this back.

PR #78, opened 2021-07-22
Don't put invalid characters in label names for 'juju_unit'

rbarry82:
'/' is not allowed. See [here](https://prometheus.io/docs/concepts/data_model/#metric-names-and-labels).

Also, don't add `juju_unit` to the relabel config on _every_ loop, only once we know we have a unit. This was noticed in scalability testing.


balbirthomas (approved):

balbirthomas (approved):

balbirthomas (commented):
There is a test failure.

lib/charms/prometheus_k8s/v0/prometheus.py
```diff
@@ -494,14 +494,14 @@ def _labeled_static_job_config(self, job, job_name_prefix, hosts, scrape_metadat
...
                 config["static_configs"].append(unitless_config)
-            else:
-                relabel_config["source_labels"].append("juju_unit")
 
             for host_name, host_address in hosts.items():
                 static_config = self._labeled_unit_config(
                     host_name, host_address, ports, labels, scrape_metadata
                 )
                 config["static_configs"].append(static_config)
+                if "juju_unit" not in relabel_config["source_labels"]:
```
balbirthomas, line 503: Not all scrape targets have unit labels. In particular "fully qualified hosts" (i.e. those whose address is not automatically gathered by the Prometheus charm library but set explicitly by consumer charms) do not have unit labels. Doesn't this line mean unit labels are attached to all units regardless. If so how does the unit get determined ?

balbirthomas: On further thought this should be fine since the `hosts` list is the automatically gathered list.

balbirthomas (approved):

PR #81, opened 2021-08-02
Workflow for ingesting Alerting

balbirthomas:
This PR provides a workflow for ingesting alert rules from consumer charms. This works as follows

1. Consumer charms must have a directory named `prometheus_alert_rules` at the top level in their source repository.
2. This directory can contain any number of files with extension `.rule`. Each such file can have one alert rule in YAML format for example
```
- alert: ScheduleableInstances
        expr: sum(hypervisor_schedulable_instances{nova_service_status="enabled",%%juju_topology%%}) < {{ min_schedulable_instances }}
        for: {{ wait_time }}
        labels:
          override_group_by: host
          severity: page
        annotations:
          description: |
            There is low capacity on the cloud, less than {{ min_schedulable_instances }} instances can be scheduled
          summary: There is low capacity on the cloud, less than {{ min_schedulable_instances }} instances can be scheduled
```
3. All occurrences of `%%juju_topology%%` in the alert rule `expr` are replaced by appropriate Juju topology labels of the charm submitting the alert rule.


lib/charms/prometheus_k8s/v0/prometheus.py
```diff
@@ -375,6 +379,27 @@ def jobs(self):
...
+                continue
+
+            alert_rules = json.loads(relation.data[relation.app].get("alert_rules", "{}"))
+            scrape_metadata = json.loads(relation.data[relation.app].get("scrape_metadata", "{}"))
+            if not scrape_metadata:
+                continue
+
+            if alert_rules:
+                alerts[relation.id] = {
+                    "groups": alert_rules["groups"],
```
mmanciop, line 395: I assume this is the `name` on an item in the `groups` list. Shouldn't it be `group` then?

balbirthomas: [Alert rules](https://prometheus.io/docs/prometheus/latest/configuration/alerting_rules/) YAML uses `groups` so I stuck with that.

mmanciop: Yeah but it is unclear imo. Like this, you imply an alert can have multiple groups and that ain’t the case, right?

mmanciop:
@balbirthomas I think we need more docs and example about why using `%%juju_topology%%` for _each_ metric is so important.

src/charm.py
```diff
@@ -315,6 +337,7 @@ def _prometheus_config(self):
 
         scrape_config = {
             "global": self._prometheus_global_config(),
+            "rule_files": ["{}/juju_*.rules".format(RULES_DIR)],
```
mmanciop, line 340: Is it `.rules` or `.rule`? I hope the latter :-)

lib/charms/prometheus_k8s/v0/prometheus.py
balbirthomas: I am only trying to imitate the exact same structure of YAML files that Prometheus itself uses, so the code can be easily extended to any future use cases. A Prometheus alerting rules file consists of a list of groups. Each group consists of a list of alerts that are executed sequentially. For this charm library each deployment of the consumer charm produces a single alert rules file on the provider side. This rules file has only one group, but it still needs a `groups` key under whose value is a list of one group. This structure is dictated by Prometheus. At present we only have one group because the convention adopted was to have one rule per file. Hence the consumer charm does not specify any group or its name. It only provides a list of alerts. So all these alerts are grouped into a single group. In the future if it is decided that we want to allow the consumer to provide multiple groups then the current implementation can be easily extended by allow that files with extension `.group` can contain more than one alert, all of which belong to the same group.

src/charm.py
balbirthomas: The file gathers all the rules, provided by a single consumer into a list of alerts, this list of alerts is a "list of rules" belonging to a group, in a list of groups. The structure of this file has to conform exactly what Prometheus expects it to be - a list of `groups`.

balbirthomas:
> @balbirthomas I think we need more docs and example about why using `%%juju_topology%%` for _each_ metric is so important.

Agreed. I am about to start on unit tests and then documentation. Hence this is still a draft PR.

CONTRIBUTING.md
```diff
@@ -115,15 +189,15 @@ and hence must be used with caution.
...
 
 - Configure scrape targets through Juju relations.
-- Enable alerting through a relation with Alertmanager.
+- Configure alerting rules through relations with scrape target charms.
+- Enable alert forwarding through a relation with Alertmanager.
 - Support metrics visualisation through Grafana.
 
 ## Road Map
 
 Roughly by order of priority
```
mmanciop, line 198: You know, I'd remove the entire section. The roadmap is something we will eventually discuss on Discourse, has no place IMO in a Git history.

balbirthomas: OK. Will do.

lib/charms/prometheus_k8s/v0/prometheus.py
```diff
@@ -241,6 +240,51 @@ def _on_scrape_targets_changed(self, event):
...
+  expr: job:request_latency_seconds:mean5m{my_key=my_value, %%juju_topology%%} > 0.5
+  for: 10m
+  labels:
+    severity: Medium
+    type: HighLatency
+  annotations:
+    summary: High request latency for {{ $labels.instance }}.
+```
+
+It is **very important** to note the `%%juju_topology%%` filter in the
```
mmanciop, line 267: In a couple of weeks we should be able to link to a blog post about it :D

balbirthomas: I can put it in then if you like.

lib/charms/prometheus_k8s/v0/prometheus.py
```diff
@@ -684,6 +794,77 @@ def _set_unit_ip(self, event):
...
+        Args:
+            rule: a dictionary representing a Prometheus alert rule.
+
+        Returns:
+            a dictionary representing Prometheus alert rule with Juju
+            topology labels.
+        """
+        metadata = self._scrape_metadata
+        labels = rule.get("labels", {})
+        labels["juju_model"] = metadata["model"]
```
mmanciop, line 809: We should not override these labels if they exist. We should have the equivalent of the `honor_labels = true` semantics.

mmanciop: I foresee some very unlikely corner cases involving remapping models for Bootstack where this may matter. Overall, I think it is good practice for us to be mindful of generally not plainly overriding metadata.

mmanciop (changes requested):
I'd say almost there :-)

lib/charms/prometheus_k8s/v0/prometheus.py
balbirthomas: If we do this there are no guarantees on the sanity of these labels. I do not see why a consumer charm would like to set Juju topology labels explicitly. From what you have suggested uptil now I understood that you wanted these labels to be added automatically.

CONTRIBUTING.md
balbirthomas: Done.

lib/charms/prometheus_k8s/v0/prometheus.py
balbirthomas: Could we consider changing this behaviour when we actually hit those corner cases and can not find any way out ? For the moment could we advertise the behaviour as "Please do not set Juju Topology labels since these will be set for you automatically".

CONTRIBUTING.md
```diff
@@ -98,8 +98,82 @@ object provides an interface to a running Prometheus instance. This
...
+and `alerts()` methods of the `PrometheusProvider`.
+
+The `jobs()` method gathers a list of scrape jobs from all related
+consumer charms. In doing so it invokes the `_static_scrape_config()`
+method for each such relation. `_static_scrape_config()` generates a
+unique job name for each job, sanitizes the job and associates Juju
+topology labels with it. Labeling of Juju topology is done by
+`_labeled_static_job_config()`. However the labeling differs for
+scrape targets whose host address was automatically gathered and those
+were addresses were explicitly specified. Hence three separate
```
sed-i, line 119: > were addresses were

typo?

CONTRIBUTING.md
```diff
@@ -98,8 +98,82 @@ object provides an interface to a running Prometheus instance. This
...
+functions are required for labeling `_set_juju_labels()` for labels
+common to both but `_labeled_unitless_config()` and
+`_labeled_unit_config()` for the differing labels. The list of
```
sed-i, line 122: Consider simplifying

CONTRIBUTING.md
```diff
@@ -98,8 +98,82 @@ object provides an interface to a running Prometheus instance. This
...
+automatically gathered host address is provided by
+`_relation_hosts()`.
+
+The `PrometheusConsumer` is responsible for forwarding scrape
+configuration, scrape target addresses, scrape metadata and alert
+rules to the Prometheus provider. In doing so it also ensures that the
+alert rules have Juju topology labels and filters injected into
+them. The `_set_unit_ip()` methods forwards scrape target host
+addresses using unit relation data. The `_set_scrape_metadata()`
+method forwards all the other information using application relation
```
sed-i, line 132: > all the other information

consider clarifying

CONTRIBUTING.md
```diff
@@ -98,8 +98,82 @@ object provides an interface to a running Prometheus instance. This
...
+  `_is_valid_timespec()` and `_are_valid_labels()`.
+- `_alerting_config()` generates configuration related to
+  Alertmanager(s), using the Alertmanager charm library.
+- `jobs()` and `alerts()` methods of the `PrometheusProvider`
+  object.
+
 ## Design Choices
 
+This charm manages a [Prometheus](https://prometheus.io) workload and
+in doing so it obtains configuration data through its relation with
```
sed-i, line 169: relation -> relations

CONTRIBUTING.md
```diff
@@ -98,8 +98,82 @@ object provides an interface to a running Prometheus instance. This
...
+other charms. The key design choice here was to ensure that the
+structure of this data exchanged closely mirrors the format of
+Prometheus' own configuration. This ensures that these relational data
+structures are at least as extensible as Prometheus' own
+configuration. Besides these data structure would already be familiar
```
sed-i, line 174: Besides -> Besides,

CONTRIBUTING.md
```diff
@@ -98,8 +98,82 @@ object provides an interface to a running Prometheus instance. This
...
+to Prometheus domain experts.
+
 This Prometheus charm does not support peer relations because as yet
```
sed-i, line 177: as **of** yet?

lib/charms/prometheus_k8s/v0/prometheus.py
```diff
@@ -241,6 +240,51 @@ def _on_scrape_targets_changed(self, event):
...
+  labels:
+    severity: Medium
+    type: HighLatency
+  annotations:
+    summary: High request latency for {{ $labels.instance }}.
+```
+
+It is **very important** to note the `%%juju_topology%%` filter in the
+expression for the alert rule shown above. This filter is a stub that
+is automatically replaced by the consumer charm's Juju topology. Such
```
sed-i, line 269: > the consumer charm's Juju topology

consider adding an example, e.g. "model-app-unitnum", because "juju topology" is vague

lib/charms/prometheus_k8s/v0/prometheus.py
```diff
@@ -241,6 +240,51 @@ def _on_scrape_targets_changed(self, event):
...
+a topology filter is essential to ensure that alert rules submitted by
+one consumer charm generates alerts only for that same charm.  The
+Prometheus provider charm may be related to multiple consumer
+charms. Without this filter rules submitted by one consumer charm will
```
sed-i, line 273: Without this -> Without this,

lib/charms/prometheus_k8s/v0/prometheus.py
```diff
@@ -684,6 +794,77 @@ def _set_unit_ip(self, event):
...
+
+        All rules from files for a consumer charm are loaded into a single
+        group. The generated name of this group includes Juju topology
+        prefixes.
+
+        Returns:
+            a list of Prometheus alert rule groups.
+        """
+        alerts = []
+        for path in Path(ALERT_RULES_PATH).glob("*.rule"):
```
sed-i, line 848: If organizing rules is subfolders is a potential use case then double start glob

```suggestion
        for path in Path(ALERT_RULES_PATH).glob("**/*.rule"):
```

lib/charms/prometheus_k8s/v0/prometheus.py
```diff
@@ -270,6 +316,8 @@ def _on_scrape_targets_changed(self, event):
 # to 0 if you are raising the major API version
 LIBPATCH = 2
 
+ALERT_RULES_PATH = "prometheus_alert_rules"
```
sed-i, line 319: I would prefix `./` for clarity

lib/charms/prometheus_k8s/v0/prometheus.py
```diff
@@ -375,6 +423,62 @@ def jobs(self):
...
+            )
+            if not scrape_metadata:
+                continue
+
+            if alert_rules:
+                alerts[relation.id] = {
+                    "groups": alert_rules["groups"],
+                    "model": scrape_metadata["model"],
+                    "model_uuid": scrape_metadata["model_uuid"][:7],
+                    "application": scrape_metadata["application"],
```
sed-i, line 477: 1. Potential KeyError. Does it make more sense to have an exception raised or to log an error and abort without an exception?
2. Consider using [TypedDict](https://www.python.org/dev/peps/pep-0589/) for documenting/anchoring the implicit protocol.

lib/charms/prometheus_k8s/v0/prometheus.py
```diff
@@ -673,6 +777,12 @@ def _set_scrape_metadata(self, event):
             self._scrape_jobs
         )
 
+        alert_groups = self._labeled_alert_groups
+        if alert_groups:
+            event.relation.data[self._charm.app]["alert_rules"] = json.dumps(
```
sed-i, line 782: leader guard?

lib/charms/prometheus_k8s/v0/prometheus.py
```diff
@@ -684,6 +794,77 @@ def _set_unit_ip(self, event):
...
+            a dictionary representing a Prometheus alert rule that filters based
+            on juju topology.
+        """
+        metadata = self._scrape_metadata
+        topology = (
+            'juju_model="{}", juju_model_uuid="{}", juju_application="{}"'.format(
+                metadata["model"], metadata["model_uuid"], metadata["application"]
+            )
+        )
+        expr = rule["expr"]  # a rule has to have an "expr"
```
sed-i, line 831: Potential KeyError

lib/charms/prometheus_k8s/v0/prometheus.py
```diff
@@ -684,6 +794,77 @@ def _set_unit_ip(self, event):
...
+        expr = expr.replace("%%juju_topology%%", topology)
```
sed-i, line 832: Is there a generic way to add a topology filter to a rule, to obviate "%%juju_topology%%"?

lib/charms/prometheus_k8s/v0/prometheus.py
```diff
@@ -684,6 +794,77 @@ def _set_unit_ip(self, event):
...
+        groups = []
+        if alerts:
+            group = {
+                "name": "{}_{}_{}_alerts".format(
+                    metadata["model"], metadata["model_uuid"], metadata["application"]
+                ),
+                "rules": alerts,
+            }
+            groups.append(group)
+        return groups
```
sed-i, line 866: Based on the above this is a list of 0 or 1 items. Missing loop?

src/charm.py
```diff
@@ -108,6 +112,30 @@ def _configure(self):
...
+        """Create alert rule files for all Prometheus consumers.
+
+        Args:
+            container: the Prometheus workload container into which
+                alert rule files need to be created.
+        """
+        if not self.provider_ready:
+            return
+
+        container.remove_path(RULES_DIR, recursive=True)
```
sed-i, line 125: shouldn't the rules folder be on a persistent storage and the auto generated rules (a subfolder) within that?

sed-i: iirc remove_path can also raise ConnectionError

lib/charms/prometheus_k8s/v0/prometheus.py
```diff
@@ -684,6 +794,77 @@ def _set_unit_ip(self, event):
...
+                    rule = rules[0]  # each file is list of one rule
+                    rule = self._label_alert_topology(rule)
+                    rule = self._label_alert_expression(rule)
+                    alerts.append(rule)
+        metadata = self._scrape_metadata
+        groups = []
+        if alerts:
+            group = {
+                "name": "{}_{}_{}_alerts".format(
+                    metadata["model"], metadata["model_uuid"], metadata["application"]
```
sed-i, line 861: This set of string literals is repeated several times

src/charm.py
```diff
@@ -108,6 +112,30 @@ def _configure(self):
...
+
+        for rel_id, alerts in self.prometheus_provider.alerts().items():
+            filename = "juju_{}_{}_{}_rel_{}_alert.rules".format(
+                alerts["model"], alerts["model_uuid"], alerts["application"], rel_id
+            )
+            try:
+                path = "{}/{}".format(RULES_DIR, filename)
```
sed-i, line 132: os.path.join

src/charm.py
sed-i: os.path.join

tests/test_consumer.py
```diff
@@ -141,3 +144,55 @@ def test_consumer_sanitizes_jobs(self, _):
...
+                self.assertIn("juju_model", labels)
+                self.assertIn("juju_model_uuid", labels)
+                self.assertIn("juju_application", labels)
+
+
+def expression_labels(expr):
+    """Extract labels from an alert rule expression.
+
+    Args:
+        expr: a string representing and alert expression.
```
sed-i, line 187:
```suggestion
        expr: a string representing an alert expression.
```

sed-i (dismissed):

lib/charms/prometheus_k8s/v0/prometheus.py
mmanciop: > I do not see why a consumer charm would like to set Juju topology labels explicitly.

How about THIS charm: https://warthogs.atlassian.net/browse/OP-234?atlOrigin=eyJpIjoiODIzN2VmNTg3YWVjNDc3ODgzODEzN2M2NjIyNjkxN2EiLCJwIjoiaiJ9 :D

mmanciop: I am also thinking that [Scrape Configuration Charm](https://warthogs.atlassian.net/browse/OP-223?atlOrigin=eyJpIjoiOTU5ODIyZTAyZTI3NDU4Mjg0NGJmNDQ4YjNiZjY3YmEiLCJwIjoiaiJ9) will also need to override the topology labels so that the values of the actually-scraped charm will be used instead.

lib/charms/prometheus_k8s/v0/prometheus.py
```diff
@@ -375,6 +423,62 @@ def jobs(self):
...
+        rules are all placed into a single group, even though Prometheus itself
+        allows for multiple groups within a single alert rules file.
+
+        Returns:
+            a dictionary of alert rule groups and associated scrape
+            metadata indexed by relation ID.
+        """
+        alerts = {}
+        for relation in self._charm.model.relations[self._relation_name]:
+            if len(relation.units) == 0:
```
jnsgruk, line 460: This might flow better as 

```python
if len(relation.units):
    # do stuff
```

lib/charms/prometheus_k8s/v0/prometheus.py
```diff
@@ -375,6 +423,62 @@ def jobs(self):
...
+                continue
+
+            alert_rules = json.loads(
+                relation.data[relation.app].get("alert_rules", "{}")
+            )
+            scrape_metadata = json.loads(
+                relation.data[relation.app].get("scrape_metadata", "{}")
+            )
+            if not scrape_metadata:
```
jnsgruk, line 469: again, is there any reason to not use:

```python
if scrape_metadata:
    # do stuff
```

lib/charms/prometheus_k8s/v0/prometheus.py
```diff
@@ -375,6 +423,62 @@ def jobs(self):
...
+                continue
+
+            if alert_rules:
+                alerts[relation.id] = {
+                    "groups": alert_rules["groups"],
+                    "model": scrape_metadata["model"],
+                    "model_uuid": scrape_metadata["model_uuid"][:7],
```
jnsgruk, line 476: `scrape_metadata["model_uuid"][:7]` is a bit opaque? A comment here would not go amiss

lib/charms/prometheus_k8s/v0/prometheus.py
jnsgruk: Agreed, either catching the exception or using `dict.get()` to provide some value in the case it's missing is probably a better call. Though I'm not sure what the default would be.

lib/charms/prometheus_k8s/v0/prometheus.py
```diff
@@ -673,6 +777,12 @@ def _set_scrape_metadata(self, event):
             self._scrape_jobs
         )
 
+        alert_groups = self._labeled_alert_groups
+        if alert_groups:
```
jnsgruk, line 781:
```suggestion
        if self._labeled_alert_groups:
```

lib/charms/prometheus_k8s/v0/prometheus.py
```diff
@@ -673,6 +777,12 @@ def _set_scrape_metadata(self, event):
             self._scrape_jobs
         )
 
+        alert_groups = self._labeled_alert_groups
```
jnsgruk, line 780:
```suggestion
```

lib/charms/prometheus_k8s/v0/prometheus.py
```diff
@@ -673,6 +777,12 @@ def _set_scrape_metadata(self, event):
...
+        if alert_groups:
+            event.relation.data[self._charm.app]["alert_rules"] = json.dumps(
+                {"groups": alert_groups}
```
jnsgruk, line 783:
```suggestion
                {"groups": self._labeled_alert_groups}
```

lib/charms/prometheus_k8s/v0/prometheus.py
jnsgruk: You could also do a conditional assignment as a simple guard against this?

```python
labels["juju_model"] = metadata["model"] if not labels["juju_model"] else labels["juju_model"]
```

Probably makes the code a little more coarse to read, but does get around the issue.

That said, does Balbir have a point here, do we in face want people to be able to override these fields at all?

lib/charms/prometheus_k8s/v0/prometheus.py
```diff
@@ -684,6 +794,77 @@ def _set_unit_ip(self, event):
...
+            rule: a dictionary representing a Prometheus alert rule.
+
+        Returns:
+            a dictionary representing a Prometheus alert rule that filters based
+            on juju topology.
+        """
+        metadata = self._scrape_metadata
+        topology = (
+            'juju_model="{}", juju_model_uuid="{}", juju_application="{}"'.format(
+                metadata["model"], metadata["model_uuid"], metadata["application"]
```
jnsgruk, line 828: You could consider using an f-string here

lib/charms/prometheus_k8s/v0/prometheus.py
```diff
@@ -684,6 +794,77 @@ def _set_unit_ip(self, event):
...
+            )
+        )
+        expr = rule["expr"]  # a rule has to have an "expr"
+        expr = expr.replace("%%juju_topology%%", topology)
+        rule["expr"] = expr
+        return rule
```
jnsgruk, line 834:
```suggestion
        if expr := rule.get("expr", None):  # a rule has to have an "expr"
            expr = expr.replace("%%juju_topology%%", topology)
            rule["expr"] = expr
            return rule
        # raise an exception?
```

lib/charms/prometheus_k8s/v0/prometheus.py
balbirthomas: Both Scrape Target and Scrape Configuration could accept `scrape_metadata` from the scrape targets they are related too. In fact `scrape_metadata` which contains the Juju topology is part of the `prometheus_scrape` interface. I have created a [https://github.com/canonical/prometheus-operator/issues/83] to think this through and if necessary refactor the Juju topology labeling workflow in a future pull request.

lib/charms/prometheus_k8s/v0/prometheus.py
```diff
@@ -684,6 +794,77 @@ def _set_unit_ip(self, event):
...
+        """
+        alerts = []
+        for path in Path(ALERT_RULES_PATH).glob("*.rule"):
+            if path.is_file():
+                with path.open() as rule_file:
+                    rules = yaml.safe_load(rule_file)
+                    rule = rules[0]  # each file is list of one rule
+                    rule = self._label_alert_topology(rule)
+                    rule = self._label_alert_expression(rule)
+                    alerts.append(rule)
```
jnsgruk, line 855:
```suggestion
                with path.open() as rule_file:
                    rules = yaml.safe_load(rule_file)
                    if len(rules):
                        rule = rules[0]  # each file is list of one rule
                        rule = self._label_alert_topology(rule)
                        rule = self._label_alert_expression(rule)
                        alerts.append(rule)
```

CONTRIBUTING.md
balbirthomas: Yes thanks.

lib/charms/prometheus_k8s/v0/prometheus.py
mmanciop: Ah, right. Then I retire my objection :-)

lib/charms/prometheus_k8s/v0/prometheus.py
jnsgruk: Even with this change, this code becomes a little hard to read because there are so many calls out to various functions. I don't think the complexity of `_label_alert_topology` and `label_alert_expression` are sufficient enough to necessarily warrant a separate function. It might just be simpler to have these elements in this main function...

src/charm.py
jnsgruk: It can, but you could use

```python
with container.is_ready():
    # do stuff
````

to handle that case...

File: src/charm.py [hunk 34]
```diff
@@ -108,6 +112,30 @@ def _configure(self):
...
+        if not self.provider_ready:
+            return
+
+        container.remove_path(RULES_DIR, recursive=True)
+
+        for rel_id, alerts in self.prometheus_provider.alerts().items():
+            filename = "juju_{}_{}_{}_rel_{}_alert.rules".format(
+                alerts["model"], alerts["model_uuid"], alerts["application"], rel_id
+            )
+            try:
```
jnsgruk, line 131: Replace this try/except and use `with container.is_ready():`?

src/charm.py
jnsgruk: +1, use `pathlib`

jnsgruk (changes requested):

CONTRIBUTING.md
balbirthomas: You mean simplifying the wording ?

CONTRIBUTING.md
balbirthomas: The first sentence in the paragraph lists all the information forwarded. Let me know if this does not make it clear.

CONTRIBUTING.md
balbirthomas: Thanks.

CONTRIBUTING.md
balbirthomas: Thanks.

CONTRIBUTING.md
balbirthomas: "as yet" is common usage too.

lib/charms/prometheus_k8s/v0/prometheus.py
```diff
@@ -241,6 +240,51 @@ def _on_scrape_targets_changed(self, event):
...
         ...
 
+## Alerting Rules
+
+This charm library also supports gathering alerting rules from all
+related consumer charms and enabling corresponding alerts within the
+provider charm.  Alert rules are automatically gathered by Consumer
+charms when using this library, from a directory conventionally named
+`prometheus_alert_rules`. This directory must reside at the top level
+in the Consumer charm. Each file in this directory is assumed to be a
```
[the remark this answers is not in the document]
balbirthomas: Yes.

lib/charms/prometheus_k8s/v0/prometheus.py
balbirthomas: Thanks.

lib/charms/prometheus_k8s/v0/prometheus.py
balbirthomas: Good suggestion but I am inclined to hold of on this until specifically requested by a user.

lib/charms/prometheus_k8s/v0/prometheus.py
balbirthomas: Only reason I did not do that is so that this constant can be used as part of others for example in writing tests. Besides `./` can also give the misleading impression that it is relative to the `lib/charms/prometheus_k8s/v0` directory.

lib/charms/prometheus_k8s/v0/prometheus.py
```diff
@@ -375,6 +423,62 @@ def jobs(self):
...
+        """Fetch alerts for all relations.
+
+        A Prometheus alert rules file consists of a list of "groups". Each
+        group consists of a list of alerts (`rules`) that are sequentially
+        executed. This method returns all the alert rules provided by each
+        related consumer charm. These rules may be used to generate a
+        separate alert rules file for each relation since the returned list
+        of alert groups are indexed by relation ID. Also for each relation ID
+        associated scrape metadata such as Juju model, UUID and application
+        name are provided so the a unique name may be generated for the rules
```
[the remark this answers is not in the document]
balbirthomas: Thanks. Changed locally along with changes for other suggestions of yours. Will push shortly.

lib/charms/prometheus_k8s/v0/prometheus.py
balbirthomas: There should never be a KeyError (if that happens, someone broke the library, did not test it, or abused the charm relation). Either scrape metadata exists or doesn't, like wise for alert rules and both these are explicitly checked for. Any protocol there is, is between Prometheus consumer and provider, both of which are implemented in this library. My own view on any validation of relation data schemas should be supported through constructs within the operator framework rather than each charm having to invent its own.

lib/charms/prometheus_k8s/v0/prometheus.py
balbirthomas: It exists.

lib/charms/prometheus_k8s/v0/prometheus.py
balbirthomas: An alert rule without `expr` is not an alert rule.

lib/charms/prometheus_k8s/v0/prometheus.py
balbirthomas: Not sure what you mean by "generic way". A PromQL parser would be pretty generic.

lib/charms/prometheus_k8s/v0/prometheus.py
balbirthomas: It is indeed a list of zero or one group, hence no need for a loop. These seemingly odd constructs ensure that the data passed from the consumer to provider has a structure that mimics Prometheus' own configuration. This is explained in `CONTRIBUTING.md`.

src/charm.py
balbirthomas: Yes it can. I am not a huge fan of littering my code with lot of Try Catch blocks and neither of using a try catch block for doing more than one thing. Alerts are pushed to workload containers after the Prometheus configuration itself is. If it happens that Prometheus configuration push succeeded and immediately after that alert failed it would be very very odd. But I will give this another thought.

lib/charms/prometheus_k8s/v0/prometheus.py
balbirthomas: Clarified.

lib/charms/prometheus_k8s/v0/prometheus.py
```diff
@@ -241,6 +240,51 @@ def _on_scrape_targets_changed(self, event):
...
+
+It is **very important** to note the `%%juju_topology%%` filter in the
+expression for the alert rule shown above. This filter is a stub that
+is automatically replaced by the consumer charm's Juju topology. Such
+a topology filter is essential to ensure that alert rules submitted by
+one consumer charm generates alerts only for that same charm.  The
+Prometheus provider charm may be related to multiple consumer
+charms. Without this filter rules submitted by one consumer charm will
+also result in corresponding alerts for other consumer charms. Hence
+every alert rule expression must include such a topology filter stub.
```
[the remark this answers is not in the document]
balbirthomas: Yes, by parsing PromQL expressions. Asking the user to do it for now  is the simpler option.

lib/charms/prometheus_k8s/v0/prometheus.py
balbirthomas: As opposed to defining ALL CAPS global constants and repeating them several times, with the benefit of being able to change them in one place.

src/charm.py
balbirthomas: Done.

tests/test_consumer.py
balbirthomas: Thanks. done. will push shortly.

lib/charms/prometheus_k8s/v0/prometheus.py
balbirthomas: I can certainly change it to `if not len(relation.units)` . Please see my comment below on my preference for negative conditionals.

lib/charms/prometheus_k8s/v0/prometheus.py
balbirthomas: Indeed there is. The positive condition forces greater depth in nesting. However a negative conditional just aborts further processing.

balbirthomas: I prefer to keep code as flat as possible and only nest when there is an intrinsic need for processing nested data structures.

lib/charms/prometheus_k8s/v0/prometheus.py
balbirthomas: Ok. It was @mmanciop suggestion to keep the names short, by just using first seven characters of uuid and this seemed reasonable to me. I will comment it.

lib/charms/prometheus_k8s/v0/prometheus.py
jnsgruk: What I meant was adjust the flow of the function to drop the `continue`, so rather than explicitly bailing when the length is zero, just do work when it isn't

balbirthomas: I understand. What I am saying is by bailing out when the length is zero I do not add another level of nesting. Just my preference. Perhaps taking the Zen `Flat is better than nested` a bit too far maybe, but I think not.

lib/charms/prometheus_k8s/v0/prometheus.py
```diff
@@ -375,6 +425,62 @@ def jobs(self):
...
+            if not scrape_metadata:
+                continue
+
+            if alert_rules:
+                alerts[relation.id] = {
+                    "groups": alert_rules["groups"],
+                    "model": scrape_metadata["model"],
+                    "model_uuid": scrape_metadata["model_uuid"][:7],
+                    "application": scrape_metadata["application"],
+                }
```
jnsgruk, line 480:
```suggestion
            if relation.units:
                # Load alert rules and scrape data from relation data bag
                alert_rules = json.loads(relation.data[relation.app].get("alert_rules", "{}"))
                scrape_metadata = json.loads(relation.data[relation.app].get("scrape_metadata", "{}"))
                
                if scrape_metadata and alert_rules:
                    alerts[relation.id] = {
                        "groups": alert_rules["groups"],
                        "model": scrape_metadata["model"],
                        "model_uuid": scrape_metadata["model_uuid"][:7],
                        "application": scrape_metadata["application"],
                    }
```

This is quite a lot shorter, and imo easier to read. One thing to check is if the top conditional is valid (i.e. `if relation.units` has the desired effect)

lib/charms/prometheus_k8s/v0/prometheus.py
balbirthomas: I can switch this to `if alert_groups := self._labeled_alert_groups:` to avoid having to invoke `self._labeled_alert_groups` in the body of the if.

lib/charms/prometheus_k8s/v0/prometheus.py
jnsgruk: Also subject to line breaks/length limits ;)

lib/charms/prometheus_k8s/v0/prometheus.py
balbirthomas: Breaking out code into small functions with intention revealing name is a stratagem to make it self documenting. The function name states the intent of the function body.

lib/charms/prometheus_k8s/v0/prometheus.py
jnsgruk: Was really just a comment that in this case `self._labeled_alert_groups` is not too hard to understand, or long to write - and thus I'm not sure what value the extra variable gives?

src/charm.py
balbirthomas: Nice. Thanks. Will do. Container `is_ready` seems new and was too busy to keep myself updated. It will be good if there was a way to get automated notifications when [sdk docs](https://juju.is/docs/sdk) or [ops docs](https://ops.rtfd.io) change.

lib/charms/prometheus_k8s/v0/prometheus.py
jnsgruk: Updated...

src/charm.py
balbirthomas: `pathlib` done. Thanks.

src/charm.py
balbirthomas: Thanks. switched to using `is_ready`.

lib/charms/prometheus_k8s/v0/prometheus.py
balbirthomas: I am not convinced that `rule.get("expr", None)` is a good idea. Suppose a user creates an alert rule that does not have an alert expression (clearly and invalid alert rule), what should be the behaviour ? As it stands a `KeyError` exception will be raised. However with the proposed change the invalid rule will be silently ignored. I do not think being silent would be a good idea. I think it may be reasonable to throw and explicit custom exception here or set the charm into a blocked state. What would be better ?

jnsgruk: Yeh, so my comment at the end was an indication that there probably needs to be an else, or an exception raised to indicate that something invalid was specified

jnsgruk: The `if expr := rule.get("expr", None):` will evaluate false if a rule isnt set and give you the opportunity to handle that better

balbirthomas: OK. Will add a custom exception `InvalidAlertRule` implement, test, and document it then push. I copy and paste this suggested code change into my changes instead of creating commit within github (hope that is ok).

jnsgruk: of course :)

lib/charms/prometheus_k8s/v0/prometheus.py
balbirthomas: I think a custom exception will be required here too to prevent silent failure to safe load alert rule. I need to think if it can be the same as `InvalidAletRule` (my current inclination) or it needs a different one. Again will implement, test, document and push rather than commit code from within github.

lib/charms/prometheus_k8s/v0/prometheus.py
balbirthomas: Hmm :) . Still does not feel right. What if `alert_rules` exist but not `scrape_metadata` or vice versa. Seems like in this case an incomplete alert will be returned in the list of alerts. I think your previous edit was on the right track in using a conditional that was the conjunction of the two. 

But let me give it another thought. I will see if walrus operator and conjunction in the conditional shortens it. But I still prefer using negative conditionals to abort processing any alert if any one of its requirements are not met. I am inclined to convert even the `if alert_rules` into a negative conditional. 

With regard to semantics of `relation.units` and its behaviour in boolean contexts, it would save time if these things were explicitly documented thereby avoiding the need to dig into Operator Framework code base to figure it out.

jnsgruk: (And BTW, the behaviour you describe above is the same in the initial implementation...)

CONTRIBUTING.md
sed-i: I mean increase cognitive ease. It is difficult for me to read and I don't really understand it.

lib/charms/prometheus_k8s/v0/prometheus.py
sed-i: Not caps necessarily. I would be inclined towards `@dataclass` if there was better interconversion with dicts.

lib/charms/prometheus_k8s/v0/prometheus.py
sed-i: hopefully information is distributed equally across all bits or concentrated on the lower bits

lib/charms/prometheus_k8s/v0/prometheus.py
balbirthomas: Thank you for your suggestion. I would have used things like data classes if I wanted to encapsulate and compose sets of behaviour free information into larger objects. I do think my present use cases warrants this. Perhaps in the future if my code complexity grows I will consider it.

lib/charms/prometheus_k8s/v0/prometheus.py
balbirthomas: Added `InvalidAlertRule` exception here on failure to load rule. Will push shortly.

lib/charms/prometheus_k8s/v0/prometheus.py
balbirthomas: Code has be refactored to use a conjunction as the conditional. Please check last two commits. I still retain negative conditional on `relation.units` to avoid an unnecessary extra level of nesting. But most importantly invalid rules errors are now logged and will not raise exceptions.

I will add unit tests and then request re-review.

lib/charms/prometheus_k8s/v0/prometheus.py
balbirthomas: This has been refactored and invalid expression now logs and error instead of raising a KeyError. Unit tests follow.

lib/charms/prometheus_k8s/v0/prometheus.py
balbirthomas: Refactored so to use walrus and avoid using `self._labeled_alert_groups` twice.

lib/charms/prometheus_k8s/v0/prometheus.py
balbirthomas: refactored.

lib/charms/prometheus_k8s/v0/prometheus.py
balbirthomas: refactored.

lib/charms/prometheus_k8s/v0/prometheus.py
balbirthomas: switched to `relation.units` but retained negative conditional to avoid an additional level of nesting.

sed-i (approved):

lib/charms/prometheus_k8s/v0/prometheus.py
balbirthomas: I have enclosed this in a Try/Catch block, log the error and ignore the alerts for that relation. I do not see any sensible defaults either. As mentioned above this problem can only occur if the library is broken or the relation is abused.

lib/charms/prometheus_k8s/v0/prometheus.py
```diff
@@ -375,6 +425,64 @@ def jobs(self):
...
+            if alert_rules and scrape_metadata:
+                try:
+                    alerts[relation.id] = {
+                        "groups": alert_rules["groups"],
+                        "model": scrape_metadata["model"],
+                        "model_uuid": scrape_metadata["model_uuid"][:7],
+                        "application": scrape_metadata["application"],
+                    }
+                except KeyError:
+                    logger.error("Relation %s has invalid data", relation.id)
```
jnsgruk, line 482: Hmm, what happens when if this exception fires? Do you have a half formed alert in the alerts list, or does it fail to assign completely?


below is ugly but trying to convey the point

```suggestion
           valid = ("groups" in alert_rules.keys() and 
                          all(i in ["model","model_uuid","application"] for i in scrape_metadata.keys()))
            
            if valid:
                alerts[relation.id] = {
                    "groups": alert_rules["groups"],
                    "model": scrape_metadata["model"],
                    "model_uuid": scrape_metadata["model_uuid"][:7],
                    "application": scrape_metadata["application"],
                }
```

sed-i, line 482:
```suggestion
                except KeyError as e:
                    logger.error("Relation %s has invalid data: '%s' key is missing", relation.id, e)
```

lib/charms/prometheus_k8s/v0/prometheus.py
```diff
@@ -684,6 +805,88 @@ def _set_unit_ip(self, event):
...
+            with path.open() as rule_file:
+                # Load a list of rules from file then add labels and filters
+                try:
+                    rules = yaml.safe_load(rule_file)
+                    rule = rules[0]  # each file is list of one rule
+                    rule = self._label_alert_topology(rule)
+                    rule = self._label_alert_expression(rule)
+                    alerts.append(rule)
+                except Exception:
+                    logger.error("Failed to read alert rules from %s", path.name)
```
sed-i, line 876: This is typically a linting error on PyCharm: `except Exception` may be used only if followed by `raise`; otherwise should be specific on the exception types.
For example you can specify yaml.scanner.ScannerError, IndexError, etc.

src/charm.py
```diff
@@ -108,6 +113,31 @@ def _configure(self):
...
+            container: the Prometheus workload container into which
+                alert rule files need to be created.
+        """
+        if not self.provider_ready:
+            return
+
+        with container.is_ready():
+            container.remove_path(RULES_DIR, recursive=True)
+
+        for rel_id, alerts in self.prometheus_provider.alerts().items():
```
sed-i, line 129: shouldn't the if block be under the context manager as well, instead of the remaining try-except?

tests/test_consumer.py
```diff
@@ -2,16 +2,18 @@
...
+import re
 import unittest
 from unittest.mock import patch
 
 from ops.charm import CharmBase
 from ops.framework import StoredState
 
-# from ops.model import Network
 from ops.testing import Harness
+import charms.prometheus_k8s.v0.prometheus as prometheus_lib
```
sed-i, line 13: This is one reason why I prefer "static" class variables, such as `PrometheusConsumer.ALERT_RULES_PATH`. Just saying :D

sed-i (dismissed):

src/charm.py
balbirthomas: which "if block" ?

lib/charms/prometheus_k8s/v0/prometheus.py
balbirthomas: In general it would be a good idea to trap specific exceptions. In this case it is largely pointless since all errors are handled in the same way.

CONTRIBUTING.md
balbirthomas: This is only a very high level bird eye view of the codebase. It can not be a substitute for the code. It is intendend to facilitate reading the code by giving a lay of the land - what is there and where is it. 

If you have a more specific issue I can address it.

src/charm.py
balbirthomas: I am trapping `ConnectionError` on a per alert rule basis.

lib/charms/prometheus_k8s/v0/prometheus.py
balbirthomas: Also this is why I think there is no substitute for human over sight such as code review. Automated tools are just that and hence can be pedantic.

tests/test_consumer.py
balbirthomas: This was a good suggestion. Adopted now. Thanks.

sed-i (approved):

jnsgruk (approved):

PR #84, opened 2021-08-06
Add ingress relation

xavpaice:
Uses the nginx-ingress-integrator charm v0 ingress library to add an
ingress relation to the charm.

Issue: https://github.com/canonical/prometheus-operator/issues/25

### DIFF
--- a/charmcraft.yaml
+++ b/charmcraft.yaml
@@ -6,3 +6,7 @@ bases:
     run-on:
     - name: "ubuntu"
       channel: "20.04"
+parts:
+  charm:
+    build-packages:
+    - git
--- a/src/charm.py
+++ b/src/charm.py
@@ -14,6 +14,7 @@ from ops.model import ActiveStatus, MaintenanceStatus
 from ops.pebble import ConnectionError
 from prometheus_server import Prometheus
 from charms.grafana_k8s.v1.grafana_source import GrafanaSourceConsumer
+from charms.nginx_ingress_integrator.v0.ingress import IngressRequires
 from charms.prometheus_k8s.v0.prometheus import PrometheusProvider
 from charms.alertmanager_k8s.v0.alertmanager import AlertmanagerConsumer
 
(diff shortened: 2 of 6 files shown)


balbirthomas (approved):

src/charm.py
```diff
@@ -57,6 +58,24 @@ def __init__(self, *args):
...
+        )
+
+    @property
+    def _external_hostname(self):
+        """Return the external hostname to be passed to ingress via the relation."""
+        # It is recommended to default to `self.app.name` so that the external
+        # hostname will correspond to the deployed application name in the
+        # model, but allow it to be set to something specific via config.
+
+        return self.config["web-external-url"] or f"{self.app.name}.juju"
```
mmanciop, line 78: Yeah but shouldn't this be consumed in the [`_prometheus_config`](https://github.com/canonical/prometheus-operator/blob/c76f8699f792abd7d8d01023b40ae73f31acf9b8/src/charm.py#L308) method?

balbirthomas: It should set the Prometheus command line argument for `--web.external-url=<...>` and trigger a restart of Prometheus. The command line arguments are set in `_cli_args()` in `charm.py`.

mmanciop: I got it the `web-external-url` work at https://github.com/mmanciop/prometheus-operator/tree/ingress

src/prometheus_server.py
```diff
@@ -30,8 +30,11 @@ def build_info(self):
         except urllib3.exceptions.MaxRetryError:
             return {}
 
-        info = json.loads(response.data.decode("utf-8"))
-        if info["status"] == "success":
-            return info["data"]
-        else:
+        try:
```
balbirthomas, line 33: This commit has been merged into master, with some modifications. You may want to rebase.

src/charm.py
```diff
@@ -180,6 +221,8 @@ def _cli_args(self):
 
         return args
 
+    # TODO This should throw and be handled. RIght now, we just
```
balbirthomas, line 224: How would throwing an exception be handled ? Only reasonable options I can think of is either set the charm in blocked state or log an error and let the charm continue running since at least it will still be scraping metrics from all other relations.

xavpaice:
TODO:

* Tests fail, test_provider_preserves_builtin_juju_topology_labels missing juju_unit.  Need to fix the test, or fix the problem (the charm appears to work fine)
* _is_valid_timespec has some added comments in this PR but the actual change is unrelated to this PR

xavpaice:
I've split this out to have the commits for the ingress relation in this MR only.

charmcraft.yaml
```diff
@@ -6,3 +6,7 @@ bases:
     run-on:
     - name: "ubuntu"
       channel: "20.04"
+parts:
```
balbirthomas, line 9: This producing the error
```
Bad charmcraft.yaml content:
- extra field 'charm' not permitted in 'parts' configuration (full execution logs in '/home/balbir/snap/charmcraft/common/charmcraft-log-1lw3qw7d')
```
I am using latest stable charmcraft version 1.1.1.

mmanciop:
Yeah you will need to update to 1.2.0, it is pretty unavoidable :-)

charmcraft.yaml
mmanciop: Yeah you will need to update to 1.2.0, it is pretty unavoidable :-)

balbirthomas (approved):

PR #86, opened 2021-08-06
Defensively parse JSON in build_info

mmanciop:
There are situations in which Prometheus is not ready yet that will not return valid JSON (and may return actually an empty response body).

This commit adds error handling for invalid JSON in the response returned by the `api/v1/status/buildinfo` endpoint.

### DIFF
--- a/src/prometheus_server.py
+++ b/src/prometheus_server.py
@@ -30,8 +30,11 @@ class Prometheus:
         except urllib3.exceptions.MaxRetryError:
             return {}
 
-        info = json.loads(response.data.decode("utf-8"))
-        if info["status"] == "success":
-            return info["data"]
-        else:
+        try:
+            info = json.loads(response.data.decode("utf-8"))
+            if info["status"] == "success":
+                return info["data"]
+            else:
+                return {}
+        except:
             return {}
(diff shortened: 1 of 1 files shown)


src/prometheus_server.py
```diff
@@ -30,8 +30,11 @@ def build_info(self):
...
-        info = json.loads(response.data.decode("utf-8"))
-        if info["status"] == "success":
-            return info["data"]
-        else:
+        try:
+            info = json.loads(response.data.decode("utf-8"))
+            if info["status"] == "success":
+                return info["data"]
+            else:
+                return {}
```
balbirthomas, line 38: There are now three `return {}` in this method. This code is growing organically :) . I wonder if some refactor can be effected here to make it nicer. For example just log a debug message in the `except` block and then a default `return {}` at the end of the function.

Seems like you are trying to trap *all* exceptions from `json.loads`. I am surprised this is raising an exception because the data in this case is generated by Prometheus (upstream code). Also `info["status"]` should be provided by `urllib`.

balbirthomas (approved):

balbirthomas (approved):

PR #89, opened 2021-08-09
Enable forwarding of scrape jobs through proxy charms

mmanciop:
The `prometheus_scrape` relation works remarkably differently from the `http` relation used in the Prometheus machine charm. We considered several options to achieve interoperability without having to modify existing machine charms. One that was promising was to implement the machine-charm relations in the Prometheus Operator charm, but we found out that we cannot correctly create the Juju topology, because the remote [model name and uuid are not available over cross-model relations](https://chat.charmhub.io/charmhub/pl/wois69md6tfnprf3cagta7bjch).

So, we went for The [LMAProxyOperator charm](https://github.com/mmanciop/lma-proxy-operator): a "proxy" machine charm that is meant to be deployed in the same model as the machine charms to be scraped by Prometheus, and that will translate between the machine-charm Prometheus relations and `prometheus_scrape`.

Since this use-case is somewhat of a corner-case for the Prometheus charm library, it is not worth it to provide a first-level support like allowing the charm using `PrometheusConsumer` to specify custom `scrape_metadata` or the like.

However, there are a few corner cases to adjust:
1. Prometheus must not override the Juju topology labels in a `static_config` if they are present. One of the things that the LMAProxyOperator charm is supposed to do, is to _fill in the Juju topology_, since it is the only charm that (1) understands how `prometheus_scrape` works and (2) has access to the model name and uuid (since it is deployed into that model)
2. The Prometheus operator should not generate a `target` for the charm that is _forwarding_ scrape jobs; that is, `LMAProxyOperator` should have no target of its own, although it is related with Prometheus over a `prometheus_scrape`. THis case can be detected by comparing the Juju topology labels in the `static_config` against the `scrape_metadata` and, if differences in values are detected, do not add additional targets.

### DIFF
--- a/lib/charms/prometheus_k8s/v0/prometheus.py
+++ b/lib/charms/prometheus_k8s/v0/prometheus.py
@@ -689,12 +689,14 @@ class PrometheusProvider(ProviderBase):
             topology information with the exception of unit name.
         """
         juju_labels = labels.copy()  # deep copy not needed
-        with_topology = not JUJU_TOPOLOGY_LABEL_SET.isdisjoint(labels.keys())
 
-        if not with_topology:
-            juju_labels["juju_model"] = scrape_metadata["model"]
-            juju_labels["juju_model_uuid"] = scrape_metadata["model_uuid"]
-            juju_labels["juju_application"] = scrape_metadata["application"]
+        juju_labels.update(
+            {
+                "juju_model": scrape_metadata["model"],
+                "juju_model_uuid": scrape_metadata["model_uuid"],
+                "juju_application": scrape_metadata["application"],
+            }
+        )
 
         return juju_labels
 
@@ -759,7 +761,10 @@ class PrometheusProvider(ProviderBase):
             A dictionary containing the static scrape configuration
             for a single wildcard host.
         """
-        juju_labels = self._set_juju_labels(labels, scrape_metadata)
+        with_topology = not JUJU_TOPOLOGY_LABEL_SET.isdisjoint(labels.keys())
+
+        if not with_topology:
+            juju_labels = self._set_juju_labels(labels, scrape_metadata)
 
         if "juju_unit" not in juju_labels:
             # '/' is not allowed in Prometheus label names. It technically works,
--- a/lib/charms/prometheus_k8s/v0/prometheus.py
+++ b/lib/charms/prometheus_k8s/v0/prometheus.py
@@ -616,9 +616,12 @@ class PrometheusProvider(ProviderBase):
                         )
                     )
 
+        return self._to_job_config(job_name, job["metrics_path"], static_configs)
+
+    def _to_job_config(self, job_name, metrics_path, static_configs):
         return {
             "job_name": job_name,
-            "metrics_path": job["metrics_path"],
+            "metrics_path": metrics_path,
             "relabel_configs": [
                 # This relabelling rule will match metrics that do not have the full Juju topology,
                 # and ensure that metrics without a juju_unit are still unique per instance
@@ -648,8 +651,6 @@ class PrometheusProvider(ProviderBase):
             ],
             "static_configs": static_configs,
         }
-
-
 
     def _is_aggregated_static_config(self, static_config, scrape_metadata):
         # Charms that forward scrape jobs on behalf of other charms
--- a/lib/charms/prometheus_k8s/v0/prometheus.py
+++ b/lib/charms/prometheus_k8s/v0/prometheus.py
@@ -302,11 +302,12 @@ relation data provide eponymous information.
 """
 
 import json
-import yaml
 import logging
 from pathlib import Path
-from ops.framework import EventSource, EventBase, ObjectEvents
-from ops.relation import ProviderBase, ConsumerBase
+
+import yaml
+from ops.framework import EventBase, EventSource, ObjectEvents
+from ops.relation import ConsumerBase, ProviderBase
 
 # The unique Charmhub library identifier, never change it
 LIBID = "bc84295fef5f4049878f07b131968ee2"
@@ -419,6 +420,7 @@ class PrometheusProvider(ProviderBase):
 
         for relation in self._charm.model.relations[self._relation_name]:
             static_scrape_jobs = self._static_scrape_config(relation)
+
             if static_scrape_jobs:
                 scrape_jobs.extend(static_scrape_jobs)
 
@@ -457,6 +459,7 @@ class PrometheusProvider(ProviderBase):
             metadata indexed by relation ID.
         """
         alerts = {}
+
         for relation in self._charm.model.relations[self._relation_name]:
             if not relation.units:
                 continue
@@ -518,6 +521,7 @@ class PrometheusProvider(ProviderBase):
         hosts = self._relation_hosts(relation)
 
         labeled_job_configs = []
+
         for job in scrape_jobs:
             config = self._labeled_static_job_config(
                 _sanitize_scrape_configuration(job),
@@ -543,9 +547,11 @@ class PrometheusProvider(ProviderBase):
             the specified relation.
         """
         hosts = {}
+
         for unit in relation.units:
             if host_address := relation.data[unit].get("prometheus_scrape_host"):
                 hosts[unit.name] = host_address
+
         return hosts
 
     def _labeled_static_job_config(self, job, job_name_prefix, hosts, scrape_metadata):
@@ -586,13 +592,16 @@ class PrometheusProvider(ProviderBase):
         }
 
         for static_config in static_configs:
+
             labels = static_config.get("labels", {}) if static_configs else {}
             all_targets = static_config.get("targets", [])
 
             ports = []
             unitless_targets = []
+
             for target in all_targets:
                 host, port = target.split(":")
+
                 if host.strip() == "*":
                     ports.append(port.strip())
                 else:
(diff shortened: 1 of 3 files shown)


lib/charms/prometheus_k8s/v0/prometheus.py
```diff
@@ -573,48 +579,91 @@ def _labeled_static_job_config(self, job, job_name_prefix, hosts, scrape_metadat
...
-                config["static_configs"].append(static_config)
-                if "juju_unit" not in relabel_config["source_labels"]:
-                    relabel_config["source_labels"].append("juju_unit")
+            if ports := [
+                target.split(":")[1] for target in targets if target.startswith("*:")
+            ]:
+                if self._is_aggregated_static_config(static_config, scrape_metadata):
+                    # Do not add targets for proxy applications that publish scrape
+                    # jobs with Juju labels that do not match the scrape metadata
+                    logger.error(
```
balbirthomas, line 604: If I am not mistaken this would happen only if `PrometheusConsumer` is being used in `lma-proxy` charm, since `PrometheusConsumer` always sets a default scrape job with target `*:80`. If so, I am concerned that we may be using `PrometheusConsumer` for two distinct responsibilities

1. Manage metric scrape relation between a consumer charm and the Prometheus provider.
2. Aggregate scrape jobs from multiple charms and forward all of them together to the Prometheus provider.

It should be possible to create a third object (say `PrometheusAggregator`) in a separate library module (say `lib.charms.v0.prometheus.proxy.py`) that provides much of the functionality required by 2, in a reusable format.

lib/charms/prometheus_k8s/v0/prometheus.py
```diff
@@ -573,48 +579,91 @@ def _labeled_static_job_config(self, job, job_name_prefix, hosts, scrape_metadat
...
-                if host.strip() == "*":
-                    ports.append(port.strip())
-                else:
-                    unitless_targets.append(target)
-
-            if unitless_targets:
-                unitless_config = self._labeled_unitless_config(
-                    unitless_targets, labels, scrape_metadata
+            if unitless_targets := [
+                target for target in targets if not target.startswith("*:")
```
balbirthomas, line 590: So consumer charm writer must be warned through documentation that there should be no space between `*` and `:`.

lib/charms/prometheus_k8s/v0/prometheus.py
```diff
@@ -573,48 +579,91 @@ def _labeled_static_job_config(self, job, job_name_prefix, hosts, scrape_metadat
...
 
-        config["relabel_configs"] = [relabel_config]
+                for host_name, host_address in hosts.items():
+                    static_configs.append(
+                        self._labeled_unit_config(
+                            host_name, host_address, ports, labels, scrape_metadata
+                        )
+                    )
 
-        return config
```
balbirthomas, line 617: This is a rather big return statement, not to mention it makes the function itself rather long. Besides most of the dictionary is boiler plate with only two substitutions - `job["metrics_path"]` and `static_configs` that is easy to miss in the "boiler plate forest". What is being substituted can be made explicit in the context of a separate function for example

```
return self._job_configs(job["metrics_path"], static_configs)
```

tests/test_provider.py
```diff
@@ -208,14 +230,14 @@ def validate_jobs(self, jobs):
                 self.assertIn("juju_model_uuid", labels)
                 self.assertIn("juju_application", labels)
 
-            relabel_configs = job["relabel_configs"]
-            self.assertEqual(len(relabel_configs), 1)
+            # relabel_configs = job["relabel_configs"]
```
balbirthomas, line 233: Should we not check for presence of `relabel_configs` ?

tests/test_provider.py
```diff
@@ -345,10 +367,42 @@ def test_provider_overwrites_juju_topology_labels(self):
...
+        original_labels = juju_job_labels(JOBS_WITH_BUILT_IN_JUJU_TOPOLOGY[0])
         labels = juju_job_labels(jobs[0])
         for label_name, label_value in labels.items():
-            self.assertNotEqual(label_value, bad_labels[label_name])
+            self.assertEqual(label_value, original_labels[label_name])
+
+    def test_provider_skips_jobs_with_juju_topology_and_ports_and_mismatch_scrape_metadata(
+        self,
+    ):
+        """This test checks that the provider skips static_configs that
```
balbirthomas, line 378: +1: Nice comment.

lib/charms/prometheus_k8s/v0/prometheus.py
```diff
@@ -542,11 +548,11 @@ def _relation_hosts(self, relation):
             A dictionary that maps unit names to unit addresses for
             the specified relation.
         """
-        hosts = {}
-        for unit in relation.units:
-            if host_address := relation.data[unit].get("prometheus_scrape_host"):
-                hosts[unit.name] = host_address
-        return hosts
+        return {
```
balbirthomas, line 551: I do not think this change is warranted. It is simple enough to read `return hosts` and understand that `hosts` are being returned. I do not think `return {... multi-line dictionary comprehension ... }` improves readability and any (if any) miniscule performance gains are useful here.

lib/charms/prometheus_k8s/v0/prometheus.py
```diff
@@ -573,48 +579,91 @@ def _labeled_static_job_config(self, job, job_name_prefix, hosts, scrape_metadat
...
-                config["static_configs"].append(unitless_config)
 
-            for host_name, host_address in hosts.items():
-                static_config = self._labeled_unit_config(
-                    host_name, host_address, ports, labels, scrape_metadata
-                )
-                config["static_configs"].append(static_config)
-                if "juju_unit" not in relabel_config["source_labels"]:
-                    relabel_config["source_labels"].append("juju_unit")
+            if ports := [
```
balbirthomas, line 598: I do not think this change is warranted either. Nesting a conditional list comprehension **in** a walrus assignment expression **in** a if condition (that is immediately followed by another if condition which could have been a conjunction),  does not improve readability and any miniscule performance gains are not required here.

lib/charms/prometheus_k8s/v0/prometheus.py
```diff
@@ -517,17 +525,15 @@ def _static_scrape_config(self, relation):
...
+        return [
+            self._labeled_static_job_config(
                 _sanitize_scrape_configuration(job),
                 job_name_prefix,
                 hosts,
                 scrape_metadata,
             )
-            labeled_job_configs.append(config)
-
-        return labeled_job_configs
```
balbirthomas, line 530: Similar comments elsewhere in this review. I do not think there is any reason to replace a  `return labeled_job_configs` which makes clear what is being returned by naming it. The suggested replacement of a nested list comprehension with nested function calls that nest function calls is not a readability improvement or a required performance gain, if any.

lib/charms/prometheus_k8s/v0/prometheus.py
```diff
@@ -573,48 +579,91 @@ def _labeled_static_job_config(self, job, job_name_prefix, hosts, scrape_metadat
...
+                    ],
+                    "separator": "_",
+                    "target_label": "instance",
+                    "regex": "(.*)",
+                },
+            ],
+            "static_configs": static_configs,
+        }
+
+    def _is_aggregated_static_config(self, static_config, scrape_metadata):
```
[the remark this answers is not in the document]
balbirthomas: On second thoughts here is an alternative that may be worth considering : `scrape_metadata` is passed only by `PrometheusConsumer` since it is used for creating job names and labels. However if a different object such as `PrometheusAggregator` may be free to take an entirely different approach, leaving `scrape_metadata` only for use by individual charms. For example `PrometheusAggregator` can submit the entire job spec and be wholly responsible for setting job names, labels etc. Then the absence of `scrape_metadata` or if required an alternative key say `aggregated_metadata` can be indicative of what `PrometheusProvider` needs to deal with. With this approach the changes to `PrometheusProvider` are minimal, with little cross talk between the two cases (such as checking for intersecting labels), and no changes to `PrometheusConsumer`. 

Finally there remains the case of aggregating alert rules. I have not had the opportunity to think this through but my instinct is to by keeping `PrometheusAggregator` as a separate object from `PrometheusConsumer` this use case will also be easy to solve.

tests/test_provider.py
mmanciop: Yeah I guess so, in case we refactor

lib/charms/prometheus_k8s/v0/prometheus.py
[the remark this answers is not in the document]
mmanciop: > But what happens if a errant charm accidentally does set a Juju topology label or intentionally wants to ?

Accidental: Bug
Intentional: Feature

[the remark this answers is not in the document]
mmanciop: > could have an incomplete set of topology labels.

This is worth testing for!

lib/charms/prometheus_k8s/v0/prometheus.py
mmanciop: I did not propose this for reasons of performance. I found myself debugging stuff in this codebase because of list modification via `append`, and I feel safer with a more declarative approach. If you vero this change, I'll revert.

lib/charms/prometheus_k8s/v0/prometheus.py
mmanciop: We are indeed using `PrometheusConsumer` for two distinct use-cases right now. This PR is more a draft to understand what the `PrometheusAggregator` should look like, than a real PR atm.

mmanciop: For clarity I turned this PR into a Draft PR

balbirthomas:
I am closing this pull request because a new object `MetricsEndpointAggregator` is now available in an Prometheus updated charm library on the [aggregator branch](https://github.com/balbirthomas/prometheus-operator/tree/aggregator). This object subsumes the functionality of forwarding scrape jobs from a proxy charm and also supports forwarding of alert rules. An updated version of the [lma-proxy charm](https://github.com/balbirthomas/lma-proxy-operator/tree/aggregator) which uses the updated Prometheus charm library is also available. The unit tests of the lma-proxy charm have also been updated and it has been functionally tested manually using the steps described in its `README.md` and found to forward both scrape jobs and alerts. Both these git branches are now being cleaned up, unit tested, documented and prepared for code review.

PR #92, opened 2021-08-11
Implement reload of configurations rather than Prometheus restart

### LINKED ISSUE #36: Explore converting pod restart to pod reload where possible
Instead of restarting the pod investigate and replace pod restarts with Prometheus config reload where possible.

mmanciop:
In case the configuration of Prometheus changes, but not its startup command, there is no reason to restart the process: we already activate the Web API to trigger the hot reload of configurations from file, and this commit adds support for it in `PrometheusServer`, and its triggering when we detect and configuration reload that would create the same layer as the existing one in terms of startup commands.

Closes #36

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,4 +1,3 @@
 git+git://github.com/canonical/operator
 pyaml
 requests
-deepdiff
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,3 +1,4 @@
 git+git://github.com/canonical/operator
 pyaml
-urllib3
+requests
+deepdiff
--- a/src/charm.py
+++ b/src/charm.py
@@ -108,7 +108,7 @@ class PrometheusCharm(CharmBase):
         # If the startup arguments are the same and we use the
         # we lifecycle, sent the config reload HTTP request instead
 
-        services_changed = not DeepDiff(
+        services_changed = DeepDiff(
             current_layer.get("services", {}),
             new_layer.get("services", {}),
         )
--- a/src/charm.py
+++ b/src/charm.py
@@ -7,6 +7,8 @@ import logging
 import yaml
 import json
 
+from deepdiff import DeepDiff
+
 from ops.charm import CharmBase
 from ops.framework import StoredState
 from ops.main import main
@@ -32,6 +34,8 @@ class PrometheusCharm(CharmBase):
         logger.debug("Initializing Charm")
 
         super().__init__(*args)
+
+        self._prometheus = Prometheus("localhost", str(self.model.config["port"]))
 
         self._stored.set_default(provider_ready=False)
 
@@ -97,10 +101,20 @@ class PrometheusCharm(CharmBase):
         self._set_alerts(container)
 
         # setup the workload (Prometheus) container and its services
-        layer = self._prometheus_layer()
         plan = container.get_plan()
-        if plan.services != layer["services"]:
-            container.add_layer("prometheus", layer, combine=True)
+        current_layer = plan.services if plan else {}
+        new_layer = self._prometheus_layer()
+
+        # If the startup arguments are the same and we use the
+        # we lifecycle, sent the config reload HTTP request instead
+        if (
+            current_layer
+            and "services" in current_layer
+            and not DeepDiff(current_layer["services"], new_layer["services"])
+        ):
+            self._prometheus.trigger_configuration_reload()
+        else:
+            container.add_layer("prometheus", new_layer, combine=True)
 
             if container.get_service("prometheus").is_running():
                 container.stop("prometheus")
@@ -405,8 +419,7 @@ class PrometheusCharm(CharmBase):
             a string consisting of the Prometheus version information or
             None if Prometheus server is not reachable.
         """
-        prometheus = Prometheus("localhost", str(self.model.config["port"]))
-        info = prometheus.build_info()
+        info = self._prometheus.build_info()
         if info:
             return info.get("version", None)
         return None
--- a/src/charm.py
+++ b/src/charm.py
@@ -33,7 +33,9 @@ class PrometheusCharm(CharmBase):
 
         super().__init__(*args)
 
-        self._prometheus_server = Prometheus("localhost", str(self.model.config["port"]))
+        self._prometheus_server = Prometheus(
+            "localhost", str(self.model.config["port"])
+        )
 
         self._stored.set_default(provider_ready=False)
 
--- a/src/charm.py
+++ b/src/charm.py
@@ -35,7 +35,7 @@ class PrometheusCharm(CharmBase):
 
         super().__init__(*args)
 
-        self._prometheus = Prometheus("localhost", str(self.model.config["port"]))
+        self._prometheus_api = Prometheus("localhost", str(self.model.config["port"]))
 
         self._stored.set_default(provider_ready=False)
 
@@ -115,7 +115,7 @@ class PrometheusCharm(CharmBase):
         )
 
         if not prometheus_service_changed:
-            self._prometheus.reload_configuration()
+            self._prometheus_api.reload_configuration()
             logger.info("Configuration reloaded")
         else:
             container.add_layer("prometheus", new_layer, combine=True)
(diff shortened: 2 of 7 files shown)


mmanciop:
Converting to draft because I wanna run one more manual test before merge

balbirthomas:
This should be linked to https://github.com/canonical/prometheus-operator/issues/36

mmanciop:
I spoke too soon, a test went red ;-)

src/charm.py
```diff
@@ -107,20 +107,28 @@ def _configure(self):
 
         # If the startup arguments are the same and we use the
         # we lifecycle, sent the config reload HTTP request instead
-        if (
-            current_layer
-            and "services" in current_layer
-            and not DeepDiff(current_layer["services"], new_layer["services"])
-        ):
+
+        services_changed = not DeepDiff(
```
balbirthomas, line 111: Doesn't Pebble itself check if there is a change to service layers ? If not shouldn't such a feature at least have an associated issue ticket in Operator Framework of Pebble ?

mmanciop: In principle what you say sounds sensible, but what we would be doing it sending to a different container, the one running Pebble, the new service plan over RPC, and use the result of a diff check over there to decide whether to trigger a restart from the charm container. Sounds a bit much to me.

mmanciop: @balbirthomas `add_layer` is stateless: https://github.com/canonical/operator/blob/7344b5c284ff4009dbde29c4df99205eaad2e8ba/ops/pebble.py#L870

We could add an API to Pebble to check for diffs, but I am ambivalent on whether it is scope creep for `ops.pebble` or not. @jnsgruk WDYT?

src/prometheus_server.py
```diff
@@ -1,17 +1,36 @@
...
             host: host address of Prometheus application.
             port: port on which Prometheus service is exposed.
         """
-        self.host = host
-        self.port = port
-        self.http = urllib3.PoolManager()
+        self.base_url = "http://{}:{}".format(host, port)
+        self.api_timeout = api_timeout
+
+    def trigger_configuration_reload(self):
```
balbirthomas, line 18: Minor: rather wordy method name. If context (name of object this method is used on) is considered a shorter method name may suffice. For example `self._prometheus.reload_configuration()` .

src/charm.py
```diff
@@ -97,16 +101,34 @@ def _configure(self):
...
         plan = container.get_plan()
-        if plan.services != layer["services"]:
-            container.add_layer("prometheus", layer, combine=True)
+        current_layer = plan.services if plan else {}
+        new_layer = self._prometheus_layer()
+
+        # If the startup arguments are the same and we use the
+        # we lifecycle, sent the config reload HTTP request instead
+
+        services_changed = DeepDiff(
```
balbirthomas, line 111: I still think this should be a utility method in the Pebble framework (if it isn't already) instead of every charm having to reinvent the wheel here. If there isn't such a feature in Pebble (which is not fully documented as yet) then as a stop gap measure this is fine but perhaps with a reference to an open ticket on github.

tests/test_server.py
```diff
@@ -1,26 +1,50 @@
...
 from prometheus_server import Prometheus
 
 
 class TestServer(unittest.TestCase):
     def setUp(self):
         self.prometheus = Prometheus("localhost", "9090")
 
-    @patch("urllib3.PoolManager.request")
-    def test_prometheus_server_returns_valid_data(self, request):
+    @responses.activate
```
balbirthomas, line 14: +1: nice

balbirthomas (approved):
Approving but please note two things I am concerned about

1. Creating a bespoke mechanism to detect change in `plan.services`
2. How does reloading interact with alerting.

src/charm.py
```diff
@@ -97,16 +101,34 @@ def _configure(self):
         self._set_alerts(container)
 
         # setup the workload (Prometheus) container and its services
-        layer = self._prometheus_layer()
         plan = container.get_plan()
-        if plan.services != layer["services"]:
-            container.add_layer("prometheus", layer, combine=True)
+        current_layer = plan.services if plan else {}
```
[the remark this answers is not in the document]
mmanciop: Yeah it needed refactoring

src/charm.py
```diff
@@ -97,16 +101,34 @@ def _configure(self):
...
+
+        # If the startup arguments are the same and we use the
+        # we lifecycle, sent the config reload HTTP request instead
+
+        services_changed = DeepDiff(
+            current_layer.get("services", {}),
+            new_layer.get("services", {}),
+        )
+
+        if current_layer and not services_changed:
```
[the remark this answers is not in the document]
mmanciop: Yes, reload is sufficient when new rules are added. And the alert rule changes are coming through a relation change, that triggers the reconfiguration

src/charm.py
mmanciop: There is no such feature in Pebble that I can see. I am also not sure it is warranted. This is the first charm I see this logic needed, and usually I go by the Rule of Three before asking for a new API.

requirements.txt
```diff
@@ -1,3 +1,4 @@
 git+git://github.com/canonical/operator
 pyaml
-urllib3
+requests
+deepdiff
```
sed-i, line 4: Could you compare between `json.dumps({...}, sort_keys=True)` instead?
I think that avoiding a 3rd party dep is worth it considering the current use-case.

sed-i (commented):
Nice. How about using it as part of `_configure()`?

src/charm.py
```diff
@@ -33,6 +35,8 @@ def __init__(self, *args):
 
         super().__init__(*args)
 
+        self._prometheus = Prometheus("localhost", str(self.model.config["port"]))
```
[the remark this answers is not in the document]
mmanciop: `_prometheus_api` it is

src/charm.py
```diff
@@ -97,16 +101,35 @@ def _configure(self):
...
+            new_layer.get("services", {}).get("prometheus", {}),
+        )
+
+        if not prometheus_service_changed:
+            self._prometheus.reload_configuration()
+            logger.info("Configuration reloaded")
+        else:
+            container.add_layer("prometheus", new_layer, combine=True)
+
+            restart_needed = container.get_service("prometheus").is_running()
```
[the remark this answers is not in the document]
mmanciop: @sed-i what is your advise? Handle the exceptions? In my eyes, if Prometheus must update the config and cannot, the hook has to crash. It does not strike me as the kind of transient error we could really handle.

src/prometheus_server.py
```diff
@@ -1,17 +1,36 @@
...
             host: host address of Prometheus application.
             port: port on which Prometheus service is exposed.
         """
-        self.host = host
-        self.port = port
-        self.http = urllib3.PoolManager()
+        self.base_url = "http://{}:{}".format(host, port)
+        self.api_timeout = api_timeout
+
+    def reload_configuration(self):
```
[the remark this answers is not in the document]
mmanciop: It is, check around like 118 in charm.py

src/prometheus_server.py
```diff
@@ -1,17 +1,36 @@
...
+        """Send a POST request to to hot-reload the config.
+        This reduces down-time compared to restarting the service.
+        Returns:
+          True if reload succeeded (returned 200 OK); False otherwise.
+        """
+        url = "{}/-/reload".format(self.base_url)
```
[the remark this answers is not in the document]
mmanciop: I felt the value of the extra dependency was not worth it

requirements.txt
mmanciop: I am kinda wary of nested datastructures. If we compares just the prometheus server, that could work though

src/charm.py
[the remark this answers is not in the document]
sed-i: ah yes, I didn't notice the add_layer.

src/prometheus_server.py
```diff
@@ -1,19 +1,38 @@
...
+
+            if response.status_code == 200:
+                return True
+        except (ConnectionError, ConnectTimeout) as e:
+            logger.debug("config reload error via %s: %s", url, str(e))
+
+        return False
 
-    def build_info(self):
+    def build_info(self) -> dict:
```
sed-i, line 35:
```suggestion
    def build_info(self) -> Dict[str, str]:
```

sed-i (approved):

src/prometheus_server.py
mmanciop: I am fairly sure we need an additional import though

PR #93, opened 2021-08-12
Add optional override for rules path

xavpaice:
Prometheus alert rules are found in a named directory, this change adds
the option to override that path in the PrometheusConsumer class should
a charm need it.

### DIFF
--- a/lib/charms/prometheus_k8s/v0/prometheus.py
+++ b/lib/charms/prometheus_k8s/v0/prometheus.py
@@ -302,11 +302,12 @@ relation data provide eponymous information.
 """
 
 import json
-import yaml
 import logging
 from pathlib import Path
-from ops.framework import EventSource, EventBase, ObjectEvents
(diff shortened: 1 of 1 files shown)


lib/charms/prometheus_k8s/v0/prometheus.py
```diff
@@ -757,10 +765,14 @@ def __init__(self, charm, name, consumes, service_event, jobs=[], multi=False):
...
                 this object must support interaction with multiple
                 Prometheus monitoring service providers.
+            alert_rules_path: an optional path for the location of alert rules
+                files.  Defaults to "prometheus_alert_rules" at the top level of
+                the charm.
         """
         super().__init__(charm, name, consumes, multi)
 
         self._charm = charm
+        self._ALERT_RULES_PATH = alert_rules_path
```
balbirthomas, line 775: Not sure if all caps was absolutely necessary here :-).

balbirthomas (approved):

PR #94, opened 2021-08-13
Correct external_url path

xavpaice:
The command line switch for web.external-url needs to be set without the application name in the path.


balbirthomas (approved):

balbirthomas (changes requested):
Seems to have test failure.

balbirthomas (approved):

PR #95, opened 2021-08-13
Add observe for events.relation_changed

xavpaice:
The Prometheus library needs to observe events.relation_changed as well
as events.relation_joined, in order to trigger the correct relation data
changes in PrometheusConsumer.

### DIFF
--- a/lib/charms/prometheus_k8s/v0/prometheus.py
+++ b/lib/charms/prometheus_k8s/v0/prometheus.py
@@ -317,7 +317,7 @@ LIBAPI = 0
 
 # Increment this PATCH version before using `charmcraft publish-lib` or reset
 # to 0 if you are raising the major API version
-LIBPATCH = 5
+LIBPATCH = 4
 
 
 logger = logging.getLogger(__name__)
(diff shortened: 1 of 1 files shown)


balbirthomas:
@xavpaice until now most relation data was being set in relation joined and there after it never changes. Hence I am a bit surprised for the need to handle relation changed. Never the less I am approving. It will help if you can explain (say in commit message) what is the use case that prompted you to request this change.

lib/charms/prometheus_k8s/v0/prometheus.py
```diff
@@ -317,7 +317,7 @@ def _on_scrape_targets_changed(self, event):
 
 # Increment this PATCH version before using `charmcraft publish-lib` or reset
 # to 0 if you are raising the major API version
-LIBPATCH = 4
```
balbirthomas, line 320: I would not mind if you leave it at 4. Since we have not released to edge and are still accruing features. Your call.

balbirthomas (approved):

lib/charms/prometheus_k8s/v0/prometheus.py
balbirthomas: On second thoughts I would like to request you to leave it at 4 since otherwise we will have missing version numbers in our release history for this charm library.

balbirthomas (approved):

PR #96, opened 2021-08-13
Move default alert rules dir to src/prometheus_alert_rules

mmanciop:
This commit changes the default location of the Prometheus alert rules built into charms from `<repo_root>/prometheus_alert_rules` to `<repo_root>/src/prometheus_alert_rules`.

This change simplifies the user experience with charmcraft 1.2.0+, as it avoids the need of having to explicitly mark the rule files in `charmcraft.yaml` with [`charm/prime`](https://juju.is/docs/sdk/charmcraft-config).

### DIFF
--- a/lib/charms/prometheus_k8s/v0/prometheus.py
+++ b/lib/charms/prometheus_k8s/v0/prometheus.py
@@ -247,9 +247,9 @@ related consumer charms and enabling corresponding alerts within the
 provider charm.  Alert rules are automatically gathered by Consumer
 charms when using this library, from a directory conventionally named
 `prometheus_alert_rules`. This directory must reside at the top level
-in the Consumer charm. Each file in this directory is assumed to be a
-single alert rule in YAML format. The format of this alert rule
-conforms to [Prometheus
+in the `src` folder of the consumer charm. Each file in this directory
+is assumed to be a single alert rule in YAML format. The format of this
+alert rule conforms to [Prometheus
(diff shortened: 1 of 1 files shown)


lib/charms/prometheus_k8s/v0/prometheus.py
```diff
@@ -317,7 +317,7 @@ def _on_scrape_targets_changed(self, event):
 
 # Increment this PATCH version before using `charmcraft publish-lib` or reset
 # to 0 if you are raising the major API version
-LIBPATCH = 4
+LIBPATCH = 5
```
balbirthomas, line 320: Not required since we have not yet published 4. If we push this to 5 then 4 will be missing from version history of charm library.

lib/charms/prometheus_k8s/v0/prometheus.py
```diff
@@ -713,7 +713,7 @@ def __init__(
         service_event,
         jobs=[],
         multi=False,
-        alert_rules_path="prometheus_alert_rules",
+        alert_rules_path="src/prometheus_alert_rules",
```
balbirthomas, line 716: Ideally should be `os.path.join` but I am ambivalent because I am not sure charms will ever run on any other OS other than a Linux.

balbirthomas (approved):

lib/charms/prometheus_k8s/v0/prometheus.py
mmanciop: Precisely

PR #97, opened 2021-08-13
ingress facade

sed-i:
Encapsulating all ingress related logic into a Consumer-like library.

### DIFF
--- a/src/charm.py
+++ b/src/charm.py
@@ -17,11 +17,51 @@ from charms.grafana_k8s.v1.grafana_source import GrafanaSourceConsumer
 from charms.nginx_ingress_integrator.v0.ingress import IngressRequires
 from charms.prometheus_k8s.v0.prometheus import PrometheusProvider
 from charms.alertmanager_k8s.v0.alertmanager import AlertmanagerConsumer
+from ops.relation import ConsumerBase
+from ops.framework import EventSource, EventBase
+
 
 PROMETHEUS_CONFIG = "/etc/prometheus/prometheus.yml"
 RULES_DIR = "/etc/prometheus/rules"
 
 logger = logging.getLogger(__name__)
+
+
+class IngressConsumer(ConsumerBase):
+    class IngressChanged(EventBase):
+        pass
+
+    ingress_changed = EventSource(IngressChanged)
+
+    def __init__(self, charm: CharmBase, relation_name: str, consumes: dict, multi: bool = False):
+        super().__init__(charm, relation_name, consumes, multi)
+
+        # It is recommended to default to `self.app.name` so that the external
+        # hostname will correspond to the deployed application name in the
+        # model, but allow it to be set to something specific via config.
+        self._external_hostname = charm.config["web-external-url"] or f"{charm.app.name}.juju"
+
+        self.ingress = IngressRequires(
+            charm,
+            {
+                "service-hostname": self._external_hostname,
+                "service-name": charm.app.name,
+                "service-port": str(charm.model.config["port"]),
+            },
+        )
+
+        self.framework.observe(charm.on[relation_name].relation_joined, self._on_ingress_changed)
+        self.framework.observe(charm.on[relation_name].relation_changed, self._on_ingress_changed)
+        self.framework.observe(charm.on[relation_name].relation_broken, self._on_ingress_changed)
+
+    @property
+    def external_hostname(self):
+        """Return the external hostname to be passed to ingress via the relation."""
+        return self._external_hostname
+
+    def _on_ingress_changed(self, event):
+        if event.unit:
+            self.ingress_changed.emit()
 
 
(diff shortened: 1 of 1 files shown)


src/charm.py
```diff
@@ -17,13 +17,53 @@
...
+from ops.framework import EventSource, EventBase
+
 
 PROMETHEUS_CONFIG = "/etc/prometheus/prometheus.yml"
 RULES_DIR = "/etc/prometheus/rules"
 
 logger = logging.getLogger(__name__)
 
 
+class IngressConsumer(ConsumerBase):
```
balbirthomas, line 30: Good idea. But why not create a PR against the Ingress charm. Also may be a good idea to factor this out into a separate file `src/ingress.py` and if required any unit tests for it `tests/test_ingress.py`.

mmanciop: I'd be in favor of proposing this as a PR for the upstream repo

tox.ini
```diff
@@ -29,7 +29,7 @@ deps =
     flake8
 commands =
     flake8 {toxinidir}/src {toxinidir}/tests {toxinidir}/lib/charms/prometheus_k8s
-    black --check --diff --target-version=py38 {toxinidir}/src {toxinidir}/tests {toxinidir}/lib/charms/prometheus_k8s
+    black --line-length=99 --check --diff --target-version=py38 {toxinidir}/src {toxinidir}/tests {toxinidir}/lib/charms/prometheus_k8s
```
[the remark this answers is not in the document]
sed-i: We use 99 throughout, for example
https://github.com/canonical/alertmanager-operator/blob/6e785756a98e60d005fc4ec5a561cc408d5a8668/tox.ini#L36
but it sounds like you intentionally omitted it from both flak8 and black... :)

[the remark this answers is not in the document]
balbirthomas: You seem to be referring to yourself as "we" :-) . I never remember this being a discussed.

src/charm.py
sed-i: @mthaddon @jnsgruk what do you think about adding this as a library to [charm-k8s-ingress](https://launchpad.net/charm-k8s-ingress)? ([diff ref](https://github.com/canonical/prometheus-operator/pull/97/commits/5ccc81b5736e34e8ea84afa810e6a3d45fbceda0))

tox.ini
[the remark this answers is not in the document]
sed-i: prometheus had 99 before tox migration
https://github.com/canonical/prometheus-operator/blob/34c0dca535c422029e50ba88c3897364031d42ff/.flake8#L2

other charms as well.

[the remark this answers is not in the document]
balbirthomas: Which was not default behaviour and I found myself surprised when creating tox ini that black reformatted my code. Then I realized this. In standup we had agreed to stick with black defaults.

lib/charms/prometheus_k8s/v0/prometheus.py
```diff
@@ -462,13 +458,9 @@ def alerts(self):
             if not relation.units:
                 continue
 
-            alert_rules = json.loads(
-                relation.data[relation.app].get("alert_rules", "{}")
-            )
```
jnsgruk, line 467: There is a lot of this which is clearly a `black` thing - can you seperate the work you're doing from the `black` formatting to make the diff a bit more readable?

src/charm.py
jnsgruk: What are we proposing to add? The `ingress-changed` event? If so, I don't see a strong reason why that couldn't be added to @mthaddon's upstream library.

lib/charms/prometheus_k8s/v0/prometheus.py
sed-i: it is already divided into two separate commits

tox.ini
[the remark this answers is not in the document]
balbirthomas: Ok. Let us merge the reformatting commit.

src/charm.py
sed-i: Essentially replace
`self.ingress = IngressRequires(...)`
with
`self.ingress_lib = IngressConsumer(...)`

to be more precise:
1. Add a consumer library - new class that inherits from `ops.relation.ConsumerBase`
2. Delegate all relevant event observation to the new class
3. Expect users to observe a (custom) event coming from the new lib, `self.ingress_lib.ingress_changed`

Whoever currently uses `IngressRequires` directly could still continue and do it.

lib/charms/prometheus_k8s/v0/prometheus.py
sed-i: let me drop that linting commit for now (gh action may fail)

balbirthomas:
Closing this, as it is no longer relevant.

PR #99, opened 2021-08-16
tox + black fixes

sed-i:


balbirthomas (approved):
I presuming the linting failure will not happen on the next PR.

PR #102, opened 2021-08-19
Patch K8s service to list port 9090

mmanciop:
Hook into the charm's `install` and `upgrade_charm` hooks the patching of the K8s service, so that we get the following awesomeness:

```
michele@boombox:~/git/prometheus-operator$ microk8s.kubectl get services -n prometheus-test
NAME                   TYPE        CLUSTER-IP       EXTERNAL-IP   PORT(S)     AGE
modeloperator          ClusterIP   10.152.183.212   <none>        17071/TCP   18m
prometheus-endpoints   ClusterIP   None             <none>        <none>      13m
prometheus             ClusterIP   10.152.183.88    <none>        9090/TCP    77s
```

### DIFF
--- a/src/charm.py
+++ b/src/charm.py
@@ -487,7 +487,7 @@ class PrometheusCharm(CharmBase):
         # hostname will correspond to the deployed application name in the
         # model, but allow it to be set to something specific via config.
 
-        return self.config["web-external-url"] or f"{self.app.name}.juju"
+        return self.config["web-external-url"] or f"{self.app.name}"
 
     @property
     def port(self):
--- a/src/charm.py
+++ b/src/charm.py
@@ -2,10 +2,12 @@
 # Copyright 2020 Canonical Ltd.
 # See LICENSE file for licensing details.
 
-import os
-import logging
-import yaml
 import json
+import logging
+import os
+import yaml
+
+from kubernetes_service import K8sServicePatch, PatchFailed
 
 from ops.charm import CharmBase
 from ops.framework import StoredState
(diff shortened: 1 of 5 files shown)


src/charm.py
```diff
@@ -449,6 +453,20 @@ def _prometheus_layer(self):
 
         return layer
 
+    def _patch_k8s_service(self):
+        """Fix the Kubernetes service that was setup by Juju with correct port numbers"""
+        if self.unit.is_leader() and not self._stored.k8s_service_patched:
```
balbirthomas, line 458: So every time the leader changes, k8s port will be patched again. Is this the required behaviour ?

balbirthomas (approved):

src/charm.py
mmanciop: In that case it would effectively be a no-op, cuz there is nothing to patch :-)

src/charm.py
```diff
@@ -462,6 +480,20 @@ def version(self):
...
         return None
 
+    @property
+    def _external_hostname(self):
+        """Return the external hostname to be passed to ingress via the relation."""
+        # It is recommended to default to `self.app.name` so that the external
+        # hostname will correspond to the deployed application name in the
+        # model, but allow it to be set to something specific via config.
+
+        return self.config["web-external-url"] or f"{self.app.name}.juju"
```
jnsgruk, line 490:
```suggestion
        return self.config["web-external-url"] or f"{self.app.name}"
```

jnsgruk: We sort of stopped doing the `.juju` thing by consensus some time ago... might just be worth grepping the repo to make sure this change doesn't break some docs/tests

jnsgruk (approved):

src/charm.py
sed-i: It would be nice though to have the patch logic check if it is already patched instead of removing and creating a new service each time.

PR #106, opened 2021-08-20
Flip provider and consumer

balbirthomas:
This pull request reverses the roles of Prometheus provider and consumer and renames the provider consumer relation.


CONTRIBUTING.md
```diff
@@ -97,28 +97,21 @@ validates all configurations options when provided before generating
...
-interface is limited to only those aspects of Prometheus required by
-this charm.
+[charm library](lib/charms/prometheus_k8s/v1/prometheus.py).
 
 ### Library Details
 
-The Prometheus charm library exposes a provider and consumer object -
-`PrometheusProvider` and `PrometheusConsumer` along with the custom
+The Prometheus charm library exposes a consumer and provider object -
+`MetricsConsumer` and `MetricsProvider` along with the custom
```
mmanciop, line 105: I thought we'd call them "MetricsEndpoint{Provider|Consumer}"

CONTRIBUTING.md
```diff
@@ -97,28 +97,21 @@ validates all configurations options when provided before generating
...
 charm event `TargetsChanged` within the `MonitoringEvents` event
-descriptor. `PrometheusProvider` emits the `TargetsChagned` event in
+descriptor. `MetricsConsumer` emits the `TargetsChagned` event in
```
mmanciop, line 107: `TargetsChagned` is mispelled

CONTRIBUTING.md
```diff
@@ -97,28 +97,21 @@ validates all configurations options when provided before generating
...
-library requires that Prometheus informs its "Consumers" (scrape targets)
-of the actual Prometheus version that was deployed. In order to determine
-this version at runtime `PrometheuCharm` uses the
-[`Prometheus`](src/prometheus_server.py) object. The `Prometheus`
-object provides an interface to a running Prometheus instance. This
-interface is limited to only those aspects of Prometheus required by
-this charm.
+[charm library](lib/charms/prometheus_k8s/v1/prometheus.py).
 
 ### Library Details
```
mmanciop, line 102: This section should IMO explain what the library is for (enabling providers to have their metrics endpoints scraped by the `prometheus-k8s` charm, but it does not "early enough".

CONTRIBUTING.md
balbirthomas: The relation is called `metrics-endpoint`. Since these classes derive form `ProviderBase` and `ConsumerBase` they are named `MetricsProvider` and `MetricsConsmer` .

mmanciop (changes requested):
Code looks fine, besides the naming, but I think the docu needs love

CONTRIBUTING.md
balbirthomas: Ack. Will fix.

CONTRIBUTING.md
balbirthomas: It is metrics that is being provided and consumed. Isn't it ?

dstathis (approved):

sed-i (approved):

CONTRIBUTING.md
mmanciop: I think I understand what you went for, but if we wanna go philosophical, it is not metrics that are being provided and consumed: it is metrics endpoints. The metrics are the outcome of consuming the metrics endpoint, which happens on a completely different level, outside of Juju. If the actual metrics data were somehow shared over relations (horrible, horrible idea that nobody should even consider), then it would be `MetricsProvider` and `MetricsConsumer`.

PR #107, opened 2021-08-20
Improve structure and documentation of *.rule files

### LINKED ISSUE #103: Alert rules should not require a list item per file
We are asking to have one list item per `.rule` file. Since it is supposed to be one YAML object per file anyhow, what is the value of having a list item in there in the first place? If anything, that would bait people in trying to set multiple alert_rules in one file, and have [all after the first silently ignored](https://github.com/canonical/prometheus-operator/blob/d29f8388f545ca39860afeee981165b73450cf87/lib/charms/prometheus_k8s/v0/prometheus.py#L874).

mmanciop:
Instead of expecting singleton rules in a `*.rule` file, simply expect the alert_rule object.

Also, add more logging for alert rules and ensure that the reload of alert_rules is triggered on alert reload.

(Pending: Avoid unnecessary rule reloads)

Closes #103

### DIFF
--- a/lib/charms/prometheus_k8s/v0/prometheus.py
+++ b/lib/charms/prometheus_k8s/v0/prometheus.py
@@ -231,21 +231,21 @@ related `MetricsEndpointProvider` charms and enabling corresponding alerts withi
 Prometheus charm.  Alert rules are automatically gathered by `MetricsEndpointProvider`
 charms when using this library, from a directory conventionally named
 `prometheus_alert_rules`. This directory must reside at the top level
-in the `src` folder of the provider charm. Each file in this directory
-is assumed to be a single alert rule in YAML format. The format of this
-alert rule conforms to [Prometheus
-documentation](https://prometheus.io/docs/prometheus/latest/configuration/alerting_rules/).
+in the `src` folder of the consumer charm. Each file in this directory
+is assumed to be a single alert rule in YAML format. The file name must
+have the `.rule` extension. The format of this alert rule conforms to the
+[Prometheus documentation](https://prometheus.io/docs/prometheus/latest/configuration/alerting_rules/).
 An example of the contents of one such file is shown below.
 
 ```
-- alert: HighRequestLatency
-  expr: job:request_latency_seconds:mean5m{my_key=my_value, %%juju_topology%%} > 0.5
-  for: 10m
-  labels:
-    severity: Medium
-    type: HighLatency
-  annotations:
-    summary: High request latency for {{ $labels.instance }}.
+alert: HighRequestLatency
+expr: job:request_latency_seconds:mean5m{my_key=my_value, %%juju_topology%%} > 0.5
+for: 10m
+labels:
+  severity: Medium
+  type: HighLatency
+annotations:
+  summary: High request latency for {{ $labels.instance }}.
 ```
 
 It is **very important** to note the `%%juju_topology%%` filter in the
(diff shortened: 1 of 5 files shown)


src/charm.py
```diff
@@ -176,22 +180,35 @@ def _set_alerts(self, container):
...
-            try:
-                path = os.path.join(RULES_DIR, filename)
-                rules = yaml.dump({"groups": alerts["groups"]})
-                logger.debug("Rules for relation %s : %s", rel_id, rules)
-                container.push(path, rules, make_dirs=True)
-            except ConnectionError:
-                logger.error("Failed to write alert rule %s", filename)
+        container.remove_path(RULES_DIR, recursive=True)
+
+        alert_rules_by_rel_id = self.prometheus_provider.alerts()
```
balbirthomas, line 190: I presume this is just a style change. If so I like it because it makes more explicit the return value that is being iterated on.

lib/charms/prometheus_k8s/v0/prometheus.py
```diff
@@ -870,8 +870,7 @@ def _labeled_alert_groups(self):
             with path.open() as rule_file:
                 # Load a list of rules from file then add labels and filters
                 try:
-                    rules = yaml.safe_load(rule_file)
-                    rule = rules[0]  # each file is list of one rule
+                    rule = yaml.safe_load(rule_file)
```
balbirthomas, line 873: I am a bit worried about this. Because now, the Prometheus charm library (in particular the aggregator) will have to support two different formats for alerts rules - one for Bootstack team and one for every one else. This is of course possible and may be done with just a conditional here or there. If there is a more universal approach that will be ideal.

src/charm.py
```diff
@@ -116,14 +116,18 @@ def _configure(self):
         logger.info("Configuring Prometheus")
         container = self.unit.get_container("prometheus")
 
+        if not container.is_ready():
```
balbirthomas, line 119: I think it makes sense to do this check right at the beginning and abort.  This does presume when the container gets ready a pebble ready event will be triggered and the Prometheus configuration will be updated. I think this is correct but I will give it a second thought.

balbirthomas (approved):

src/charm.py
```diff
@@ -116,14 +116,18 @@ def _configure(self):
...
         # check if configuration file has changed and if so push the
         # new config file to the workload container
         prometheus_config = self._prometheus_config()
         try:
             container.push(PROMETHEUS_CONFIG, prometheus_config)
             logger.info("Pushed new configuration and alerts")
         except ConnectionError:
-            logger.info("Ignoring config changed since pebble is not ready")
+            logger.info("Ignoring config and alert_rules changes since pebble is not ready")
             return
```
jnsgruk, line 131: I would encourage:

```python
with container.is_ready():
    container.push(....)
    logger.info(...)
```

src/charm.py
mmanciop: No, the translation from machine charm alert rules to these is in the lmaproxy

mmanciop: Is == will be

lib/charms/prometheus_k8s/v0/prometheus.py
mmanciop: No need to be concerned, the translation from machine charm alert rules to these will be in the LMAProxy

src/charm.py
mmanciop: good idea

src/charm.py
```diff
@@ -176,22 +180,34 @@ def _set_alerts(self, container):
...
-            try:
-                path = os.path.join(RULES_DIR, filename)
-                rules = yaml.dump({"groups": alerts["groups"]})
-                logger.debug("Rules for relation %s : %s", rel_id, rules)
-                container.push(path, rules, make_dirs=True)
-            except ConnectionError:
-                logger.error("Failed to write alert rule %s", filename)
+            alert_rules_by_rel_id = self.prometheus_provider.alerts()
+
+            if alert_rules_by_rel_id:
```
balbirthomas, line 193: This if condition is not necessary. `alerts()` returns and empty dictionary if no alerts are available. Hence the for loop iteration will always work but in the case of an empty dictionary no iterations will occur (same as if condition).

balbirthomas (approved):

jnsgruk (approved):

PR #122, opened 2021-09-14
Switched to new container can connect API

balbirthomas:
This commit replace the obsolete container.is_ready() API
with the container.can_connect() method. It is just a single
line change despite appearences, because all the code that
was previously nested in the context manager is now unindented
one level but guarded with the can_connect() method.

### DIFF
--- a/src/charm.py
+++ b/src/charm.py
@@ -102,30 +102,31 @@ class PrometheusCharm(CharmBase):
         """
         container = self.unit.get_container(self._name)
 
-        with container.is_ready():
+        if not container.can_connect():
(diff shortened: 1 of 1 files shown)


jnsgruk (approved):

jnsgruk:
You could consider adding a `self.unit.status = WaitingStatus("waiting for Pebble API")` before the early return

balbirthomas:
> You could consider adding a `self.unit.status = WaitingStatus("waiting for Pebble API")` before the early return

Good point will add it into next PR.

PR #126, opened 2021-09-20
add pointers for manual testing

sed-i:


CONTRIBUTING.md
```diff
@@ -83,6 +83,16 @@ $ tox -e unit
 
 It is expected that unit tests should provide at least 80% code coverage.
 
+### Manual testing
```
balbirthomas, line 86: I do not have strong opinions here but I do not think it makes much sense to provide instructions on how to use  Prometheus or its API in `CONTRIBUTING.md`. I would expect anyone who is going to make a PR against this repo has spent the time to study the [Prometheus docs](https://prometheus.io/docs/) well enough.

sed-i: I agree, I would put this in README or TESTING (if we had one). However, the "Testing" section is in CONTRIBUTING.
That being said, even after being familiar with the docs, it is not immediately obvious how to quickly probe the endpoints.

balbirthomas: [`src/prometheus_server.py`](https://github.com/canonical/prometheus-operator/blob/main/src/prometheus_server.py#L56) already has an example of using the Prometheus API. Anyone working on this charm should know how to query the Prometheus API. I see know point in re-documenting or exemplifying the use of the Prometheus API which is already documented.

sed-i: Should add `api/v1/status/buildinfo` to the list of "poorly documented endpoints"

dstathis: I agree, it seems like you are going to have to look up the Prometheus docs anyway so I don't really see the need for this.

sed-i:
I captured the proposed changes in a [gist](https://gist.github.com/sed-i/76e846846deb2eff614c7aa771aa18a1). Closing.

PR #127, opened 2021-09-20
Additional Scrape configuration options

balbirthomas:
This PR enhances the Prometheus charm library to support additional scrape configuration options.

### DIFF
--- a/lib/charms/prometheus_k8s/v0/prometheus_scrape.py
+++ b/lib/charms/prometheus_k8s/v0/prometheus_scrape.py
@@ -315,7 +315,7 @@ LIBPATCH = 6
 logger = logging.getLogger(__name__)
 
 
-JOB_KEYS = {
+ALLOWED_KEYS = {
     "job_name",
     "metrics_path",
     "static_configs",
@@ -356,7 +356,7 @@ def _sanitize_scrape_configuration(job) -> dict:
         a dictionary containing a sanitized job specification.
     """
     sanitized_job = DEFAULT_JOB.copy()
-    sanitized_job.update({key: value for key, value in job.items() if key in JOB_KEYS})
+    sanitized_job.update({key: value for key, value in job.items() if key in ALLOWED_KEYS})
     return sanitized_job
 
 
--- a/lib/charms/prometheus_k8s/v0/prometheus_scrape.py
+++ b/lib/charms/prometheus_k8s/v0/prometheus_scrape.py
@@ -591,7 +591,8 @@ class MetricsEndpointConsumer(Object):
         name = job.get("job_name")
         job_name = f"{job_name_prefix}_{name}" if name else job_name_prefix
 
-        labeled_job = {"job_name": job_name, "metrics_path": job["metrics_path"]}
+        labeled_job = job.copy()
+        labeled_job["job_name"] = job_name
 
         static_configs = job.get("static_configs")
         labeled_job["static_configs"] = []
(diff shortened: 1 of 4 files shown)


lib/charms/prometheus_k8s/v0/prometheus_scrape.py
```diff
@@ -300,21 +300,49 @@ def _on_scrape_targets_changed(self, event):
...
         a dictionary containing a sanitized job specification.
     """
-    return {
-        "job_name": job.get("job_name"),
-        "metrics_path": job.get("metrics_path", "/metrics"),
-        "static_configs": job.get("static_configs", [{"targets": ["*:80"]}]),
-    }
+    sanitized_job = DEFAULT_JOB.copy()
+    sanitized_job.update({key: value for key, value in job.items() if key in JOB_KEYS})
+    return sanitized_job
```
sed-i, line 345: add `logger.debug` for dropped (invalid) keys?

balbirthomas: Could do but I plan to document the allowed keys explicitly so there should be no surprises that keys are dropped.

balbirthomas: I am in process of explicitly documenting the allowed keys so there should be no surprises that keys are not showing up. There is already a unit test to validate that only allowed keys are put into scrape job configuration.

sed-i: unit tests here won't bubble this up when used by another charm.
a log entry would be quite valuable for debugging when a user relates their charm and nothing happens.

balbirthomas: I just pushed the updated docstring with list of allowed keys. I fail to see what is the point of logging rejected keys when it is explicitly documented that only a fixed list of keys are allowed.

sed-i:
Would having `self.metrics_endpoint = MetricsEndpointProvider(...)` in my charm's constructor cause prometheus to reload/restart every time my charm instantiates?

src/charm.py
```diff
@@ -211,8 +211,6 @@ def _is_valid_timespec(self, timeval) -> bool:
         Returns:
             True if time specification is valid and False otherwise.
         """
-        matched = re.match(r"[1-9][0-9]*[ymwdhs]", timeval) is not None
-
         if not (matched := re.match(r"[1-9][0-9]*[ymwdhs]", timeval)):
```
sed-i, line 214: tackle #128 while you're at it?

balbirthomas: I did think this through. The docs are not right. Try intervals starting with zero. For instance "0s". In any case I do not want to allow such intervals. I will be closing that ticket for the same reason.

lib/charms/prometheus_k8s/v0/prometheus_scrape.py
```diff
@@ -300,21 +315,49 @@ def _on_scrape_targets_changed(self, event):
 logger = logging.getLogger(__name__)
 
 
+JOB_KEYS = {
```
sed-i, line 318: Is it necessary to be an intermediary? Could just pass them all as they are (and remove this list entirely).
Currently you do check validity of duration,but there are many other things that would make a configuration invalid.

balbirthomas: I would prefer keeping the list of allowed job keys as a global constant. This is and important data structure for the entire library. This also allows reuse if required and keeps my santization function small. Prometheus charm does do validation.

src/charm.py
sed-i: The existing regex does not support expression such as `1h30m` though.
It is odd however to silently ignore user input in the charm (e.g `0`) that is permissible by official docs.

sed-i: you mean prom supports expressions such as `3600s` and `90m`?
anyway, this is a minor thing and I think it would be best to be consistent with prom.

balbirthomas: Yes. Exactly.

sed-i (approved):

mmanciop (approved):

PR #129, opened 2021-09-21
add scrape interval to configs

sed-i:


jnsgruk:
Hmm, we removed this not long ago. I thought @balbirthomas added a default here. My view is there should be a sensible default, or the scrape interval should just be specified by a specific scrape config

balbirthomas:
@leon you can use [this branch](https://github.com/canonical/prometheus-operator/pull/127) to specify the scrape interval of jobs as discussed yesterday. Scrape interval was removed as a config option after the last code review.

sed-i:
The reason I added the globals was that the following had no effect on the scrape settings:

```python
        self.metrics_endpoint = MetricsEndpointProvider(
            self,
            "metrics-endpoint",
            self.on.avalanche_pebble_ready,
            jobs=[
                {
                    "job_name": self.unit.name,
                    "metrics_path": "/metrics",
                    "static_configs": [{"targets": [f"*:{self.port}"]}],
                    "scrape_interval": "1s",
                    "scrape_timeout": "1s",
                }
            ],
        )
```

I see in the docs that jobs settings override global settings, so maybe there is a bug in the current state of the branch?

balbirthomas:
> I see in the docs that jobs settings override global settings, so maybe there is a bug in the current state of the branch?

If there is a bug then that can be addressed. But if you want to bring back configuration options that were explicitly removed then perhaps this needs to be discussed in the team. Besides doesn't our `CONTRIBUTING.md` suggest that such a discussion is advisable when changing feature sets.

balbirthomas:
@sed-i The bug you reported should now be fixed (see attached screenshot). Feel free to create a bug report if you find any other issues that are blocking you. I am closing this PR now as I believe your use case for performance benchmarking should be unblocked. 
Screenshot from 2021-09-21 14-40-04

PR #130, opened 2021-09-22
first pass at remote write still needs work

dstathis:


dstathis:
For some reason I thought the ingress addres field in the relation data was the bind address. This is not the case so I need to change things up a bit. Here is what I have so far.

metadata.yaml
```diff
@@ -24,6 +24,8 @@ containers:
 provides:
   grafana-source:
     interface: grafana_datasource
+  remote-write-server:
```
balbirthomas, line 27: Nitpick: Is `remote-write-server` the agreed upon relation name ? I was wondering what does `remote-write-sink` sound like, keeping in with the tradition established by `grafana-source` ?

dstathis: The idea here is that at some point Prometheus might both require and provide this interface, at which point we would add remote-write-client. I think that is the most clear way to do it but I am happy to change it to source and sink if you would like.

rbarry82: On the one hand, `grafana-source` has only stayed stable by luck of the draw rather than any well-thought-out naming 😉

On the other hand, if this is Prometheus only, I’d make it very explicit and just call it `promtail `

lib/charms/prometheus_k8s/v0/prometheus_remote_write.py
```diff
@@ -0,0 +1,93 @@
...
+            A list of remote write endpoints.
+        """
+        for relation in self.model.relations[self._relation_name]:
+            for unit in relation.units:
+                # If external-address is provided use that, else use ingress-address
+                if (address := relation.data[unit].get("address")) is None:
+                    continue
+                if (port := relation.data[unit].get("port")) is None:
+                    continue
+                yield f"http://{address}:{port}/api/v1/write"
```
Abuelodelanada, line 57: What are you going to `yield` if `address` or `port` are not in `relation.data[unit]` ?

Abuelodelanada, line 57: Maybe it's a very weird corner case, but if I understand correctly, we are going to `yield` nothing if `address` or `port` are sequentially not present in `relation.data[unit]` 
Is that correct?

src/charm.py
```diff
@@ -137,8 +137,19 @@ def _configure(self, _):
...
             logger.info("Prometheus (re)started")
 
+        # Provide connection info to any remote write clients.
+        self._set_remote_write()
+
         self.unit.status = ActiveStatus()
 
+    def _set_remote_write(self):
+        """Provide connection info to any remote write clients."""
+        if self.model.relations["ingress"] and self.app.planned_units() == 1:
```
balbirthomas, line 147: It is not obvious why you need `self.planned_units()` ? Could you kindly elaborate upon this in docstring or code comment.

src/charm.py
```diff
@@ -137,8 +137,19 @@ def _configure(self, _):
...
+            self.remote_write.set_endpoint(address=self._external_hostname, port=self._port)
```
balbirthomas, line 148: Since `self._external_hostname` is currently a config option, does this mean in the future we will need to replace this by a more appropriate solution that will work cross model etc. If so feel free to put a TODO reminder here that is an indication of this technical debt. Or alternatively kindly create a github issue for Prometheus providing context and explaining what kind of changes may be required in the future.

tests/test_remote_write.py
```diff
@@ -37,27 +37,19 @@ def setUp(self):
     def test_addr(self):
```
balbirthomas, line 37: Even though this has not changed in this PR, i think a name such as `test_addr` is not specific enough. Some thing like this : "test_address_can_be_set" is more indicative of the nature of the test. Same comment for other tests. An aleternative (though not self-documenting) is to put a comment. It should not be necessary to read the whole test just to figure out what the test is testing.

src/charm.py
```diff
@@ -127,8 +137,19 @@ def _configure(self, _):
             container.restart(self._name)
             logger.info("Prometheus (re)started")
 
+        # Provide connection info to any remote write clients.
+        self._set_remote_write()
```
balbirthomas, line 141: The name of this function is a bit misleading. As I understand you are "setting" remote write (or enabling remote write) though the "self._command()" method by adding an additional command line option. This function is providing and address for the remote write client. If you agree wouldn't this method be better named as `self._set_remote_write_address()` or something like that. 

Also why does this need to happen in the charm ? I fail to see why the library can not do this and thereby make it transparent to the charm. In fact I would have expected that the library sets the remote write address in response to a relation joined with the remote write client.

balbirthomas (approved):

metadata.yaml
balbirthomas: Please do check with @mmanciop what his preferences here are.

mmanciop: I think the intuition is good, but let’s call it `receive-remote-write`

lib/charms/prometheus_k8s/v0/prometheus_remote_write.py
dstathis: Yes that's the idea. If we don't have a working address/port combination, don't return an endpoint.

src/charm.py
dstathis: Oh yes I will add a comment. Ingress won't work with more than one unit. I will go in to more depth in the comment.

src/charm.py
dstathis: I'm not sure I understand. Will the _external_hostname() method not always do what we want here? Could you elaborate on the issue?

src/charm.py
mmanciop: Don’t forget the docu in README. That limitation hits juju admins way more than charm authors

dstathis: First paragraph: I agree completely
Second paragraph: I will deliberate

lib/charms/prometheus_k8s/v0/prometheus_remote_write.py
```diff
@@ -0,0 +1,93 @@
...
+
+    @property
+    def configs(self) -> Generator[dict, None, None]:
```
sed-i, line 60: I am not convinced a generator is justifiable.
What is the length you're expecting to see here?

lib/charms/prometheus_k8s/v0/prometheus_remote_write.py
```diff
@@ -0,0 +1,93 @@
...
+
+        Returns:
+            A list of remote_write configs.
+        """
+        for endpoint in self.endpoints:
+            yield {"url": endpoint}
+
+
+class PrometheusRemoteWriteProvider(Object):
+    """A prometheus remote write provider."""
```
sed-i, line 71: docstring style: move from `__init__`?

src/charm.py
balbirthomas: I may be wrong. But `_external_hostname()` is a bodge because we have no reliable way currently for Juju to tell us what is the externally routable domain name or IP address of any given service or application. Hence `_external_hostname()` was added as a configuration option. Once Juju provides us with this feature then I presume both the remote write library and this method will use whatever Juju provides. Check with @mmanciop who may have better insight into this.

dstathis: I'm not sure juju will ever be able to provide us with this. If it can though, presumably the method `_external_hostname()` can would just change it's logic to use that instead of the config value though. Am I missing something?

src/charm.py
balbirthomas: If it will not work with **more** than one unit shouldn't you be checking the total number of units rather than just "planned_unit" . Also this is quite surprising. I do not see why Ingress will not work with more than one unit. Pray, illuminate me and help me get rid of my ignorance here.

lib/charms/prometheus_k8s/v0/prometheus_remote_write.py
```diff
@@ -0,0 +1,93 @@
...
+
+        Args:
+            charm: The charm object that instantiated this class.
+            relation_name: The relation name as defined in metadata.yaml.
+        """
+        super().__init__(charm, relation_name)
+        self._charm = charm
+        self._relation_name = relation_name
+
+    def set_endpoint(self, address: str, port: Union[str, int]) -> None:
```
sed-i, line 84: I would type hint the port as int. It would still work with str if used by mistake, and provide the user better warnings overall: with str in there, if the user passes 'bla' it won't be caught in static analysis.

lib/charms/prometheus_k8s/v0/prometheus_remote_write.py
```diff
@@ -0,0 +1,93 @@
...
+        """Set the address and port on which you will serve prometheus remote write.
+
+        Args:
+            address: The address of the remote write server
+            port: The port number
+        """
+        for relation in self.model.relations[self._relation_name]:
+            relation.data[self._charm.unit]["address"] = address
+            relation.data[self._charm.unit]["port"] = str(port)
```
sed-i, line 93: Consider fusing them here, unless the consumer specifically needs that information in parts.
```suggestion
            relation.data[self._charm.unit]["address"] = f"{address}:{port}"
```

src/charm.py
sed-i: > Ingress won't work with more than one unit

In that case I would add a leader guard - for a less messy scaledown.
But this is a suspicious limitation. What is supposed to happen to prometheus unit count in HA?

src/charm.py
balbirthomas: `_external_hostname()` is just taking a name from a config option and passing it to the Ingress charm, which exposes that name AFAIU to outside the model. All I was asking is this what is intended now and in the future. Perhaps I am going of on a tangent here. Just ignore the rest.

src/charm.py
balbirthomas: Ah I see. You mean with more than one unit of Prometheus. Indeed. In that case how are you informing the system administrator that something has gone wrong and he/she needs to do something about it ?

src/charm.py
```diff
@@ -56,6 +59,11 @@ def __init__(self, *args):
         # Gathers scrape job information from metrics endpoints
         self.metrics_consumer = MetricsEndpointConsumer(self, "metrics-endpoint")
 
+        # Exposes remote write endpoints
+        self.remote_write = PrometheusRemoteWriteProvider(
+            self, relation_name="remote-write-server"
+        )
```
sed-i, line 65: `self.remote_write_provider.set_endpoint` (with the `_provider` suffix) seems less confusing to me.

src/charm.py
```diff
@@ -127,8 +137,19 @@ def _configure(self, _):
...
+
         self.unit.status = ActiveStatus()
 
+    def _set_remote_write(self):
+        """Provide connection info to any remote write clients."""
+        if self.model.relations["ingress"] and self.app.planned_units() == 1:
+            self.remote_write.set_endpoint(address=self._external_hostname, port=self._port)
+        else:
+            address = f"{self.unit.name.replace('/','-')}.{self.app.name}-endpoints.{self.model.name}.svc.cluster.local"
+            self.remote_write.set_endpoint(address=address, port=self._port)
```
sed-i, line 151: I suspect it is impossible to differentiate ingress from cluster.local this way:
A prometheus unit may have an ingress relation for outsiders to use remote write, and at the same time there may be local charms that would like to use remote write.

src/charm.py
```diff
@@ -170,6 +191,10 @@ def _command(self) -> str:
 
             args.append(f"--web.external-url={external_url}")
 
+        # enable remote write if an instance of the relation exists
+        if self.model.relations["remote-write-server"]:
```
sed-i, line 195: iirc, this can raise KeyError when no relation present.
```suggestion
        if "remote-write-server" in self.model.relations:
```

src/charm.py
```diff
@@ -170,6 +191,10 @@ def _command(self) -> str:
...
+            args.append("--enable-feature=remote-write-receiver")
```
sed-i, line 196: Why condition this in the first place? Why not enable it all the time regardless?

sed-i: Add `--storage.remote.*` to config.yaml?

tests/test_remote_write.py
```diff
@@ -0,0 +1,79 @@
...
+        self.harness = Harness(RemoteWriteConsumerCharm, meta=METADATA)
+        self.addCleanup(self.harness.cleanup)
+        self.harness.begin()
+
+    def test_addr(self):
+        rel_id = self.harness.add_relation(RELATION, "provider")
+        self.harness.add_relation_unit(rel_id, "provider/0")
+        self.harness.update_relation_data(rel_id, "provider/0", {"address": "1.1.1.1"})
+        self.harness.update_relation_data(rel_id, "provider/0", {"port": "9090"})
+        assert list(self.harness.charm.provider.endpoints) == ["http://1.1.1.1:9090/api/v1/write"]
```
sed-i, line 42: why pytest syntax?
```suggestion
        self.assertListEqual(self.harness.charm.provider.endpoints, ["http://1.1.1.1:9090/api/v1/write"])
```

lib/charms/prometheus_k8s/v0/prometheus_remote_write.py
```diff
@@ -0,0 +1,93 @@
...
+"""
+
+from typing import Generator, Union
+
+from ops.charm import CharmBase
+from ops.framework import Object
+
+
+class PrometheusRemoteWriteConsumer(Object):
+    """A prometheus remote write consumer."""
```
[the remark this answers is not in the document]
dstathis: That is not the pattern I have seen. See prometheus_scrape library.

lib/charms/prometheus_k8s/v0/prometheus_remote_write.py
```diff
@@ -0,0 +1,93 @@
...
+
+        Args:
+            charm: The charm object that instantiated this class.
+            relation_name: The relation name as defined in metadata.yaml.
+        """
+        super().__init__(charm, relation_name)
+        self._relation_name = relation_name
+
+    @property
+    def endpoints(self) -> Generator[str, None, None]:
```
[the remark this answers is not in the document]
dstathis: Generators are more flexible and pythonic. No other reason.

lib/charms/prometheus_k8s/v0/prometheus_remote_write.py
dstathis: See above comment.

lib/charms/prometheus_k8s/v0/prometheus_remote_write.py
dstathis: You want to not allow `set_endpoint("hostname", "9090")`?

src/charm.py
dstathis: It is my understanding that if an ingress exists, it should be routable from within the cluster. If not then this needs to get a lot more complicated including config variables in the consumer.

src/charm.py
dstathis: I'm pretty sure it doesn't.

src/charm.py
dstathis: We discussed this and wanted it to be conditional. I am open to changing it back.
What is `--storage.remote.*`?

tests/test_remote_write.py
dstathis: yup yours is better here

src/charm.py
mmanciop: This is not universally true, but this risk is one I accept

lib/charms/prometheus_k8s/v0/prometheus_remote_write.py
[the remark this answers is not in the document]
sed-i: Seems like an odd choice given that the only usage of it in the code here is inside list(...).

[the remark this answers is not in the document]
sed-i: > Generators are more flexible and pythonic.

Could you please share a reference?

lib/charms/prometheus_k8s/v0/prometheus_remote_write.py
sed-i: Since "not allow" doesn't exist in python I'd say "discourage", yes. But clearly "9090" would work at runtime just as fine as `9090`. And if you'd have to `int()` the arg before calling `set_endpoint` I'd argue it's worth it.

src/charm.py
sed-i: Having it conditional has the drawback of downtime due to prometheus restart. What are the advantages of having it conditional?

Prometheus command line arguments:
```
      --storage.remote.flush-deadline=<duration>  
                                 How long to wait flushing sample on shutdown or config reload.
      --storage.remote.read-sample-limit=5e7  
                                 Maximum overall number of samples to return via the remote read interface, in a single query. 0 means no limit. This limit is ignored for streamed response
                                 types.
      --storage.remote.read-concurrent-limit=10  
                                 Maximum number of concurrent remote read calls. 0 means no limit.
      --storage.remote.read-max-bytes-in-frame=1048576  
                                 Maximum number of bytes in a single frame for streaming remote read response types before marshalling. Note that client might have limit on frame size as
                                 well. 1MB as recommended by protobuf by default.
```

balbirthomas: Some use cases may want to secure their deployments. Remote write opens a port into which you can feed any junk data you like as long as it is in openmetrics format. Keeping it conditional seems like the right choice here.

sed-i: Without a relation there is no ingress, so you're safe against malicious externals; with a relation you have an ingress so you're vulnerable again. So seems like conditional remote write has no added security.
Besides, aren't charm relations themselves quite promiscuous as it is?

src/charm.py
dstathis: The issue here is that ingress doesn't allow you to address individual pods which remote write requires. I don't see how a leader guard would have anything to do with this. The issue with detecting the issue is there is no way to determine if the related app can route to you. I will put a warning in the logs but I'm afraid that is all that can be done without resorting to messy config variables.

tests/test_remote_write.py
dstathis: I will fix this

src/charm.py
balbirthomas: I would set BlockedStatus since the admin is more likely to view juju status rather than logs. But more importantly there seems the need to also put a TODO code comment here and create a github issue. This is because as I understand you are saying this is an incomplete solution that only works with a single mongodb unit and breaks down the moment you scale up. This may not be a solveable problem now and may require support from Juju/Ops but at least we should note it down that we have this technical debt lest we forget.
@mmanciop  WDYT.

sed-i: Might be worth it to look into improving the ingress charm.

lib/charms/prometheus_k8s/v0/prometheus_remote_write.py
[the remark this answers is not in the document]
dstathis: > > Generators are more flexible and pythonic.
> 
> Could you please share a reference?

Just my experience. Can others comment?

lib/charms/prometheus_k8s/v0/prometheus_remote_write.py
dstathis: I don't really see the benefit of this. Just to prevent people from putting junk in the port? If you set port to "bla" and it doesn't work, you shouldn't really be surprised by that.

lib/charms/prometheus_k8s/v0/prometheus_remote_write.py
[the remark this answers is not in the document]
balbirthomas: I always hesitate to mandate a programming style particularly on a talented programmer such as you :-) . However  you know I also like to eschew any feature that is absolutely not necessary. Generators are useful when you want to implement a "streaming" logic for instance for "online processing" of large datasets. You may want to ask yourself. Is that your use case here.

lib/charms/prometheus_k8s/v0/prometheus_remote_write.py
balbirthomas: It may be Ok to have two docstrings. Class level docstrings explain responsibility of the class and example of how to use the class beyond instantiating it. Constructor docstring can just focus on arguments of constructor, and perhaps how to instantiate the class. WDYT.

src/charm.py
dstathis: I agree

src/charm.py
dstathis: No the port is still open without an ingress relation. While I agree that it doesn't matter in practice. When it comes to security, people tend to care about "in theory"

src/charm.py
dstathis: Given discussion on the way we use ingress with remote write in prometheus, I think it makes sense to leave this up to the charm. The implementation here is very use case specific and I think two charms using this library are not particularly likely to use the same logic. Thoughts @balbirthomas?

lib/charms/prometheus_k8s/v0/prometheus_remote_write.py
```diff
@@ -0,0 +1,93 @@
+# Copyright 2021 Canonical Ltd.
+# See LICENSE file for licensing details.
+"""This library facilitates the prometheus_remote_write interface.
+
+For the provider charm no communication is necesssary. For the consumer charm,
```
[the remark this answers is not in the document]
dstathis: The information appears to be out of date as well. I will revise this.

lib/charms/prometheus_k8s/v0/prometheus_remote_write.py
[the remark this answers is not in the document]
dstathis: Not mine but might be others. Upon further reflection, though, I see that is unlikely. I'll change it to a list.

lib/charms/prometheus_k8s/v0/prometheus_remote_write.py
dstathis: To me I like putting stuff about the class in the class docstring and stuff about the constructor in the constructor docstring. Just makes more sense to me. Definitely not a big deal either way though.

lib/charms/prometheus_k8s/v0/prometheus_remote_write.py
dstathis: Is there benefit to fusing them? In particular, does multiple writes here mean multiple relation changed triggers? If so, what about `relation.data[self._charm.unit]["endpoint"] = {"address": address, "port": port}`?

src/charm.py
balbirthomas: Your call @dstathis . Personally I would ensure as much as possible the entire responsibility of managing the relation is within the library. This way you docs can proudly say "Hey look using this library is easy - just instantiate it passing these constructor arguments, and go for a picnic" :-).

lib/charms/prometheus_k8s/v0/prometheus_remote_write.py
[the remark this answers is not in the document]
balbirthomas: There is way way more coolness in the elegance of simplicity. :-)

lib/charms/prometheus_k8s/v0/prometheus_remote_write.py
sed-i: iiuc, this will trigger only one relation changed. That is not the issue. The question is if it useful or useless to have them fused or in parts. What good is an IP address without the port number? And if your provider populates that data, why not do it right at the source?

src/charm.py
dstathis: My concern here is that the default behavior will be relevant to only prometheus-k8s/grafana-agent-k8s and anyone using the library will have to define this anyway. For example the docs will have to read "It's really easy unless you want a different behavior regarding ingress, then you need to pass a custom method to the constructor".

balbirthomas: 1. Even if the default behaviour is relevant only to grafana agent the fact is AFAIK grafana agent is the dominant use case for this library. 
2. By putting it in the library you get the opportunity to also document it in **one and the same place**, how address selection process works by default and how rest of the library woks. I view this as another reason to put it into the library. Of course you can mention in passing that if this default behaviour is not desired then the library object can be derived and extended.
3. Having all the behaviour of the remote write interface located in **one and the same place** will be the more maintainable solution in the long run.

metadata.yaml
dstathis: done

lib/charms/prometheus_k8s/v0/prometheus_remote_write.py
dstathis: @sed-i are you okay with me keeping this as is?

lib/charms/prometheus_k8s/v0/prometheus_remote_write.py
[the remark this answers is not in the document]
balbirthomas: On additional point you may want to address in your introduction to the library is "Why/when would you want to use this remote write library". You do mention "How to use it" below.

lib/charms/prometheus_k8s/v0/prometheus_remote_write.py
```diff
@@ -0,0 +1,124 @@
...
+```
+from charms.prometheus_k8s.v0.prometheus_remote_write import PrometheusRemoteWriteConsumer
+
+def __init__(self, *args):
+    ...
+    self.remote_write = PrometheusRemoteWriteConsumer(self, "remote-write")
+    ...
+```
+
+Then access the endpoints with:
```
balbirthomas, line 18: It is not clear why a charm instantiating the `PrometheusRemoteWriteConsumer` would want to access the remote write endpoints or the config. I wonder if some clarification here perhaps with an example will be useful. For instance "use the config retuned by `config()` to set the ...."

lib/charms/prometheus_k8s/v0/prometheus_remote_write.py
```diff
@@ -0,0 +1,124 @@
...
+
+class PrometheusRemoteWriteConsumer(Object):
+    """A prometheus remote write consumer."""
+
+    def __init__(self, charm: CharmBase, relation_name: str):
+        """A prometheus remote write consumer.
+
+        Args:
+            charm: The charm object that instantiated this class.
+            relation_name: The relation name as defined in metadata.yaml.
```
balbirthomas, line 39: Shouldn't this be a bit more explicit on which relation name ? For instance "name of the relation for the promethneus_remote_write interface".

lib/charms/prometheus_k8s/v0/prometheus_remote_write.py
```diff
@@ -0,0 +1,124 @@
...
+                if (address := relation.data[unit].get("address")) is None:
+                    continue
+                if (port := relation.data[unit].get("port")) is None:
+                    continue
+                endpoints.append(f"http://{address}:{port}/api/v1/write")
+        return endpoints
+
+    @property
+    def configs(self) -> list:
+        """A config object ready to be dropped in to a prometheus config file.
```
balbirthomas, line 64: Is this config only used for "prometheus config file" ? Could it also be used for "grafana agent config file" ? Even though the two may be exactly the same I wonder if it should be made explicit which applications this config file is compatible with.

lib/charms/prometheus_k8s/v0/prometheus_remote_write.py
```diff
@@ -0,0 +1,124 @@
...
+                # If external-address is provided use that, else use ingress-address
+                if (address := relation.data[unit].get("address")) is None:
+                    continue
+                if (port := relation.data[unit].get("port")) is None:
+                    continue
+                endpoints.append(f"http://{address}:{port}/api/v1/write")
+        return endpoints
+
+    @property
+    def configs(self) -> list:
```
balbirthomas, line 63: A bit surprising that the function is called `config()` but it returns a list of endpoints. `config()` gives the idea that this method will return configuration file in YAML or JSON or some such format. Would it make sense to rename this as `endpoints()` or some such thing. It isn't returning a *complete configuration*, is it ?

lib/charms/prometheus_k8s/v0/prometheus_remote_write.py
```diff
@@ -0,0 +1,124 @@
...
+        """A config object ready to be dropped in to a prometheus config file.
+
+        Returns:
+            A list of remote_write configs.
+        """
+        return [{"url": endpoint} for endpoint in self.endpoints]
+
+
+class PrometheusRemoteWriteProvider(Object):
+    """A prometheus remote write provider."""
```
balbirthomas, line 73: The docstring is a bit cryptic. I wonder if we should clarify here that the Provider is a Prometheus Charm. I am not sure that is obvious. Can it be any other type of charm ?

lib/charms/prometheus_k8s/v0/prometheus_remote_write.py
```diff
@@ -0,0 +1,124 @@
...
+
+        Args:
+            charm: The charm object that instantiated this class.
+            relation_name: The relation name as defined in metadata.yaml.
+        """
+        super().__init__(charm, relation_name)
+        self._charm = charm
+        self._relation_name = relation_name
+
+    def set_endpoint(
```
balbirthomas, line 86: This is a much much better implementation. Kudos. I like the fact that it makes "simple things easy, and hard things possible". However seems like you are still expecting the `set_endpoint()` method to be explicitly invoked. Is there any reason not to prefer the alternative where the endpoint address is an optional constructor argument. If this constructor argument is not provided then this provider library object attempts to determine an end point address automatically, and failing to do so sets blocked status. This approach has the benefit that in the simple case, to use this provider object requires nothing more than instantiating it.

lib/charms/prometheus_k8s/v0/prometheus_remote_write.py
```diff
@@ -0,0 +1,124 @@
...
+        self,
+        port: Union[str, int],
+        address: Optional[str] = None,
+        ingress_relation: Optional[str] = None,
+        ingress_address: Optional[str] = None,
+    ) -> None:
+        """Set the address and port on which you will serve prometheus remote write.
+
+        If address is provided, all automatic logic will be skipped and the address provided will
+        be used. Otherwise the method will attempt to use the ingress information if it is
```
balbirthomas, line 96: I will be nice if you can add a little more detail regarding the logic here. But will leave that upto you. Consider how will a charm writer determine if he should leave it upto your library to set the endpoint or make his/her charm set it. The docstring needs to provide enough information to the charm writer to make this decision.

lib/charms/prometheus_k8s/v0/prometheus_remote_write.py
```diff
@@ -0,0 +1,124 @@
...
+        provided. Finally it will fall back to a predictable hostname if neither of the other two
+        methods worked.
+
+        Args:
+            port: The port number
+            address: The address of the remote write server
+            ingress_relation: The name of the ingress relation to use
+            ingress_address: The ingress address
```
balbirthomas, line 104: Why does `ingress_address` need to be a method argument when `ingress_relation` is already an optional one. For example if `ìngress_relation` is provided can it not be used to determine the `ingress_address` ?

lib/charms/prometheus_k8s/v0/prometheus_remote_write.py
```diff
@@ -0,0 +1,124 @@
...
+
+    def __init__(self, charm: CharmBase, relation_name: str):
+        """A prometheus remote write consumer.
+
+        Args:
+            charm: The charm object that instantiated this class.
+            relation_name: The relation name as defined in metadata.yaml.
+        """
+        super().__init__(charm, relation_name)
+        self._relation_name = relation_name
```
balbirthomas, line 42: I am a bit surprised that this charm library does not respond to any events. Is that because it only supports interaction with a single remote write endpoint ? What if no remote write endpoint is available but becomes available at a later point in time. In this case it it worth emitting a event to notify the consumer charm that a remote write endpoint has become available ? I suppose this question is related to the question in docstring above on how to use the `endpoints()` and `config()` method.

lib/charms/prometheus_k8s/v0/prometheus_remote_write.py
```diff
@@ -0,0 +1,124 @@
...
+            ):
+                if self._charm.app.planned_units() > 1:
+                    self._charm.unit.status = BlockedStatus(
+                        "Ingress does not support scale greater than 1"
+                    )
+                    return
+                address = ingress_address
+            else:
+                address = f"{self._charm.unit.name.replace('/','-')}.{self._charm.app.name}-endpoints.{self._charm.model.name}.svc.cluster.local"
+        for relation in self.model.relations[self._relation_name]:
```
balbirthomas, line 122: Aren't `address` and `port` too generic ? Would it be better to set `remote_write_address` and `remote_write_port` ? It may be a good idea to run this through @mmanciop .

lib/charms/prometheus_k8s/v0/prometheus_remote_write.py
```diff
@@ -0,0 +1,124 @@
...
+    def __init__(self, charm: CharmBase, relation_name: str):
+        """A prometheus remote write provider.
+
+        Args:
+            charm: The charm object that instantiated this class.
+            relation_name: The relation name as defined in metadata.yaml.
+        """
+        super().__init__(charm, relation_name)
+        self._charm = charm
+        self._relation_name = relation_name
```
balbirthomas, line 84: Again I am surprised that this library object does not handle any events ? What happens when a new remote write consumer charm unit joins ?

balbirthomas (approved):

lib/charms/prometheus_k8s/v0/prometheus_remote_write.py
dstathis: done

lib/charms/prometheus_k8s/v0/prometheus_remote_write.py
dstathis: In the grafana agent config file, you include a literal section of a prometheus config file. This is basically saying "can be put in a config file as defined by Prometheus"

lib/charms/prometheus_k8s/v0/prometheus_remote_write.py
dstathis: what about `endpoint_configs()` do you think that would be better?

lib/charms/prometheus_k8s/v0/prometheus_remote_write.py
dstathis: I am waiting to fix the library docs until it is decided what the library will look like. Changes will be made here for sure

lib/charms/prometheus_k8s/v0/prometheus_remote_write.py
dstathis: I'll do this once the code settles.

lib/charms/prometheus_k8s/v0/prometheus_remote_write.py
dstathis: The relation can't give you the address. For example, in Prometheus, how do I get `_external_hostname` from the relation object?

lib/charms/prometheus_k8s/v0/prometheus_remote_write.py
dstathis: Are you referring to the argument names. I wouldn't mind changing them if you want but do you not think it's rather obvious that address and port should be those of "remote_write"?

lib/charms/prometheus_k8s/v0/prometheus_remote_write.py
dstathis: Nothing happens when a new consumer unit joins. All the data it needs is already there. One endpoint per provider unit.

lib/charms/prometheus_k8s/v0/prometheus_remote_write.py
dstathis: So in the past we have written libraries to always be the object that listens to `relation_changed` even if we were not doing anything with it. This is something that was more or less dictated to us a long time ago. I think, however that it does not always make sense.

Specifically regarding this charm and the comment above. If new endpoints become available, a `relation_changed` event will trigger. One of two things can happen. First, the charm itself could listen to relation changed and do whatever it needs to do. Second, we could listen here, do nothing, and emit a different custom event for the charm to respond to. To me the former seems a lot more simple and transparent.

There are definitely cases when custom events are appropriate, I think, however, that in this case it would be no more than a rename of the already aptly named `relation_changed`.

lib/charms/prometheus_k8s/v0/prometheus_remote_write.py
dstathis: It could be for example Grafana Agent. In addition there are many other metrics solutions that support ingesting via remote write.

lib/charms/prometheus_k8s/v0/prometheus_remote_write.py
balbirthomas: You mean application relation data has already been set so nothing else needs to be done. But there can be multiple relations with the same name. Would relation data be set for all these relations ?

lib/charms/prometheus_k8s/v0/prometheus_remote_write.py
balbirthomas: OK no problem, perhaps not constructor then. Also relevant here is the discussions above about these library objects listening and responding to relation events. If these library objects do respond to event in this way, they can invoke the `set_endpoint()` method to set the endpoints in response to the appropriate event. If some library user would like to provide different endpoint behaviour they can always derive from you library object and override the `set_endpoint()` method. This way the using the library only requires instantiating it and yet it supports customisable behaviour (making "simple things easy and hard things possible").

lib/charms/prometheus_k8s/v0/prometheus_remote_write.py
balbirthomas: OK. Again the discussion here may be related to discussions above regarding library objects responding to events and setting endpoint transparently without the charm needing to invoke a method to set endpoints. If we adopt this approach then the arguments to this method may change. Something like "external hostname" can certainly be a constructor argument to this library object. There may be other options here for you to explore.

balbirthomas (approved):

mmanciop:
Works for me, tested with https://github.com/canonical/grafana-agent-operator/pull/1 . Merge and ship please? ;-)

balbirthomas:
> Works for me, tested with [canonical/grafana-agent-operator#1](https://github.com/canonical/grafana-agent-operator/pull/1) . Merge and ship please? ;-)

No please don't succumb to our PM's preasure. We need to protect him from himself ;-). @dstathis do think over the various concerns raised.

lib/charms/prometheus_k8s/v0/prometheus_remote_write.py
dstathis: That sounds fine I will change it.

lib/charms/prometheus_k8s/v0/prometheus_remote_write.py
dstathis: I don't understand. We are watching relation_created and add all the data then.

PR #131, opened 2021-09-24
Handle upgrade event and multiple consumers

### LINKED ISSUE #109: Alert rules not updated over juju refresh
When a new version of a charm embeds different alert rules than the previous one, the new alert rules seem not to be propagated to Prometheus. Probably we miss a hook in the Prometheus library for the provider charm update. This lack of update on charm update is also likely to be the case for scrape jobs by the way :-)

balbirthomas:
This commit refactors the logic of MetricsEndpointProvider to
do two addtional things

1. Ensure that scrape job information is update when the metrics
   endpoint provider charm is upgraded.
2. Ensure that all related metrics endpoint consumer charms are
   notified of any changes in scrape job information.

Closes #109

### DIFF
--- a/lib/charms/prometheus_k8s/v0/prometheus_scrape.py
+++ b/lib/charms/prometheus_k8s/v0/prometheus_scrape.py
@@ -789,12 +789,13 @@ class MetricsEndpointProvider(Object):
         self._jobs = [_sanitize_scrape_configuration(job) for job in jobs]
 
         events = self._charm.on[self._relation_name]
-        self.framework.observe(events.relation_joined, self._set_scrape_metadata)
-        self.framework.observe(events.relation_changed, self._set_scrape_metadata)
+        self.framework.observe(events.relation_joined, self._set_scrape_job_spec)
+        self.framework.observe(events.relation_changed, self._set_scrape_job_spec)
         self.framework.observe(self._service_event, self._set_unit_ip)
+        self.framework.observe(self._charm.on.upgrade_charm, self._set_scrape_job_spec)
 
-    def _set_scrape_metadata(self, event):
-        """Ensure scrape targets metadata is made available to Prometheus.
+    def _set_scrape_job_spec(self, event):
+        """Ensure scrape target information is made available to Prometheus.
 
         When a metrics provider charm is related to a Prometheus charm, the
         metrics provider sets metadata related to its own scrape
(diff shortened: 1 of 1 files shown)


lib/charms/prometheus_k8s/v0/prometheus_scrape.py
```diff
@@ -807,20 +808,19 @@ def _set_scrape_metadata(self, event):
...
+        for relation in self._charm.model.relations[self._relation_name]:
+            relation.data[self._charm.app]["scrape_metadata"] = json.dumps(self._scrape_metadata)
+            relation.data[self._charm.app]["scrape_jobs"] = json.dumps(self._scrape_jobs)
 
-        if alert_groups := self._labeled_alert_groups:
-            event.relation.data[self._charm.app]["alert_rules"] = json.dumps(
-                {"groups": alert_groups}
-            )
+            if alert_groups := self._labeled_alert_groups:
+                relation.data[self._charm.app]["alert_rules"] = json.dumps(
```
sed-i, line 821: Isn't app data shared across all relation's data?

lib/charms/prometheus_k8s/v0/prometheus_scrape.py
```diff
@@ -789,12 +789,13 @@ def __init__(
...
-        self.framework.observe(events.relation_joined, self._set_scrape_metadata)
-        self.framework.observe(events.relation_changed, self._set_scrape_metadata)
+        self.framework.observe(events.relation_joined, self._set_scrape_job_spec)
+        self.framework.observe(events.relation_changed, self._set_scrape_job_spec)
         self.framework.observe(self._service_event, self._set_unit_ip)
+        self.framework.observe(self._charm.on.upgrade_charm, self._set_scrape_job_spec)
 
-    def _set_scrape_metadata(self, event):
-        """Ensure scrape targets metadata is made available to Prometheus.
+    def _set_scrape_job_spec(self, event):
```
sed-i, line 797: Would this be eventually called if LeaderElectedEvent fires? Needed for scaledown.

sed-i (approved):

lib/charms/prometheus_k8s/v0/prometheus_scrape.py
balbirthomas: No I do not think so. As far as I understand there is an app data bucked per relation.

lib/charms/prometheus_k8s/v0/prometheus_scrape.py
balbirthomas: There is nothing that needs to be done in particular to handle a leader change. Whoever is the leader will automatically do the right thing.

lib/charms/prometheus_k8s/v0/prometheus_scrape.py
sed-i: This may not be the case in the following scenario which I observed multiple times with alertmanager:
1. Deploy number of units from the start, such that the leader is not unit/0.
2. Scale down a number of units under the current leader.
3. Units are removed but the new leader is elected only ~10 seconds after scaledown.

lib/charms/prometheus_k8s/v0/prometheus_scrape.py
sed-i: This may not be the case in the following scenario which I observed multiple times with alertmanager:
1. Deploy number of units from the start, such that the leader is not unit/0.
2. Scale down a number of units under the current leader.
3. Units are removed but the new leader is elected only ~10 seconds after scaledown.

balbirthomas: I still do not see why it should matter. Do elaborate if you see a particular issue ? Do we have a test case ?

sed-i: Looking more thoroughly at the code I see that scrape job spec carry no information about prom units (e.g. IP addresses). In that case it shouldn't matter.

PR #132, opened 2021-09-27
Check return value for configuration reload

### LINKED ISSUE #105: Prometheus charm does not notify on config-reload issues
When Prometheus has an invalid configuration due, for example, malformed alert rules, the charm does not go into `BlockedStatus`, but rather silently fails to reload configuration.

balbirthomas:
This commit ensures that blocked status is set on failure to
reload Prometheus configuration

Closes #105

### DIFF
--- a/src/charm.py
+++ b/src/charm.py
@@ -120,14 +120,17 @@ class PrometheusCharm(CharmBase):
         # Restart prometheus only if command line arguments have changed,
         # otherwise just reload its configuration.
         if current_services == new_layer.services:
-            self._prometheus_server.reload_configuration()
-            logger.info("Prometheus configuration reloaded")
+            reloaded = self._prometheus_server.reload_configuration()
+            if not reloaded:
+                self.unit.status = BlockedStatus("Failed to load Prometheus config")
+            else:
+                self.unit.status = ActiveStatus()
+                logger.info("Prometheus configuration reloaded")
         else:
             container.add_layer(self._name, new_layer, combine=True)
             container.restart(self._name)
             logger.info("Prometheus (re)started")
-
-        self.unit.status = ActiveStatus()
+            self.unit.status = ActiveStatus()
 
     def _set_alerts(self, container):
         """Create alert rule files for all Prometheus consumers.


mmanciop (approved):

mmanciop (approved):

mmanciop:
Did not QA it, but it seems like the `unit.status` should not be overwritten by other hooks

Abuelodelanada (approved):

src/charm.py
```diff
@@ -120,14 +120,17 @@ def _configure(self, _):
         # Restart prometheus only if command line arguments have changed,
         # otherwise just reload its configuration.
         if current_services == new_layer.services:
-            self._prometheus_server.reload_configuration()
-            logger.info("Prometheus configuration reloaded")
+            reloaded = self._prometheus_server.reload_configuration()
```
sed-i, line 123: walrus?

src/charm.py
```diff
@@ -120,14 +120,17 @@ def _configure(self, _):
...
+            if not reloaded:
+                self.unit.status = BlockedStatus("Failed to load Prometheus config")
```
sed-i, line 125: I would first attempt a service restart, followed by health check. Only then Blocked. For your consideration.

sed-i (approved):

src/charm.py
balbirthomas: The intention is not to restart service when Pebble layer has not changed. If Prometheus fails to reload service without Pebble layer change then almost invariably the cause is a bad config. Service restart will not solve this problem because it will also try to load a bad config.

src/charm.py
balbirthomas: Actually in this case even walrus is not necessary. On can just write `if not self._prometheus_server.reload_configuration()`. Is that easier to understand than `if not reloaded` ? Which of is clearer (more like natural language) to read ?

sed-i: Yes I see you do not use the value of `reloaded` inside. In that case I would swap the `if` to be positive:
```python
if self._prometheus_server.reload_configuration():
    # etc
else:
    self.unit.status = BlockedStatus("Failed to load Prometheus config")
```

src/charm.py
sed-i: My understanding is slightly different:

For example, if the workload app itself crashed, restarting the service here would be helpful. If the config is indeed invalid, then you'd get the same result anyway.
If you do not validate the config yourself, then you probably shouldn't make assumptions on validity; instead make best (blind) effort to get it running.

If OTOH you were validating the config yourself before passing it to the app, then yes, you should actively minimize downtime.

src/charm.py
balbirthomas: Even then is it nicer to read `if reloaded` or `if self._prometheus_server.reload_configuration()` ?

src/charm.py
balbirthomas: - Validating the config can be feature for a future pull request.
- Should the charm be monitoring if the workload application crashed and then restart the app ?

PR #135, opened 2021-10-01
Prometheus scrape lib streamline

mmanciop:
Remove a lot of setup overhead for the charm author that wants to use the `MetricsEndpointProvider` object:
    
* Avoid the need in most use-cases to specify the relation  with the `prometheus_scrape` the charm provides.
  In virtually all cases, there will be exactly one such relation, so we can identify it with introspection of the Charm `meta` information.

* Get rid of the `service_event` parameter: this argument has been introduced to work around a Juju quirk about not having a good way of getting the Pod IP address on K8s, which can be reliability be accessed when a workload container has a `pebble_ready` event.
  This commit removes the `service_event` entirely, and uses charm `meta` introspection to discover which workload containers are available, and register the update of the unit ip on the relation when any of the workload containers fires a `pebble_ready` event.
    
* Make `alert_rules_path` resilient against different repository structures: hardcoding `src/prometheus_alert_rules` as default for the `alert_rules_path` breaks the "magic" when the charm repository structure does not have the canonical structure, e.g., when the charm code needs to coexist with a Java+Maven/Gradle codebase (see the https://github.com/mmanciop/spring-music repository for one such example).
  Instead, we now look up by default the `prometheus_alert_rules` relative to the location of the Charm main file (usually `charm.py`).
    
Also, the doc of `MetricsEndpointProvider` has been overhauled to be more friendly to the charm authors.
    
TODO: Apply the same simplifications to `MetricsEndpointConsumer` and `MetricsEndpointAggregator` when the macro changes in this PR are agreed upon.

### DIFF
--- a/lib/charms/prometheus_k8s/v0/prometheus_scrape.py
+++ b/lib/charms/prometheus_k8s/v0/prometheus_scrape.py
@@ -296,10 +296,15 @@ relation data provide eponymous information.
 
 import json
 import logging
+import sys
+from enum import Enum
 from pathlib import Path
+from typing import Optional
 
 import yaml
+from ops.charm import CharmBase
 from ops.framework import EventBase, EventSource, Object, ObjectEvents
+from ops.model import ModelError
 
 # The unique Charmhub library identifier, never change it
 LIBID = "bc84295fef5f4049878f07b131968ee2"
@@ -309,7 +314,7 @@ LIBAPI = 0
 
 # Increment this PATCH version before using `charmcraft publish-lib` or reset
 # to 0 if you are raising the major API version
-LIBPATCH = 7
+LIBPATCH = 8
 
 
 logger = logging.getLogger(__name__)
@@ -333,6 +338,81 @@ DEFAULT_JOB = {
     "metrics_path": "/metrics",
     "static_configs": [{"targets": ["*:80"]}],
 }
+
+
+RELATION_INTERFACE_NAME = "prometheus_scrape"
+
+
+class RelationDirection(Enum):
+    """Relations can be either provided or required by a charm."""
+
+    PROVIDED = 1
+    REQUIRED = 2
+
+
+class NoRelationWithInterfaceFoundError(Exception):
+    """No relations with the given interface are found in the charm meta."""
+
+    def __init__(self, charm: CharmBase, relation_interface: str):
+        self.charm = charm
+        self.relation_interface = relation_interface
+        self.message = (
+            f"No relations with interface '{relation_interface}' found in the meta "
+            f"of the '{charm.meta.name}' charm"
+        )
+
+        super().__init__(self.message)
+
+
+class MultipleRelationsWithInterfaceFoundError(Exception):
+    """Multiple relations with the given interface are found in the charm meta."""
+
+    def __init__(self, charm: CharmBase, relation_interface: str, relations: list):
+        self.charm = charm
+        self.relation_interface = relation_interface
+        self.relations = relations
+        self.message = (
+            f"Multiple relations with interface '{relation_interface}' found in the meta "
+            f"of the '{charm.name}' charm"
+        )
+
+        super().__init__(self.message)
+
+
+def _get_single_relation_by_interface(
+    charm: CharmBase, relation_interface: str, relation_direction: RelationDirection
+) -> str:
+    """Retrive the only relation in the charm meta that uses the given interface.
+
+    Args:
+        charm: a `CharmBase` object to scan for the matching relation.
+        relation_interface: the string name of the relation interface to look up.
+            If `charm` has exactly one relation with this interface, the relation's
+            name is returned. If none or multiple relations with the provided interface
+            are found, this method will raise either an exception of type
+            NoRelationWithInterfaceFoundError or MultipleRelationsWithInterfaceFoundError,
+            respectively.
+        relation_direction: whether the relation to look up is either one of the provided
+            or required relations.
+    """
+    relations_with_right_direction = (
+        charm.meta.provides
+        if relation_direction == RelationDirection.PROVIDED
+        else charm.meta.requires
+    )
+
+    relations = [
+        relation_name
+        for relation_name, relation_meta in relations_with_right_direction.items()
+        if relation_meta.interface_name == relation_interface
+    ]
+
+    if len(relations) == 1:
+        return relations[0]
+    elif len(relations) == 0:
+        raise NoRelationWithInterfaceFoundError(charm, relation_interface)
+    else:
+        raise MultipleRelationsWithInterfaceFoundError(charm, relation_interface, relations)
 
 
 def _sanitize_scrape_configuration(job) -> dict:
@@ -387,19 +467,39 @@ class MetricsEndpointConsumer(Object):
 
     on = MonitoringEvents()
 
-    def __init__(self, charm, name):
+    def __init__(self, charm: CharmBase, relation_name: Optional[str] = None):
         """A Prometheus based Monitoring service provider.
 
         Args:
             charm: a `CharmBase` instance that manages this
                 instance of the Prometheus service.
-            name: string name of the relation over which scrape target
+            relation_name: string name of the relation over which scrape target
                 information is gathered by the Prometheus charm.
         """
-        super().__init__(charm, name)
+        if not relation_name:
+            try:
+                # Check if there is just one relation with the right interface
+                relation_name = _get_single_relation_by_interface(
+                    charm, RELATION_INTERFACE_NAME, RelationDirection.REQUIRED
+                )
+            except NoRelationWithInterfaceFoundError:
+                raise ModelError(
+                    f"No required relation with the '{RELATION_INTERFACE_NAME}' interface "
+                    f"found; did you add a relation with the '{RELATION_INTERFACE_NAME}' "
+                    "interface in the charm's metadata.yaml file?"
+                )
+            except MultipleRelationsWithInterfaceFoundError as e:
+                raise ModelError(
+                    f"Multiple required relations with the '{RELATION_INTERFACE_NAME}' "
+                    f"interface found: {e.relations}; you must specify which relation "
+                    f"should be managed by this {type(self)} by providing the "
+                    "`relation_name` argument."
+                )
+
+        super().__init__(charm, relation_name)
         self._charm = charm
-        self._relation_name = name
-        events = self._charm.on[name]
+        self._relation_name = relation_name
+        events = self._charm.on[relation_name]
         self.framework.observe(events.relation_changed, self._on_metrics_provider_relation_changed)
         self.framework.observe(
             events.relation_departed, self._on_metrics_provider_relation_departed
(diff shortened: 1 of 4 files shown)


lib/charms/prometheus_k8s/v0/prometheus_scrape.py
```diff
@@ -732,66 +733,157 @@ class MetricsEndpointProvider(Object):
...
+                `/metrics` endpoint pooling all units of this charm on port `80`.
             alert_rules_path: an optional path for the location of alert rules
-                files.  Defaults to "src/prometheus_alert_rules" at the top level
-                of the charm repository.
+                files.  Defaults to "./prometheus_alert_rules",
+                resolved from the directory hosting the charm entry file.
+                The alert rules are automatically update on charm upgrade.
         """
-        super().__init__(charm, name)
+        if not relation_name:
```
balbirthomas, line 835: This whole section of finding a relation can be extracted into a separate member function something like `find_relation()`. Naming sections of code by encapsulating them into functions makes the code easier to understand. Isn't that is software engineering 101 :-).

lib/charms/prometheus_k8s/v0/prometheus_scrape.py
```diff
@@ -732,66 +733,157 @@ class MetricsEndpointProvider(Object):
...
+                raise ModelError(
+                    "No provided relation with the 'prometheus_scrape' interface found"
+                )
+            else:
+                raise ModelError(
+                    "Multiple relations with the 'prometheus_scrape' interface found: "
+                    f" {prometheus_scrape_relations}"
+                )
+
+        if not alert_rules_path:
```
balbirthomas, line 855: Similarly extracting this logic of determining alert rules path into a separate function would be good practice.

lib/charms/prometheus_k8s/v0/prometheus_scrape.py
```diff
@@ -732,66 +733,157 @@ class MetricsEndpointProvider(Object):
...
 
         events = self._charm.on[self._relation_name]
-        self.framework.observe(events.relation_joined, self._set_scrape_job_spec)
+        self.framework.observe(events.relation_created, self._set_scrape_job_spec)
         self.framework.observe(events.relation_changed, self._set_scrape_job_spec)
-        self.framework.observe(self._service_event, self._set_unit_ip)
+
+        # Set the ip address when the containers start (as a workaround for
+        # not being able to lookup the pod ip)
+        for container_name in charm.unit.containers:
```
balbirthomas, line 881: At present there is only one container so this does not change behaviour. I wonder what the behaviour is like if more than one container is added in the future.

lib/charms/prometheus_k8s/v0/prometheus_scrape.py
```diff
@@ -732,66 +733,157 @@ class MetricsEndpointProvider(Object):
...
+                    "No provided relation with the 'prometheus_scrape' interface found"
+                )
+            else:
+                raise ModelError(
+                    "Multiple relations with the 'prometheus_scrape' interface found: "
+                    f" {prometheus_scrape_relations}"
+                )
+
+        if not alert_rules_path:
+            import sys
```
balbirthomas, line 856: There is no need to be importing modules in nested scope.

lib/charms/prometheus_k8s/v0/prometheus_scrape.py
```diff
@@ -732,66 +733,157 @@ class MetricsEndpointProvider(Object):
...
+
+            charm_file = sys.path[0]
+
+            import pathlib
```
balbirthomas, line 860: Same statement about importing modules into nested scope.

lib/charms/prometheus_k8s/v0/prometheus_scrape.py
```diff
@@ -732,66 +733,157 @@ class MetricsEndpointProvider(Object):
     def __init__(
         self,
         charm,
-        name,
-        service_event,
+        relation_name=None,
         jobs=[],
-        alert_rules_path="src/prometheus_alert_rules",
+        alert_rules_path=None,
```
balbirthomas, line 738: What is wrong with a default path ? This was your own suggestion. What prompted the change here ?

lib/charms/prometheus_k8s/v0/prometheus_scrape.py
```diff
@@ -732,66 +733,157 @@ class MetricsEndpointProvider(Object):
...
         self._ALERT_RULES_PATH = alert_rules_path
-        self._service_event = service_event
-        self._relation_name = name
+        self._relation_name = relation_name
         # Sanitize job configurations to the supported subset of parameters
         self._jobs = [_sanitize_scrape_configuration(job) for job in jobs]
 
         events = self._charm.on[self._relation_name]
-        self.framework.observe(events.relation_joined, self._set_scrape_job_spec)
+        self.framework.observe(events.relation_created, self._set_scrape_job_spec)
```
balbirthomas, line 876: Setting relation data on created is ok for application data. However quite some time ago I found this to work unreliably. But was a hard to reproduce problem so never filed a bug.  I wonder if this works more consistently now.

balbirthomas (changes requested):

balbirthomas:
In general this PR looks good. Just refactoring logic of fetching relation name and alert rules path into two separate functions will make the constructor easier to read.

lib/charms/prometheus_k8s/v0/prometheus_scrape.py
```diff
@@ -732,66 +733,157 @@ class MetricsEndpointProvider(Object):
...
+                `MetricsEndpointProvider` will look for the existence of precisely
+                one relation provided by the charm and using the `prometheus_scrape`
+                interface and, if none or multiple are found, it will raise a
+                ops.model.ModelError exception.
             jobs: an optional list of dictionaries where each
                 dictionary represents the Prometheus scrape
                 configuration for a single job. When not provided, a
                 default scrape configuration is provided for the
-                `/metrics` endpoint pooling using port `80`.
+                `/metrics` endpoint pooling all units of this charm on port `80`.
```
sed-i, line 829:
```suggestion
                `/metrics` endpoint polling all units of this charm on port `80`.
```
Also "all units of this charm" is unclear: which charm? The endpoint provider?

lib/charms/prometheus_k8s/v0/prometheus_scrape.py
```diff
@@ -732,66 +733,157 @@ class MetricsEndpointProvider(Object):
...
+                The alert rules are automatically update on charm upgrade.
         """
-        super().__init__(charm, name)
+        if not relation_name:
+            # Check if there is just one relation with the right interface
+            prometheus_scrape_relations = [
+                relation_name
+                for relation_name, relation_meta in charm.meta.provides.items()
+                if relation_meta.interface_name == "prometheus_scrape"
+            ]
```
sed-i, line 841: Consider using `filter` instead.

lib/charms/prometheus_k8s/v0/prometheus_scrape.py
```diff
@@ -732,66 +733,157 @@ class MetricsEndpointProvider(Object):
...
+
+            if len(prometheus_scrape_relations) == 1:
+                relation_name = prometheus_scrape_relations[0]
+            elif len(prometheus_scrape_relations) == 0:
+                raise ModelError(
+                    "No provided relation with the 'prometheus_scrape' interface found"
+                )
+            else:
+                raise ModelError(
+                    "Multiple relations with the 'prometheus_scrape' interface found: "
```
sed-i, line 851: add more actionable suggestion

lib/charms/prometheus_k8s/v0/prometheus_scrape.py
sed-i: I'd say a free function e.g. `get_relations_by_interface(charm, interface)`.

lib/charms/prometheus_k8s/v0/prometheus_scrape.py
```diff
@@ -732,66 +733,157 @@ class MetricsEndpointProvider(Object):
...
+                relation_name
+                for relation_name, relation_meta in charm.meta.provides.items()
+                if relation_meta.interface_name == "prometheus_scrape"
+            ]
+
+            if len(prometheus_scrape_relations) == 1:
+                relation_name = prometheus_scrape_relations[0]
+            elif len(prometheus_scrape_relations) == 0:
+                raise ModelError(
+                    "No provided relation with the 'prometheus_scrape' interface found"
```
sed-i, line 847: add more actionable suggestion

lib/charms/prometheus_k8s/v0/prometheus_scrape.py
sed-i: in fact doing so is a lint error with our current config

lib/charms/prometheus_k8s/v0/prometheus_scrape.py
sed-i: See `def get_this_script_dir() -> Path:`
https://github.com/canonical/lma-light-bundle/blob/c72b9a02adec54ac67c01642552af00d77831902/tests/integration/test_bundle.py#L35

lib/charms/prometheus_k8s/v0/prometheus_scrape.py
```diff
@@ -732,66 +733,157 @@ class MetricsEndpointProvider(Object):
...
+                )
+
+        if not alert_rules_path:
+            import sys
+
+            charm_file = sys.path[0]
+
+            import pathlib
+
+            default_alerts_dir = pathlib.Path(charm_file).joinpath("prometheus_alert_rules")
```
sed-i, line 862: I would take this opportunity to refactor the tree into
```
src
  alert_rules
    prometheus
    loki
```
etc

lib/charms/prometheus_k8s/v0/prometheus_scrape.py
sed-i: Agreed. From past experience, "joined" seems safer.

lib/charms/prometheus_k8s/v0/prometheus_scrape.py
sed-i: Add a comment acknowledging this is a "dirty fix"?

lib/charms/prometheus_k8s/v0/prometheus_scrape.py
```diff
@@ -803,10 +895,9 @@ def _set_scrape_job_spec(self, event):
         data.  In addition each of the consumer units also sets its own
         host address in Juju unit relation data.
 
-        Args:
-            event: a `CharmEvent` in response to which `MetricsEndpointProvider` will
-                forward scrape jobs, alert rules, metrics endpoint host addresses
-                and related metadata to the Prometheus charm.
+        All args are ignored. This is a commodity to enable this method to be
```
sed-i, line 898: Odd to say "all args are ignored" in an upstream function when they are ignored only downstream.
Can you instead ignore the event upstream instead of propagating it?

lib/charms/prometheus_k8s/v0/prometheus_scrape.py
```diff
@@ -732,66 +733,157 @@ class MetricsEndpointProvider(Object):
     def __init__(
```
sed-i, line 733: Consider adding a decorator to `__init__` that would modify the arguments as follows:
- relation would be added a default, if only one exists in metadata.yaml (you won't have access to `.meta` from the decorator)
- relation would **not** have a default value otherwise.

This way we earn static validation.

lib/charms/prometheus_k8s/v0/prometheus_scrape.py
mmanciop: Because I realized that that default does not work in repos with an unconventional structure. It's documented in the PR description and commit message.

lib/charms/prometheus_k8s/v0/prometheus_scrape.py
mmanciop: How about the "as a workaround" part? :-)

mmanciop: > At present there is only one container so this does not change behaviour. I wonder what the behaviour is like if more than one container is added in the future.

That we will update sometimes unnecessarily the address, which remains constant as it is constant across the entire pod lifecycle.

lib/charms/prometheus_k8s/v0/prometheus_scrape.py
mmanciop: But would anybody reading the code understand that?

lib/charms/prometheus_k8s/v0/prometheus_scrape.py
mmanciop: The reason why we went with a flat hierarchy of folders is that it is more discoverable when looking up an unknown codebase, e.g., in a Git-based web interface.

lib/charms/prometheus_k8s/v0/prometheus_scrape.py
```diff
@@ -732,66 +733,157 @@ class MetricsEndpointProvider(Object):
...
+            expr: up{%%juju_topology%%} < 1
+            for: 0m
+            labels:
+                severity: critical
+            annotations:
+              summary: Unit {{ $labels.juju_model }}/{{ $labels.juju_unit }} unavailable
+              description: >
+                The unit {{ $labels.juju_model }} {{ $labels.juju_unit }} is unavailable
+
+        The `%%juju_topology%%` will be replaced with label filters ensuring that the
```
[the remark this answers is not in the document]
mmanciop: Fixed the docs another way :-)

lib/charms/prometheus_k8s/v0/prometheus_scrape.py
```diff
@@ -732,66 +733,157 @@ class MetricsEndpointProvider(Object):
...
-        an error and does not load the particular rule.
+        In case of a charm exposing the metrics endpoint for each of its
+        units on port 8080 and the `/metrics` path, the
+        `MetricsEndpointProvider` can be instantiated as follows:
+
+            self.prometheus = MetricsEndpointProvider(
+                self,
+                jobs=[{
+                    "static_configs": [{"targets": ["*:8080"]}],
+                }])
```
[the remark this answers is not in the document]
mmanciop: Done

lib/charms/prometheus_k8s/v0/prometheus_scrape.py
```diff
@@ -732,66 +733,157 @@ class MetricsEndpointProvider(Object):
...
-        - `alert_rules`
-
-        The `alert_rules` are read from `*.rule` files in the `src/prometheus_alert_rules`
-        directory. If the syntax of these rules is invalid `MetricsEndpointProvider` logs
-        an error and does not load the particular rule.
+        In case of a charm exposing the metrics endpoint for each of its
+        units on port 8080 and the `/metrics` path, the
+        `MetricsEndpointProvider` can be instantiated as follows:
+
+            self.prometheus = MetricsEndpointProvider(
```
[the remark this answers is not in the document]
mmanciop: Done in both places

mmanciop:
Rebased the PR on top of #137, which fixes the build failure

lib/charms/prometheus_k8s/v0/prometheus_scrape.py
```diff
@@ -335,6 +339,81 @@ def _on_scrape_targets_changed(self, event):
...
+        self.relations = relations
+        self.message = (
+            f"Multiple relations with interface '{relation_interface}' found in the meta "
+            f"of the '{charm.name}' charm"
+        )
+
+        super().__init__(self.message)
+
+
+def _get_single_relation_by_interface(
```
balbirthomas, line 381: This method is only used by Provider. So what is the motivation to put this in Module scope ?

lib/charms/prometheus_k8s/v0/prometheus_scrape.py
```diff
@@ -822,15 +1020,15 @@ def _set_scrape_job_spec(self, event):
                     {"groups": alert_groups}
                 )
 
-    def _set_unit_ip(self, event):
+    def _set_unit_ip(self, _: EventBase):
```
balbirthomas, line 1023: Hmm. Rather silly to add type annotation for an unused argument. :-). Feels like religion.

lib/charms/prometheus_k8s/v0/prometheus_scrape.py
balbirthomas: I think we should resist the temptation to use programming constructs unless they are really warranted and the additional code actually simplifies something. Creating a decorator just for a single use from this viewpoint is not useful.

lib/charms/prometheus_k8s/v0/prometheus_scrape.py
balbirthomas: Most of this code looks good now so approving but I will mull about this a bit more.

balbirthomas (approved):

lib/charms/prometheus_k8s/v0/prometheus_scrape.py
mmanciop: I saw type hints everywhere, so went with "Why not"

lib/charms/prometheus_k8s/v0/prometheus_scrape.py
mmanciop: No also Consumer now

lib/charms/prometheus_k8s/v0/prometheus_scrape.py
```diff
@@ -387,19 +466,39 @@ class MetricsEndpointConsumer(Object):
...
         """
-        super().__init__(charm, name)
+        if not relation_name:
+            relation_interface = RELATION_INTERFACE_NAME
+            try:
+                # Check if there is just one relation with the right interface
+                relation_name = _get_single_relation_by_interface(
+                    charm, relation_interface, RelationDirection.REQUIRED
+                )
+            except NoRelationWithInterfaceFoundError:
```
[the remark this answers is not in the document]
mmanciop: Well, in the beginning this PR did not have custom exception types. We *could* propagate the custom exception types, I have nothing against it.

lib/charms/prometheus_k8s/v0/prometheus_scrape.py
```diff
@@ -335,6 +339,81 @@ def _on_scrape_targets_changed(self, event):
...
+        self.relation_interface = relation_interface
+        self.message = (
+            f"No relations with interface '{relation_interface}' found in the meta "
+            f"of the '{charm.meta.name}' charm"
+        )
+
+        super().__init__(self.message)
+
+
+class MultipleRelationsWithInterfaceFoundError(Exception):
```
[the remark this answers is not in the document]
mmanciop: No exceptionally good reason. I thought this would be more future-proof.

lib/charms/prometheus_k8s/v0/prometheus_scrape.py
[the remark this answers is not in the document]
mmanciop: It felt also wrong to put the hints on how to fix it in the exception constructor or the `_get_single_relation_by_interface`, so I ended up catching the custom `Exception` type anyhow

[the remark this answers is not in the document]
balbirthomas: I would have thought propagating a custom exceptions is "safer" since then the Prometheus charm can completely define the semantics of these exceptions. I am not sure if the context in which we are raising a `ModelError` aligns with its intended semantics or if we are running the risk to overload `ModelError` too much. Will ping @jnsgruk for some deeper insight and a second opinion here.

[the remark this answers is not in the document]
mmanciop: To me, this is fundamentally a "static" error (think of a compiler error, but for your charm), and not something the charm code should handle. My Python skills are not mighty enough to have an opinion about the best error/exception/curse to raise ;-)

lib/charms/prometheus_k8s/v0/prometheus_scrape.py
```diff
@@ -335,6 +340,81 @@ def _on_scrape_targets_changed(self, event):
...
 
 
+RELATION_INTERFACE_NAME = "prometheus_scrape"
+
+
+class RelationDirection(Enum):
+    """Relations can be either provided or required by a charm."""
+
+    PROVIDED = 1
+    REQUIRED = 2
```
sed-i, line 350: Reminds me of Java :)
I would replace `relation_direction: RelationDirection` with `relation_direction: typing.Literal["provides", "requires"]` so the enum wouldn't be needed.

lib/charms/prometheus_k8s/v0/prometheus_scrape.py
```diff
@@ -335,6 +340,81 @@ def _on_scrape_targets_changed(self, event):
...
+
+
+class NoRelationWithInterfaceFoundError(Exception):
```
sed-i, line 353: Compare to `class TooManyRelatedAppsError(ModelError)` in `ops.model`: inherits `ModelError`.

lib/charms/prometheus_k8s/v0/prometheus_scrape.py
sed-i: Consider renaming to `_get_relation_by_interface`: the `single` is redundant.

lib/charms/prometheus_k8s/v0/prometheus_scrape.py
```diff
@@ -387,19 +467,39 @@ class MetricsEndpointConsumer(Object):
 
     on = MonitoringEvents()
 
-    def __init__(self, charm, name):
+    def __init__(self, charm: CharmBase, relation_name: Optional[str] = None):
```
sed-i, line 470: iirc, `Optional[str] = None` is the same as `str = None` for th static inference

lib/charms/prometheus_k8s/v0/prometheus_scrape.py
```diff
@@ -727,72 +827,175 @@ def _labeled_unit_config(
...
+            try:
+                # Check if there is just one relation with the right interface
+                relation_name = _get_single_relation_by_interface(
+                    charm, RELATION_INTERFACE_NAME, RelationDirection.PROVIDED
+                )
+            except NoRelationWithInterfaceFoundError:
+                raise ModelError(
+                    f"No provided relation with the '{RELATION_INTERFACE_NAME}' interface found; "
+                    f"did you add a relation with the '{RELATION_INTERFACE_NAME}' interface in "
+                    "the charm's metadata.yaml file?"
```
sed-i, line 967:
```suggestion
                    f"The '{RELATION_INTERFACE_NAME}' interface does not exist; "
                    f"are there any relations with the '{RELATION_INTERFACE_NAME}' interface in "
                    "the charm's metadata.yaml file?"
```

lib/charms/prometheus_k8s/v0/prometheus_scrape.py
```diff
@@ -727,72 +827,175 @@ def _labeled_unit_config(
...
+                )
+            except MultipleRelationsWithInterfaceFoundError as e:
```
sed-i, line 969: Using exception messages for communicating to the developer information that ideally should have been statically inferred reminded me of "explicit is better than implicit": looking at all the changes that were necessary to have the relation an optional argument do not seem to be worth it.

lib/charms/prometheus_k8s/v0/prometheus_scrape.py
```diff
@@ -387,19 +490,26 @@ class MetricsEndpointConsumer(Object):
 
     on = MonitoringEvents()
 
-    def __init__(self, charm, name):
+    def __init__(self, charm: CharmBase, relation_name: str = DEFAULT_RELATION_NAME):
```
sed-i, line 493: I like this!

sed-i (approved):

lib/charms/prometheus_k8s/v0/prometheus_scrape.py
mmanciop: Turns out there is already `ops.charm.RelationRole` ;-)

lib/charms/prometheus_k8s/v0/prometheus_scrape.py
```diff
@@ -387,19 +490,26 @@ class MetricsEndpointConsumer(Object):
...
         Args:
             charm: a `CharmBase` instance that manages this
                 instance of the Prometheus service.
-            name: string name of the relation over which scrape target
-                information is gathered by the Prometheus charm.
+            relation_name: an optional string name of the relation between `charm`
+                and the Prometheus charmed service. The default is "metrics-endpoint".
+                It is strongly advised not to change the default, so that people
+                deploying your charm will have a consistent experience with all
+                other charms that consume metrics endpoints.
```
balbirthomas, line 503: The doc string should have a `Raises` section listing the names of exceptions that can be raised with a brief statement of the conditions under which they are raised.

lib/charms/prometheus_k8s/v0/prometheus_scrape.py
```diff
@@ -727,72 +837,154 @@ def _labeled_unit_config(
         return static_config
 
 
+def _resolve_dir_against_main_path(*path_elements: str) -> Optional[str]:
```
balbirthomas, line 840: Why does this have to be and `Optional[str]` return value. An empty string is false in any boolean context. So why not always return a string and in the failed case return and empty string. It is generally a good idea to return the same type from the same method.

lib/charms/prometheus_k8s/v0/prometheus_scrape.py
```diff
@@ -727,72 +837,154 @@ def _labeled_unit_config(
...
 
     def __init__(
         self,
         charm,
-        name,
-        service_event,
+        relation_name: str = DEFAULT_RELATION_NAME,
         jobs=[],
-        alert_rules_path="src/prometheus_alert_rules",
+        alert_rules_path: Optional[str] = None,
```
balbirthomas, line 866: Again there should be no need of `Optional[str]`. This can just be an empty string by default.

balbirthomas (approved):

lib/charms/prometheus_k8s/v0/prometheus_scrape.py
```diff
@@ -335,6 +338,106 @@ def _on_scrape_targets_changed(self, event):
...
+
+    relation: RelationMeta = charm.meta.relations[relation_name]
+
+    actual_relation_interface = relation.interface_name
+    if actual_relation_interface != expected_relation_interface:
+        raise RelationInterfaceMismatchError(
+            relation_name, expected_relation_interface, actual_relation_interface
+        )
+
+    if expected_relation_role == RelationRole.provides:
```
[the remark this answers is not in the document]
mmanciop: Not exactly: By checking first the name, then the role, we actually catch the case that the relation is created with the wrong role, rather than not at all. That can make a lot of difference for inexperienced charmers.

PR #145, opened 2021-10-14
Three fixes for build and test failures

balbirthomas:
This PR applies three fixes for build and test issues

1. A recent commit updated the k8s service patch library. This left over an API invocation for the old library which was no longer valid and has been removed.
2. The new k8s service patch library uses lightkube and lightkube-models that fails to build with charmcraft because of the lack of availabilty of source distributions on PyPi and an incomplete implementation of setup script for lightkube-models. Upstream has acknowledged the first issue and it has been fixed in their latest git main branch. The latter issue has been fixed in a forked repo. This PR points the Python requirements file to these two git repositories. With this changes the charm builds and relates to Prometheus tester.
3. This PR also removes a unit test for external labels that is no longer relevant since the charm configuration option for external labels have been removed in a previous commit.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -2,5 +2,5 @@ git+git://github.com/canonical/operator
 kubernetes
 pyaml
 requests
-git+git://github.com/gtsystem/lightkube
+lightkube >= 0.8.1
 git+git://github.com/balbirthomas/lightkube-models@pip-installer


requirements.txt
```diff
@@ -2,5 +2,5 @@ git+git://github.com/canonical/operator
 kubernetes
 pyaml
 requests
-lightkube
-lightkube-models
+git+git://github.com/gtsystem/lightkube
```
jnsgruk, line 5: This can just be 

```suggestion
lightkube >= 0.8.1
```

balbirthomas: Ok. Just waiting to check if integration tests pass. Will change.

PR #147, opened 2021-10-15
Prometheus remote_write library

mmanciop:
Implement a Prometheus `remote_write` library with all expected bells and whistles :-)

### DIFF
--- a/lib/charms/grafana_k8s/v0/grafana_source.py
+++ b/lib/charms/grafana_k8s/v0/grafana_source.py
@@ -33,7 +33,7 @@ LIBAPI = 0
 
 # Increment this PATCH version before using `charmcraft publish-lib` or reset
 # to 0 if you are raising the major API version
-LIBPATCH = 5
+LIBPATCH = 6
 
 logger = logging.getLogger(__name__)
 
@@ -268,6 +268,7 @@ class GrafanaSourceProvider(Object):
             "model": str(self._charm.model.name),
             "model_uuid": str(self._charm.model.uuid),
             "application": str(self._charm.model.app.name),
+            "type": self._source_type,
         }
         return data
 
@@ -435,6 +436,19 @@ class GrafanaSourceConsumer(Object):
     def _remove_source(self, source_name: str) -> None:
         """Remove a datasource by name."""
         self._stored.sources_to_delete.add(source_name)
+
+    def upgrade_keys(self) -> None:
+        """On upgrade, ensure stored data maintains compatibility."""
+        # self._stored.sources may have hyphens instead of underscores in key names.
+        # Make sure they reconcile.
+        sources = _type_convert_stored(self._stored.sources)
+        for rel_id in sources.keys():
+            for i in range(len(sources[rel_id])):
+                sources[rel_id][i].update(
+                    {k.replace("-", "_"): v for k, v in sources[rel_id][i].items()}
+                )
+
+        self._stored.sources = sources
 
     @property
     def sources(self) -> List[dict]:
--- a/lib/charms/grafana_k8s/v0/grafana_source.py
+++ b/lib/charms/grafana_k8s/v0/grafana_source.py
@@ -7,8 +7,22 @@ import json
 import logging
 from typing import Dict, List, Optional
 
-from ops.charm import CharmBase, CharmEvents, RelationDepartedEvent, RelationJoinedEvent
-from ops.framework import EventBase, EventSource, Object, ObjectEvents, StoredState
+from ops.charm import (
+    CharmBase,
+    CharmEvents,
+    RelationDepartedEvent,
+    RelationJoinedEvent,
+    RelationRole,
+)
+from ops.framework import (
+    EventBase,
+    EventSource,
+    Object,
+    ObjectEvents,
+    StoredDict,
+    StoredList,
+    StoredState,
+)
 from ops.model import Relation
 
 # The unique Charmhub library identifier, never change it
@@ -19,9 +33,122 @@ LIBAPI = 0
 
 # Increment this PATCH version before using `charmcraft publish-lib` or reset
 # to 0 if you are raising the major API version
-LIBPATCH = 1
+LIBPATCH = 5
 
 logger = logging.getLogger(__name__)
+
+DEFAULT_RELATION_NAME = "grafana-source"
+RELATION_INTERFACE_NAME = "grafana_datasource"
+
+
+def _type_convert_stored(obj):
+    """Convert Stored* to their appropriate types, recursively."""
+    if isinstance(obj, StoredList):
+        return list(map(_type_convert_stored, obj))
+    elif isinstance(obj, StoredDict):
+        rdict = {}
+        for k in obj.keys():
+            rdict[k] = _type_convert_stored(obj[k])
+        return rdict
+    else:
+        return obj
+
+
+class RelationNotFoundError(Exception):
+    """Raised if there is no relation with the given name."""
+
+    def __init__(self, relation_name: str):
+        self._relation_name = relation_name
+        self.message = "No relation named '{}' found".format(relation_name)
+
+        super().__init__(self.message)
+
+
+class RelationInterfaceMismatchError(Exception):
+    """Raised if the relation with the given name has a different interface."""
+
+    def __init__(
+        self,
+        relation_name: str,
+        expected_relation_interface: str,
+        actual_relation_interface: str,
+    ):
+        self._relation_name = relation_name
+        self.expected_relation_interface = expected_relation_interface
+        self.actual_relation_interface = actual_relation_interface
+        self.message = (
+            "The '{}' relation has '{}' as "
+            "interface rather than the expected '{}'".format(
+                relation_name, actual_relation_interface, expected_relation_interface
+            )
+        )
+
+        super().__init__(self.message)
+
+
+class RelationRoleMismatchError(Exception):
+    """Raised if the relation with the given name has a different direction."""
+
+    def __init__(
+        self,
+        relation_name: str,
+        expected_relation_role: RelationRole,
+        actual_relation_role: RelationRole,
+    ):
+        self._relation_name = relation_name
+        self.expected_relation_interface = expected_relation_role
+        self.actual_relation_role = actual_relation_role
+        self.message = "The '{}' relation has role '{}' rather than the expected '{}'".format(
+            relation_name, repr(actual_relation_role), repr(expected_relation_role)
+        )
+
+        super().__init__(self.message)
+
+
+def _validate_relation_by_interface_and_direction(
+    charm: CharmBase,
+    relation_name: str,
+    expected_relation_interface: str,
+    expected_relation_role: RelationRole,
+) -> None:
+    """Verifies that a relation has the necessary characteristics.
+
+    Verifies that the `relation_name` provided: (1) exists in metadata.yaml,
+    (2) declares as interface the interface name passed as `relation_interface`
+    and (3) has the right "direction", i.e., it is a relation that `charm`
+    provides or requires.
+
+    Args:
+        charm: a `CharmBase` object to scan for the matching relation.
+        relation_name: the name of the relation to be verified.
+        expected_relation_interface: the interface name to be matched by the
+            relation named `relation_name`.
+        expected_relation_role: whether the `relation_name` must be either
+            provided or required by `charm`.
+    """
+    if relation_name not in charm.meta.relations:
+        raise RelationNotFoundError(relation_name)
+
+    relation = charm.meta.relations[relation_name]
+
+    actual_relation_interface = relation.interface_name
+    if actual_relation_interface != expected_relation_interface:
+        raise RelationInterfaceMismatchError(
+            relation_name, expected_relation_interface, actual_relation_interface
+        )
+
+    if expected_relation_role == RelationRole.provides:
+        if relation_name not in charm.meta.provides:
+            raise RelationRoleMismatchError(
+                relation_name, RelationRole.provides, RelationRole.requires
+            )
+    elif expected_relation_role == RelationRole.requires:
+        if relation_name not in charm.meta.requires:
+            raise RelationRoleMismatchError(
+                relation_name, RelationRole.requires, RelationRole.provides
+            )
+    else:
+        raise Exception("Unexpected RelationDirection: {}".format(expected_relation_role))
 
 
 class SourceFieldsMissingError(Exception):
@@ -58,32 +185,30 @@ class GrafanaSourceEvents(ObjectEvents):
     sources_to_delete_changed = EventSource(GrafanaSourcesChanged)
 
 
-class GrafanaSourceConsumer(Object):
-    """A consumer object for Grafana datasources."""
+class GrafanaSourceProvider(Object):
+    """A provider object for Grafana datasources."""
 
     _stored = StoredState()
 
     def __init__(
         self,
         charm: CharmBase,
-        name: str,
         refresh_event: CharmEvents,
+        relation_name: str = DEFAULT_RELATION_NAME,
         source_type: Optional[str] = "prometheus",
         source_port: Optional[str] = "9090",
     ) -> None:
         """Construct a Grafana charm client.
 
-        The :class:`GrafanaSourceConsumer` object provides an interface
+        The :class:`GrafanaSourceProvider` object provides an interface
         to Grafana. This interface supports providing additional
         sources for Grafana to monitor. For example, if a charm
         exposes some metrics which are consumable by an ingestor
         (such as Prometheus), then an additional source can be added
-        by instantiating a :class:`GrafanaSourceConsumer` object and
+        by instantiating a :class:`GrafanaSourceProvider` object and
         adding its datasources as follows:
 
-            self.grafana = GrafanaSourceConsumer(
-                self, "grafana-datasource", {"grafana_datasource"}: ">=2.0"}
-            )
+            self.grafana = GrafanaSourceProvider(self)
             self.grafana.add_source(
                 address=<address>,
                 port=<port>
@@ -91,38 +216,46 @@ class GrafanaSourceConsumer(Object):
 
         Args:
             charm: a :class:`CharmBase` object which manages this
-                :class:`GrafanaSourceConsumer` object. Generally this is
+                :class:`GrafanaSourceProvider` object. Generally this is
                 `self` in the instantiating class.
-            name: a :string: name of the relation between `charm`
-                the Grafana charmed service.
+            relation_name: string name of the relation that is provides the
+                Grafana source service. It is strongly advised not to change
+                the default, so that people deploying your charm will have a
+                consistent experience with all other charms that provide
+                Grafana datasources.
             refresh_event: a :class:`CharmEvents` event on which the IP
                 address should be refreshed in case of pod or
                 machine/VM restart.
             source_type: an optional (default `prometheus`) source type
-                required for Grafana configuration
+                required for Grafana configuration. The value must match
+                the DataSource type from the Grafana perspective.
             source_port: an optional (default `9090`) source port
-                required for Grafana configuration
+                required for Grafana configuration.
         """
-        super().__init__(charm, name)
-        self.charm = charm
-        self.name = name
-        events = self.charm.on[name]
+        _validate_relation_by_interface_and_direction(
+            charm, relation_name, RELATION_INTERFACE_NAME, RelationRole.provides
+        )
+
+        super().__init__(charm, relation_name)
+        self._charm = charm
+        self._relation_name = relation_name
+        events = self._charm.on[relation_name]
 
         self._source_type = source_type
         self._source_port = source_port
 
         self.framework.observe(events.relation_joined, self._set_sources)
-        self.framework.observe(refresh_event, self._set_unit_ip)
+        self.framework.observe(refresh_event, self._set_unit_ip)  # type: ignore[arg-type]
 
     def _set_sources(self, event: RelationJoinedEvent):
-        """Inform the provider about the source configuration."""
-        self._set_unit_ip(event)
+        """Inform the consumer about the source configuration."""
+        self._set_unit_ip(event)  # type: ignore[arg-type]
 
-        if not self.charm.unit.is_leader():
+        if not self._charm.unit.is_leader():
             return
 
         logger.debug("Setting Grafana data sources: %s", self._scrape_data)
-        event.relation.data[self.charm.app]["grafana_source_data"] = json.dumps(self._scrape_data)
+        event.relation.data[self._charm.app]["grafana_source_data"] = json.dumps(self._scrape_data)
 
     @property
     def _scrape_data(self) -> Dict:
@@ -132,45 +265,51 @@ class GrafanaSourceConsumer(Object):
             Source configuration data for Grafana.
         """
         data = {
-            "model": str(self.charm.model.name),
-            "model_uuid": str(self.charm.model.uuid),
-            "application": str(self.charm.model.app.name),
-            "type": self._source_type,
+            "model": str(self._charm.model.name),
+            "model_uuid": str(self._charm.model.uuid),
+            "application": str(self._charm.model.app.name),
         }
         return data
 
     def _set_unit_ip(self, event: CharmEvents):
         """Set unit host address.
 
-        Each time a consumer charm container is restarted it updates its own host address in the
-        unit relation data for the Prometheus provider.
+        Each time a provider charm container is restarted it updates its own host address in the
+        unit relation data for the Prometheus consumer.
         """
-        for relation in self.charm.model.relations[self.name]:
-            relation.data[self.charm.unit]["grafana_source_host"] = "{}:{}".format(
-                str(self.charm.model.get_binding(relation).network.bind_address),
+        for relation in self._charm.model.relations[self._relation_name]:
+            relation.data[self._charm.unit]["grafana_source_host"] = "{}:{}".format(
+                str(self._charm.model.get_binding(relation).network.bind_address),
                 self._source_port,
             )
 
 
-class GrafanaSourceProvider(Object):
-    """A provider object for working with Grafana datasources."""
+class GrafanaSourceConsumer(Object):
+    """A consumer object for working with Grafana datasources."""
 
     on = GrafanaSourceEvents()
     _stored = StoredState()
 
-    def __init__(self, charm: CharmBase, name: str) -> None:
-        """A Grafana based Monitoring service consumer.
+    def __init__(self, charm: CharmBase, relation_name: str = DEFAULT_RELATION_NAME) -> None:
+        """A Grafana based Monitoring service consumer, i.e., the charm that uses a datasource.
 
         Args:
             charm: a :class:`CharmBase` instance that manages this
                 instance of the Grafana source service.
-            name: string name of the relation that is provides the
-                Grafana source service.
+            relation_name: string name of the relation that is provides the
+                Grafana source service. It is strongly advised not to change
+                the default, so that people deploying your charm will have a
+                consistent experience with all other charms that provide
+                Grafana datasources.
         """
-        super().__init__(charm, name)
-        self.name = name
-        self.charm = charm
-        events = self.charm.on[name]
+        _validate_relation_by_interface_and_direction(
+            charm, relation_name, RELATION_INTERFACE_NAME, RelationRole.requires
+        )
+
+        super().__init__(charm, relation_name)
+        self._relation_name = relation_name
+        self._charm = charm
+        events = self._charm.on[relation_name]
 
         self._stored.set_default(
             sources=dict(),
@@ -181,10 +320,10 @@ class GrafanaSourceProvider(Object):
         self.framework.observe(events.relation_departed, self._on_grafana_source_relation_departed)
 
     def _on_grafana_source_relation_changed(self, event: CharmEvents) -> None:
-        """Handle relation changes in related consumers.
+        """Handle relation changes in related providers.
 
-        If there are changes in relations between Grafana source providers
-        and consumers, this event handler (if the unit is the leader) will
+        If there are changes in relations between Grafana source consumers
+        and providers, this event handler (if the unit is the leader) will
         get data for an incoming grafana-source relation through a
         :class:`GrafanaSourcesChanged` event, and make the relation data
         is available in the app's datastore object. This data is set using
@@ -193,12 +332,12 @@ class GrafanaSourceProvider(Object):
         The Grafana charm can then respond to the event to update its
         configuration.
         """
-        if not self.charm.unit.is_leader():
+        if not self._charm.unit.is_leader():
             return
 
         sources = {}
 
-        for rel in self.charm.model.relations[self.name]:
+        for rel in self._charm.model.relations[self._relation_name]:
             source = self._get_source_config(rel)
             if source:
                 sources[rel.id] = source
@@ -208,7 +347,7 @@ class GrafanaSourceProvider(Object):
         self.on.sources_changed.emit()
 
     def _get_source_config(self, rel: Relation):
-        """Generate configuration from data stored in relation data by consumers."""
+        """Generate configuration from data stored in relation data by providers."""
         source_data = json.loads(rel.data[rel.app].get("grafana_source_data", "{}"))
         if not source_data:
             return
@@ -225,19 +364,19 @@ class GrafanaSourceProvider(Object):
 
             host_data = {
                 "unit": unit_name,
-                "source-name": unique_source_name,
-                "source-type": source_data["type"],
+                "source_name": unique_source_name,
+                "source_type": source_data["type"],
                 "url": "http://{}".format(host_addr),
             }
 
-            if host_data["source-name"] in self._stored.sources_to_delete:
-                self._stored.sources_to_delete.remove(host_data["source-name"])
+            if host_data["source_name"] in self._stored.sources_to_delete:
+                self._stored.sources_to_delete.remove(host_data["source_name"])
 
             data.append(host_data)
         return data
 
     def _relation_hosts(self, rel: Relation) -> Dict:
-        """Fetch host names and address of all consumer units for a single relation.
+        """Fetch host names and address of all provider units for a single relation.
 
         Args:
             rel: An `ops.model.Relation` object for which the host name to
@@ -256,14 +395,14 @@ class GrafanaSourceProvider(Object):
         return hosts
 
     def _on_grafana_source_relation_departed(self, event: RelationDepartedEvent) -> None:
-        """Update job config when consumers depart.
+        """Update job config when providers depart.
 
-        When a Grafana source consumer departs, the configuration
-        for that consumer is removed from the list of sources jobs,
-        added to a list of sources to remove, and other consumers
+        When a Grafana source provider departs, the configuration
+        for that provider is removed from the list of sources jobs,
+        added to a list of sources to remove, and other providers
         are informed through a :class:`GrafanaSourcesChanged` event.
         """
-        if not self.charm.unit.is_leader():
+        if not self._charm.unit.is_leader():
             return
 
         self._remove_source_from_datastore(event)
@@ -281,7 +420,7 @@ class GrafanaSourceProvider(Object):
             if event.unit:
                 # Remove one unit only
                 dead_unit = [s for s in removed_source if s["unit"] == event.unit.name][0]
-                self._remove_source(dead_unit["source-name"])
+                self._remove_source(dead_unit["source_name"])
 
                 # Re-update the list of stored sources
                 self._stored.sources[rel_id] = [
@@ -289,7 +428,7 @@ class GrafanaSourceProvider(Object):
                 ]
             else:
                 for host in removed_source:
-                    self._remove_source(host["source-name"])
+                    self._remove_source(host["source_name"])
 
             self.on.sources_to_delete_changed.emit()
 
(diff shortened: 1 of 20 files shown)


mmanciop:
Actually, we should build the support for alert rules like for the `prometheus_scrape` library, brb!

tests/unit/test_remote_write.py
```diff
@@ -0,0 +1,263 @@
...
+        self.assertEqual(
+            self.harness.get_relation_data(cons_rel_id, self.harness.charm.unit.name),
+            {"remote_write": json.dumps({"url": "http://my_happy_ingress:9090/api/v1/write"})},
+        )
+
+        self.assertEqual(
+            self.harness.charm.unit.status,
+            BlockedStatus(
+                "invalid combination of 'ingress', 'receive-remote-write' relations and multiple units"
+            ),
```
simskij, line 228: This test is very fragile. Make the error message a constant in the implementation and assert against that in the test to prevent it from breaking if we decide to change the phrasing.

src/charm.py
```diff
@@ -112,15 +127,43 @@ def _configure(self, _):
...
                 self.unit.status = ActiveStatus()
                 logger.info("Prometheus configuration reloaded")
         else:
             container.add_layer(self._name, new_layer, combine=True)
             container.restart(self._name)
             logger.info("Prometheus (re)started")
+
+        remote_write_plus_ingress_status_message = (
+            "invalid combination of 'ingress', 'receive-remote-write' relations and multiple units"
+        )
```
simskij, line 141: extract this into a constant.

simskij (changes requested):
Minor feedback

src/charm.py
mmanciop: Solved in 9dbaf8e

tests/unit/test_remote_write.py
mmanciop: Solved in 9dbaf8e

simskij (approved):

mmanciop:
Notice that the integration test works locally when modifying its logic to build the `grafana-agent-k8s` charm, and will fail in builds until we publish an `edge` of the `grafana-agent-k8s` with the `remote_write` support.

simskij:
If I disable the other test, the integration test passes (see https://github.com/canonical/prometheus-operator/runs/4263232727?check_suite_focus=true). Will dig a bit deeper into this prior to considering myself "done".

simskij:
@balbirthomas  and @sed-i, i'd like your critical reviews of this prior to merging. i've run the integration tests about 20 times locally to try to find any flakiness, and they seem stable. there are two major differences compared to @mmanciop's original PR:

- I've cleaned up the test_charm integration tests to not test the same thing multiple times. Both of the removed test cases are covered by the remaining test, so to keep testing time shorter, they're not tested in isolation.
- I've introduced a lot of usage of `asyncio.gather` to do suitable operations concurrently, which also speeds up the testing a lot.

One thing that I'd like to explore more going forward is to build concurrently as well, but unfortunately there seems to be stability issues in either the python-libjuju api or in charmcraft that prevents us from adopting that at the moment.

src/charm.py
```diff
@@ -112,15 +131,41 @@ def _configure(self, _):
...
+            self.unit.status = BlockedStatus(remote_write_plus_ingress_status_message)
+            logger.error(
+                "Using the ingress relation and receiving remote_write relations with more than "
+                "one Prometheus unit; remote_write data is likely to be sent to only one unit."
+            )
+        elif (
+            isinstance(self.unit.status, BlockedStatus)
+            and self.unit.status.message == remote_write_plus_ingress_status_message
+        ):
+            self.unit.status = ActiveStatus()
```
sed-i, line 155: Afaiu the pattern in _configure is to return immediately after status is set to something other than active, and only at the bottom of the function set the status to active. Not sure if this status check with string comparison is needed.

src/charm.py
```diff
@@ -112,15 +131,41 @@ def _configure(self, _):
...
+        self.remote_write_provider.update_endpoint()
+
+        if not isinstance(self.unit.status, BlockedStatus):
             self.unit.status = ActiveStatus()
 
+    def _validate_ingress_and_remote_write(self) -> bool:
+        if not (peer_relation := self.model.get_relation("prometheus-peers")):
+            return True
+
+        return not (self.model.get_relation("ingress") and len(peer_relation.units) > 0)
```
sed-i, line 167: Seems like this check could be handy in other charms.
That being said, iirc, this charm implicitly assumes in several places there is only one unit.

tests/integration/helpers.py
```diff
@@ -144,3 +162,19 @@ def oci_image(metadata_file: str, image_name: str) -> str:
...
+        ops_test: pytest-operator plugin
+        app_names: array of application names to check for
+
+    Returns:
+        whether the workloads are active or not
+    """
+    for name in app_names:
+        if ops_test.model.applications[name].units[0].workload_status != "active":
+            return False
+    return True
```
sed-i, line 180:
```suggestion
    return all(ops_test.model.applications[name].units[0].workload_status == "active" for name in app_names)
```

tests/integration/test_charm.py
```diff
@@ -71,30 +33,37 @@ async def test_prometheus_scrape_relation_with_prometheus_tester(
...
+    await asyncio.gather(
+        ops_test.model.deploy(
+            prometheus_charm, resources=prometheus_resources, application_name=prometheus_app_name
+        ),
+        ops_test.model.deploy(
+            prometheus_tester_charm, resources=tester_resources, application_name=tester_app_name
+        ),
     )
+
+    await ops_test.model.wait_for_idle(apps=app_names, status="active")
```
sed-i, line 47: Could be moved into the gather?

tests/integration/test_charm.py
```diff
@@ -71,30 +33,37 @@ async def test_prometheus_scrape_relation_with_prometheus_tester(
...
+
     await ops_test.model.block_until(
-        lambda: len(ops_test.model.applications[tester_app_name].units) > 0
+        lambda: (
+            len(ops_test.model.applications[prometheus_app_name].units) > 0
+            and len(ops_test.model.applications[tester_app_name].units) > 0
+        )
     )
```
sed-i, line 54: Consider adding a comment to remove the `block_until` when https://github.com/juju/python-libjuju/issues/574 is resolved.

tests/integration/test_remote_write.py
```diff
@@ -0,0 +1,59 @@
...
+    await ops_test.model.deploy(
+        prometheus_charm,
+        resources={"prometheus-image": oci_image("./metadata.yaml", "prometheus-image")},
+        application_name=prometheus_name,
+    )
+    await ops_test.model.deploy(
+        "grafana-agent-k8s",
+        application_name=agent_name,
+        channel="edge",
+    )
```
sed-i, line 34: Gather the two deploys?

tests/integration/test_charm.py
```diff
@@ -71,30 +33,37 @@ async def test_prometheus_scrape_relation_with_prometheus_tester(
...
 
-    assert ops_test.model.applications[prometheus_app_name].units[0].workload_status == "active"
-    assert ops_test.model.applications[tester_app_name].units[0].workload_status == "active"
-
+    assert initial_workload_is_ready(ops_test, app_names)
```
sed-i, line 56: This seems to be redundant because `wait_for_idle` is doing exactly that.

tests/integration/test_remote_write.py
```diff
@@ -0,0 +1,59 @@
...
+
+    await ops_test.model.add_relation(prometheus_name, agent_name)
+    apps = [prometheus_name, agent_name]
+    await ops_test.model.wait_for_idle(apps=apps, status="active")
+    assert initial_workload_is_ready(ops_test, apps)
```
sed-i, line 39: This seems to be redundant because `wait_for_idle` is doing exactly that.

tests/integration/test_remote_write.py
```diff
@@ -0,0 +1,59 @@
...
+    await check_prometheus_is_ready(ops_test, prometheus_name, 0)
+
+    await has_metric(
```
sed-i, line 42: Nice!

sed-i (commented):
Seems like this branch, when rebased, has reverted some of the latest changes.
Specifically the 3.5 stuff, but maybe more.
Also, there are quite a few potential merge conflicts with #161. Since this duplicates some alert rules logic I'd prefer to first merge #161.

Could you please add to the description of the PR:
- How remote write is related to MetricsProvider
- How alert rules are dealt with - is it the same as in MetricsProvider?

mmanciop:
> * How remote write is related to MetricsProvider

Not at all. Remote write and scrape are entirely different ways of getting data into Prometheus.

> * How alert rules are dealt with - is it the same as in MetricsProvider?

Yes same

lib/charms/prometheus_k8s/v0/prometheus_remote_write.py
```diff
@@ -0,0 +1,701 @@
...
+            interface: prometheus_remote_write  # Relation interface
+    ```
+
+    The charmed operator uses the `PrometheusRemoteWriteConsumer` as follows:
+
+    ```
+    def __init__(self, *args):
+        ...
+        self.remote_write_consumer = PrometheusRemoteWriteConsumer(self)
+        ...
```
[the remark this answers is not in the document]
simskij: They document different parts of the usage, so yes.

lib/charms/prometheus_k8s/v0/prometheus_remote_write.py
```diff
@@ -0,0 +1,701 @@
...
+    the relation name, `prometheus-remote-write`, but that requires you to provide the new
+    relation name to the `PrometheusRemoteWriteConsumer` via the `relation_name` constructor
+    argument. (The relation interface, on the other hand, is immutable and, if you were to change
+    it, your charm would not be able to relate with other charms using the right relation
+    interface. The library prevents you from doing that by raising an exception.) In any case, it
+    is strongly discouraged to change the relation name: having consistent relation names across
+    charms that do similar things is a very good thing for the people that will use your charm.
+    The one exception to the rule above, is if you charm needs to both consume and provide a
+    relation using the `prometheus_remote_write` interface, in which case changing the relation
+    name to differentiate between "incoming" and "outgoing" remote write interactions is necessary.
```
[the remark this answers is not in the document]
simskij: Close to where it might become an issue, so until we have a better place for this kind of docu I'd say this is completely fine.

src/charm.py
```diff
@@ -24,6 +27,10 @@
 PROMETHEUS_CONFIG = "/etc/prometheus/prometheus.yml"
 RULES_DIR = "/etc/prometheus/rules"
 
+INGRESS_MULTIPLE_UNITS_STATUS_MESSAGE = (
```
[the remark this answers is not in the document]
simskij: in general, it's a good practice to keep static strings in exported consts if they're used as part of an exception or exception-like behavior. Checking against `Blocked` might be sufficient, but I'd still consider an even narrower comparison preferable.

tests/integration/test_charm.py
simskij: That would make it less clear imo, maybe even flaky, given that the `wait_for_idle` call then will (likely) be executed prior to there being any app entries to await.

tests/integration/test_charm.py
simskij: This is not completely true, as `wait_for_idle` also resolves for unitless apps that stay idle for long enough.

lib/charms/prometheus_k8s/v0/prometheus_remote_write.py
```diff
@@ -0,0 +1,701 @@
...
+LIBPATCH = 1
+
+
+logger = logging.getLogger(__name__)
+
+
+DEFAULT_RELATION_NAME = "prometheus-remote-write"
+RELATION_INTERFACE_NAME = "prometheus_remote_write"
+
+DEFAULT_ALERT_RULES_RELATIVE_PATH = "./src/prometheus_alert_rules"
```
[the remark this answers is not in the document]
simskij: I'm not sure I get the question. We send alert rules over remote_write as well, just as when scraping.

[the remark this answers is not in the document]
sed-i: The question originated from misunderstanding remote_write and alert rule. Resolving.

lib/charms/prometheus_k8s/v0/prometheus_remote_write.py
```diff
@@ -0,0 +1,688 @@
...
+    def _set_endpoint_on_relation(self, relation: Relation) -> None:
+        """Set the the remote_write endpoint on relations.
+
+        Args:
+            relation: Optional relation. If provided, only this relation will be
+                updated. Otherwise, all instances of the `prometheus_remote_write`
+                relation managed by this `PrometheusRemoteWriteProvider` will be
+                updated.
+        """
+        address = self._endpoint_address or self._get_relation_bind_address()
```
balbirthomas, line 618: I wonder if there needs to be a check for having obtained a valid address here. I mean are we guaranteed that a bind address will always be available for a relation at any state of the lifecycle of the charm.

tests/unit/test_remote_write.py
```diff
@@ -0,0 +1,261 @@
...
+        )
+
+        self.assertEqual(
+            self.harness.charm.unit.status,
+            BlockedStatus(INGRESS_MULTIPLE_UNITS_STATUS_MESSAGE),
+        )
+
+        self.harness.remove_relation_unit(peers_rel_id, "prometheus/1")
+
+        # TODO Uncomment when https://github.com/canonical/operator/issues/638 is fixed
```
balbirthomas, line 231: This can be uncommented now since the issue seems to have been fixed.

balbirthomas (approved):
Will be running integration tests a couple of times on my system but looks good so approving.

tests/integration/test_charm.py
```diff
@@ -106,11 +80,9 @@ async def test_prometheus_scrape_relation_with_prometheus_tester(
...
+        get_prometheus_config(ops_test, prometheus_app_name, 0),
+        get_prometheus_rules(ops_test, prometheus_app_name, 0),
+    )
     assert initial_config == relation_removed_config
-
-    relation_removed_rules = await get_prometheus_rules(ops_test, prometheus_app_name, 0)
     assert initial_rules == relation_removed_rules
-
-    await ops_test.model.applications[prometheus_app_name].remove()
-    await ops_test.model.reset()
```
balbirthomas, line 116: I am see errors when one integration test run finishes will all tests passing then I run another immediately (log attached). I noticed in such cases the old model and application have not been completely removed and are still in the terminating stage. I wonder if it would be better to retain waits on removal of model and application.
[error-1.log](https://github.com/canonical/prometheus-operator/files/7595490/error-1.log)

simskij: Yes, in that case, it makes sense. I'm a little bit surprised by that though as a new model is created for each test run. I'll have a look at adding the cleanup back.

lib/charms/prometheus_k8s/v0/prometheus_remote_write.py
```diff
@@ -0,0 +1,701 @@
...
+        self,
+        relation_name: str,
+        expected_relation_interface: str,
+        actual_relation_interface: str,
+    ):
+        self.relation_name = relation_name
+        self.expected_relation_interface = expected_relation_interface
+        self.actual_relation_interface = actual_relation_interface
+        self.message = (
+            f"The '{relation_name}' relation has '{actual_relation_interface}' as "
```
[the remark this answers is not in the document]
simskij: This was fixed in one of the commits since.

lib/charms/prometheus_k8s/v0/prometheus_remote_write.py
```diff
@@ -0,0 +1,688 @@
...
+    def _set_alerts_to_all_relation(self, _):
+        for relation in self.model.relations[self._relation_name]:
+            self._set_alerts_to_relation(relation)
+
+    def _set_alerts_to_relation(self, relation: Relation):
+        if alert_groups := self._labeled_alert_groups:
+            relation.data[self._charm.app]["alert_rules"] = json.dumps({"groups": alert_groups})
+
+    @property
+    def endpoints(self) -> List[Dict[str, str]]:
```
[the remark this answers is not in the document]
simskij: Valid point! Let's cross that bridge when/if we get there.

lib/charms/prometheus_k8s/v0/prometheus_remote_write.py
```diff
@@ -0,0 +1,688 @@
...
+        """A config object ready to be dropped in to a prometheus config file.
+
+        Returns:
+            A list of remote_write endpoints.
```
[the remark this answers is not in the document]
simskij:
```suggestion
            A list of dictionaries where each dictionary provides information about a single remote_write endpoint.
```

lib/charms/prometheus_k8s/v0/prometheus_remote_write.py
```diff
@@ -0,0 +1,689 @@
+# Copyright 2021 Canonical Ltd.
+# See LICENSE file for licensing details.
+"""This library facilitates the integration of the prometheus_remote_write interface.
```
sed-i, line 3: Charmhub automatically renders an `<h2>` that looks like this: `charms.alertmanager_k8s.v0.alertmanager_dispatch`.

Should probably add a header of our own:
```suggestion
"""# Prometheus remote-write library

This library facilitates the integration of the prometheus_remote_write interface.
```
A single `#` in markdown is rendered as `<h4>` on charmhub.

lib/charms/prometheus_k8s/v0/prometheus_remote_write.py
```diff
@@ -0,0 +1,689 @@
...
+        """
+        metadata = self._consumer_metadata
+        labels = rule.get("labels", {})
+        labels["juju_model"] = metadata["model"]
+        labels["juju_model_uuid"] = metadata["model_uuid"]
+        labels["juju_application"] = metadata["application"]
+        rule["labels"] = labels
+        return rule
+
+    def _label_alert_expression(self, rule) -> dict:
```
sed-i, line 414: iirc, this function and its neighbors were refactored quite a bit in `prometheus_scrape`.

metadata.yaml
```diff
@@ -24,13 +24,19 @@ containers:
...
   metrics-endpoint:
     interface: prometheus_scrape
   alertmanager:
     interface: alertmanager_dispatch
   ingress:
     interface: ingress
+    limit: 1
+peers:
+  prometheus-peers:
+    interface: prometheus_peers
```
sed-i, line 39: Might be worth to add a comment here that clustered prom is not yet supported.

src/charm.py
```diff
@@ -51,6 +58,14 @@ def __init__(self, *args):
         # Gathers scrape job information from metrics endpoints
         self.metrics_consumer = MetricsEndpointConsumer(self)
 
+        # Exposes remote write endpoints
+        self.remote_write_provider = PrometheusRemoteWriteProvider(
+            self,
+            relation_name="receive-remote-write",
```
sed-i, line 64: Intentional mismatch with `DEFAULT_RELATION_NAME`?

src/charm.py
```diff
@@ -112,14 +131,37 @@ def _configure(self, _):
...
+                return
+            logger.info("Prometheus configuration reloaded")
         else:
             container.add_layer(self._name, new_layer, combine=True)
             container.restart(self._name)
             logger.info("Prometheus (re)started")
-            self.unit.status = ActiveStatus()
+
+        # Ensure the right address is set on the remote_write relations
+        self.remote_write_provider.update_endpoint()
```
sed-i, line 142: Can this be delegated to library code?

src/charm.py
```diff
@@ -112,14 +131,37 @@ def _configure(self, _):
...
+
+        if (
+            isinstance(self.unit.status, BlockedStatus)
+            and self.unit.status.message != INGRESS_MULTIPLE_UNITS_STATUS_MESSAGE
+        ):
+            return
+
+        self.unit.status = ActiveStatus()
+
+    def _validate_ingress_and_remote_write(self) -> bool:
```
sed-i, line 160: Docstring?

src/charm.py
```diff
@@ -161,6 +207,10 @@ def _command(self) -> str:
 
             args.append(f"--web.external-url={external_url}")
 
+        # enable remote write if an instance of the relation exists
+        if self.model.relations["receive-remote-write"]:
```
sed-i, line 211: Import relation name from library instead of string literal?

src/charm.py
```diff
@@ -305,6 +355,10 @@ def _external_hostname(self) -> str:
 
         return self.config["web_external_url"] or f"{self.app.name}"
 
+    @property
+    def _remote_write_address(self) -> str:
```
sed-i, line 359:
```suggestion
    def _remote_write_address(self) -> Optional[str]:
```

tests/unit/test_charm.py
```diff
@@ -26,7 +26,7 @@ def setUp(self):
     @patch("ops.testing._TestingPebbleClient.remove_path")
     @patch("ops.testing._TestingPebbleClient.push")
```
sed-i, line 27: patching push/pull no longer needed.

tests/unit/test_remote_write.py
```diff
@@ -0,0 +1,259 @@
...
+    PrometheusRemoteWriteConsumer,
+)
+from ops.charm import CharmBase
+from ops.model import ActiveStatus, BlockedStatus
+from ops.testing import Harness
+
+from charm import INGRESS_MULTIPLE_UNITS_STATUS_MESSAGE, Prometheus, PrometheusCharm
+
+RELATION_NAME = "receive-remote-write"
+RELATION_INTERFACE = "prometheus_remote_write"
```
sed-i, line 19: Import these from the library?

tests/unit/test_remote_write.py
```diff
@@ -0,0 +1,259 @@
...
+                        "VALUE = {{ $value }}\n  LABELS = {{ $labels }}",
+                    },
+                },
+            ],
+        }
+    ]
+}
+
+
+def fake_push(self, path, content, **kwargs):
```
sed-i, line 77: no longer needed

tests/unit/test_remote_write.py
```diff
@@ -227,9 +227,7 @@ def test_multiple_units_with_ingress(self, mock_net_get, *_):
         )
 
         self.harness.remove_relation_unit(peers_rel_id, "prometheus/1")
-
-        # TODO Uncomment when https://github.com/canonical/operator/issues/638 is fixed
-        # self.assertEqual(self.harness.charm.unit.status, ActiveStatus())
+        self.assertEqual(self.harness.charm.unit.status, ActiveStatus())
```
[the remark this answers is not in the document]
simskij: charm code should never use messages for active status, so that's a non issue. but i'll change it anyway.

lib/charms/prometheus_k8s/v0/prometheus_remote_write.py
simskij: > A single # in markdown is rendered as <h4> on charmhub.

Wait, what? Why? This sounds broken, but I'll go for it for now.

tests/unit/test_remote_write.py
simskij: Still needed from my testing. If you insist it's not, feel free to show me in a separate PR and we'll remove it.

src/charm.py
simskij: I don't think so. Fixed it.

src/charm.py
simskij: Moved to https://github.com/canonical/prometheus-k8s-operator/issues/195

src/charm.py
simskij: Fixed

lib/charms/prometheus_k8s/v0/prometheus_remote_write.py
```diff
@@ -0,0 +1,925 @@
...
+        self.framework.observe(
+            on_relation.relation_created,
+            self._set_endpoint_on_relation_change,
+        )
+        self.framework.observe(
+            on_relation.relation_joined,
+            self._set_endpoint_on_relation_change,
+        )
+
+    def update_endpoint(self, relation: Optional[Relation] = None) -> None:
```
sed-i, line 824: `Optional` is inferred when `None` is assigned
```suggestion
    def update_endpoint(self, relation: Relation = None) -> None:
```

lib/charms/prometheus_k8s/v0/prometheus_remote_write.py
```diff
@@ -0,0 +1,925 @@
...
+        """Triggers programmatically the update of the relation data.
+
+        This method should be used when the charm relying on this library needs
+        to update the relation data in response to something occurring outside
+        of the `prometheus_remote_write` relation lifecycle, e.g., in case of an
```
sed-i, line 829:
```suggestion
        of the `prometheus_remote_write` relation lifecycle, e.g., in case of a
```

lib/charms/prometheus_k8s/v0/prometheus_remote_write.py
```diff
@@ -0,0 +1,925 @@
...
+                continue
+
+            alert_rules = json.loads(relation.data[relation.app].get("alert_rules", "{}"))
+
+            if not alert_rules:
+                continue
+
+            try:
+                for group in alert_rules["groups"]:
+                    alerts[group["name"]] = group
```
sed-i, line 917: I recall refactoring this. Compare to prom scrape lib?

src/charm.py
```diff
@@ -10,9 +10,15 @@
...
-from charms.grafana_k8s.v0.grafana_source import GrafanaSourceConsumer
+from charms.grafana_k8s.v0.grafana_source import GrafanaSourceProvider
 from charms.nginx_ingress_integrator.v0.ingress import IngressRequires
 from charms.observability_libs.v0.kubernetes_service_patch import KubernetesServicePatch
+from charms.prometheus_k8s.v0.prometheus_remote_write import (
+    DEFAULT_RELATION_NAME as DEFAULT_REMOTE_WRITE_RELATION_NAME,
+)
+from charms.prometheus_k8s.v0.prometheus_remote_write import (
+    PrometheusRemoteWriteProvider,
+)
```
sed-i, line 21:
```suggestion
from charms.prometheus_k8s.v0.prometheus_remote_write import (
    DEFAULT_RELATION_NAME as DEFAULT_REMOTE_WRITE_RELATION_NAME,
    PrometheusRemoteWriteProvider,
)
```

src/charm.py
```diff
@@ -42,15 +50,22 @@ def __init__(self, *args):
...
             refresh_event=self.on.prometheus_pebble_ready,
         )
 
         # Gathers scrape job information from metrics endpoints
         self.metrics_consumer = MetricsEndpointConsumer(self)
 
+        # Exposes remote write endpoints
+        self.remote_write_provider = PrometheusRemoteWriteProvider(
+            self,
+            relation_name=DEFAULT_REMOTE_WRITE_RELATION_NAME,
```
sed-i, line 64: Should charm code omit this arg?

src/charm.py
```diff
@@ -71,6 +86,10 @@ def __init__(self, *args):
         self.framework.observe(self.on.ingress_relation_joined, self._configure)
         self.framework.observe(self.on.ingress_relation_changed, self._configure)
         self.framework.observe(self.on.ingress_relation_broken, self._configure)
+        self.framework.observe(self.on.receive_remote_write_relation_created, self._configure)
+        self.framework.observe(self.on.receive_remote_write_relation_broken, self._configure)
+        self.framework.observe(self.on.prometheus_peers_relation_joined, self._configure)
+        self.framework.observe(self.on.prometheus_peers_relation_departed, self._configure)
```
sed-i, line 92: @mmanciop is going to have lots of fun borking multi unit prom deployments

src/charm.py
```diff
@@ -112,14 +131,38 @@ def _configure(self, _):
...
+        if (
+            isinstance(self.unit.status, BlockedStatus)
+            and self.unit.status.message != INGRESS_MULTIPLE_UNITS_STATUS_MESSAGE
+        ):
+            return
+
+        self.unit.status = ActiveStatus()
+
+    def _validate_ingress_and_remote_write(self) -> bool:
+        """Validate that ingresses are created if there are peers."""
```
sed-i, line 161: iiuc docstring contradicts logic

src/charm.py
```diff
@@ -131,10 +174,14 @@ def _set_alerts(self, container):
...
         container.remove_path(RULES_DIR, recursive=True)
 
-        for topology_identifier, rules_file in self.metrics_consumer.alerts().items():
-            filename = "juju_" + topology_identifier + ".rules"
+        self._push_alert_rules_group(container, self.metrics_consumer.alerts())
+        self._push_alert_rules_group(container, self.remote_write_provider.alerts())
+
+    def _push_alert_rules_group(self, container, alerts):
+        for group_name, group in alerts.items():
+            filename = "juju_" + group_name + ".rules"
```
sed-i, line 182: I think the spec is one file per topology identifier, not one file per group name
https://github.com/canonical/prometheus-k8s-operator/blob/c285372fa51c6ef031239d39c4c66c5255ca6a48/src/charm.py#L134-L136

tests/integration/helpers.py
```diff
@@ -144,3 +165,46 @@ def oci_image(metadata_file: str, image_name: str) -> str:
         raise ValueError("Upstream source not found")
 
     return upstream_source
+
+
+def initial_workload_is_ready(ops_test, app_names) -> bool:
+    """Checks that the initial workload (ie. x/0) is ready.
```
sed-i, line 171: When is this preferred over `wait_for_idle`?

tests/integration/test_charm.py
```diff
@@ -11,90 +12,63 @@
...
-    await ops_test.model.deploy(
-        prometheus_tester_charm, resources=tester_resources, application_name=tester_app_name
-    )
-    await ops_test.model.wait_for_idle(apps=[prometheus_app_name], status="active")
-    await ops_test.model.wait_for_idle(apps=[tester_app_name], status="active")
-    await ops_test.model.block_until(
-        lambda: len(ops_test.model.applications[prometheus_app_name].units) > 0
+    app_names = [prometheus_app_name, tester_app_name]
+
+    await asyncio.gather(
```
sed-i, line 31: Given the limited resources of ghwf, maybe we should avoid using gather for deployments.

tests/unit/test_charm.py
```diff
@@ -41,14 +41,14 @@ def test_grafana_is_provided_port_and_source(self, mock_net_get, *unused):
 
     @patch("ops.testing._TestingPebbleClient.remove_path")
     @patch("ops.testing._TestingPebbleClient.push")
```
sed-i, line 43: no need to patch fs anymore

tox.ini
```diff
@@ -83,11 +83,13 @@ description = Run integration tests
 deps =
     aiohttp
     git+https://github.com/juju/python-libjuju.git
-    #juju
+    juju
```
sed-i, line 86: On pypi, python-libjuju is just juju. We use it from gh.
```suggestion
    #juju
```

lib/charms/prometheus_k8s/v0/prometheus_remote_write.py
```diff
@@ -0,0 +1,925 @@
...
+            "juju_model": self.model,
+            "juju_model_uuid": self.model_uuid,
+            "juju_application": self.application,
+            "juju_charm": self.charm_name,
+        }
+
+    def render(self, template: str):
+        """Render a juju-topology template string with topology info."""
+        return template.replace(JujuTopology.STUB, self.promql_labels)
+
```
rbarry82, line 202: This is ok for now, but `JujuTopology` has fairly large changes in the PR pending to prometheus to avoid this set of properties with boilerplate strings, and instead builds them from `as_dict`

lib/charms/prometheus_k8s/v0/prometheus_remote_write.py
```diff
@@ -0,0 +1,925 @@
...
+    The one exception to the rule above, is if you charm needs to both consume and provide a
+    relation using the `prometheus_remote_write` interface, in which case changing the relation
+    name to differentiate between "incoming" and "outgoing" remote write interactions is necessary.
+    """
+
+    def __init__(
+        self,
+        charm: CharmBase,
+        relation_name: str = DEFAULT_RELATION_NAME,
+        endpoint_schema: str = "http",
```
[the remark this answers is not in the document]
rbarry82: As with the other charm, I don't have strong opinions on that PEP either way except to say that it's not widespread.

That said, the "UX goal" in most of our libraries is "all you need to do is instantiate the class, and the rest takes defaults". Preventing users from providing them positionally is ok (even if there's also a PEP for positional-only arguments), but I'd honestly be surprised if anyone provided enough arguments to even run into this problem..

It's the internal/relation data API that I worry about breaking more than class instantiation.

tests/integration/test_charm.py
simskij: In my experience it makes no difference. After all, it's concurrent rather than parallel execution.

tests/unit/test_charm.py
simskij: I tried removing these but still got errors. If you disagree, please provide an example on how to fix it.

src/charm.py
simskij: fixed

tests/integration/helpers.py
simskij: it is not meant as a replacement for `wait_for_idle`. this is used to specifically check that unit `/0` is in a ready state, and is to be used in assertions.

lib/charms/prometheus_k8s/v0/prometheus_remote_write.py
```diff
@@ -0,0 +1,925 @@
...
+    ```
+
+    which returns a dictionary structured like the Prometheus configuration object (see
+    https://prometheus.io/docs/prometheus/latest/configuration/configuration/#remote_write).
+
+    Regarding the default relation name, `prometheus-remote-write`: if you choose to change it,
+    you would need to explicitly provide it to the `PrometheusRemoteWriteConsumer` via the
+    `relation_name` constructor argument. (The relation interface, on the other hand, is
+    immutable and, if you were to change it, your charm would not be able to relate with other
+    charms using the right relation interface. The library prevents you from doing that by
```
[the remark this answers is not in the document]
simskij: fixed

lib/charms/prometheus_k8s/v0/prometheus_remote_write.py
```diff
@@ -0,0 +1,925 @@
...
+    self.remote_write_consumer.endpoints
+    ```
+
+    which returns a dictionary structured like the Prometheus configuration object (see
+    https://prometheus.io/docs/prometheus/latest/configuration/configuration/#remote_write).
+
+    Regarding the default relation name, `prometheus-remote-write`: if you choose to change it,
+    you would need to explicitly provide it to the `PrometheusRemoteWriteConsumer` via the
+    `relation_name` constructor argument. (The relation interface, on the other hand, is
+    immutable and, if you were to change it, your charm would not be able to relate with other
```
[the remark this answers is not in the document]
simskij: fixed

lib/charms/grafana_k8s/v0/grafana_source.py
```diff
@@ -19,10 +33,123 @@
 
 # Increment this PATCH version before using `charmcraft publish-lib` or reset
 # to 0 if you are raising the major API version
-LIBPATCH = 1
+LIBPATCH = 5
```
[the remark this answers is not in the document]
simskij: thanks. done

lib/charms/prometheus_k8s/v0/prometheus_remote_write.py
```diff
@@ -0,0 +1,925 @@
...
+    relation names across charms that do similar things is good practice and more
+    straightforward for the users of your charm. The one exception to the rule above,
+    is if you charm needs to both consume and provide a relation using the
+    `prometheus_remote_write` interface, in which case changing the relation name to
+    differentiate between "incoming" and "outgoing" remote write interactions is necessary.
+
+    It is also possible to specify alert rules. By default, this library will look
+    into the `<charm_parent_dir>/prometheus_alert_rules`, which in standard charm
+    layouts resolves to `src/prometheus_alert_rules`. Each alert rule goes into a
+    separate `*.rule` file. If the syntax of a rule is invalid,
```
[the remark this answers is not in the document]
simskij: Fixed

lib/charms/prometheus_k8s/v0/prometheus_remote_write.py
```diff
@@ -0,0 +1,925 @@
...
+
+    def _set_alerts_on_relation_changed(self, event: RelationEvent):
+
+        self._set_alerts_to_relation(event.relation)
+
+    def _set_alerts_to_all_relation(self, _):
+        for relation in self.model.relations[self._relation_name]:
+            self._set_alerts_to_relation(relation)
+
+    def _set_alerts_to_relation(self, relation: Relation):
```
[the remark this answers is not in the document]
simskij: renamed to clarify intent.

lib/charms/prometheus_k8s/v0/prometheus_remote_write.py
```diff
@@ -0,0 +1,925 @@
...
+            for unit in relation.units:
+                if unit.app is self._charm.app:
+                    # This is a peer unit
+                    continue
+
+                if remote_write := relation.data[unit].get("remote_write"):
+                    deserialized_remote_write = json.loads(remote_write)
+                    endpoints.append(
+                        {
+                            "url": deserialized_remote_write["url"],
```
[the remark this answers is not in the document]
simskij: Not sure if I get what you're wondering about, but if your question is why it's not just a list of url:s directly, it's because we might (will) want to add more properties to each entry down the road (think for instance TLS).

lib/charms/prometheus_k8s/v0/prometheus_remote_write.py
```diff
@@ -0,0 +1,925 @@
...
+        self.topology = JujuTopology.from_charm(charm)
+
+        on_relation = self._charm.on[self._relation_name]
+
+        self.framework.observe(on_relation.relation_joined, self._handle_endpoints_changed)
+        self.framework.observe(on_relation.relation_changed, self._handle_endpoints_changed)
+        self.framework.observe(on_relation.relation_departed, self._handle_endpoints_changed)
+        self.framework.observe(on_relation.relation_broken, self._handle_endpoints_changed)
+        self.framework.observe(on_relation.relation_joined, self._set_alerts_on_relation_changed)
+        self.framework.observe(self._charm.on.upgrade_charm, self._set_alerts_to_all_relation)
```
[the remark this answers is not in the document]
simskij: Added. Thanks

lib/charms/prometheus_k8s/v0/prometheus_remote_write.py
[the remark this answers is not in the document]
simskij: Not doing this for now.

src/charm.py
simskij: fixed, thanks for spotting

src/charm.py
simskij: Probably! In the remote_write case multi-unit will put the app in a blocked state.

src/charm.py
simskij: Fixed. Thanks!

sed-i:
@simskij I updated the unit tests. Here's a PR to merge into remote write if you would like
https://github.com/canonical/prometheus-k8s-operator/pull/196

simskij:
> @simskij I updated the unit tests. Here's a PR to merge into remote write if you would like #196

Thanks! That PR unfortunately fails with the same error I was seeing locally 😅

sed-i:
> > @simskij I updated the unit tests. Here's a PR to merge into remote write if you would like #196
> 
> Thanks! That PR unfortunately fails with the same error I was seeing locally sweat_smile

Now that https://github.com/canonical/operator/pull/677 is merged, tests are passing.

PR #149, opened 2021-10-15
Allow a non metrics-provider charm to provide rules

sed-i:
An existing use case is to have rules coming in dynamically from a constantly-updating git repo.
For this purpose, a new mechanism is needed for continually forwarding rules information as they are updated.

Prometheus config file is comprised of the following sections (per @mmanciop):

```
global:
  ... # set by juju config of this charm (prometheus)

rule_files:
  ... # THIS PR

scrape_configs:
... # Covered by the prometheus-scrape relation (prometheus_scrape.MetricsEndpointProvider and Consumer)

alerting:
... # alertmanager interface. Partially covered by the prometheus charm code.

remote_write:
... # prometheus_remote_write interface (#147)

remote_read:
... # We may never implement this and if we do, it is a prometheus_remote_read interface
```

This PR is for allowing a non metrics-provider charm to provide rules, for example:

```python
self.prom_config_subset = RuleFilesProvider(
    self,
    "prometheus-config",
    dir_path="/git/repo/prom_rules",
    recursive=True,
)
```

- [x] add JujuTopology class
- [x] convert rule-parsing methods into free functions
- [x] Create a new RuleFilesProvider

### DIFF
--- a/lib/charms/prometheus_k8s/v0/prometheus_scrape.py
+++ b/lib/charms/prometheus_k8s/v0/prometheus_scrape.py
@@ -299,7 +299,7 @@ import logging
 import os
 from collections import defaultdict
 from pathlib import Path
-from typing import List, Union
+from typing import List, Optional, Union
 
 import yaml
 from ops.charm import CharmBase, RelationMeta, RelationRole
@@ -544,6 +544,33 @@ class JujuTopology:
         return template.replace("%%juju_topology%%", self.promql_labels)
 
 
+def load_alert_rule_from_file(path: Path, topology: JujuTopology) -> Optional[dict]:
+    """Load alert rule from a rules file.
+
+    Args:
+        path: path to a *.rule file with a single rule ("groups" super section omitted).
+        topology: a `JujuTopology` instance.
+    """
+    with path.open() as rule_file:
+        # Load a list of rules from file then add labels and filters
+        try:
+            rule = yaml.safe_load(rule_file)
+        except Exception as e:
+            logger.error("Failed to read alert rules from %s: %s", path.name, e)
+            return None
+        else:
+            # add "juju_" topology labels
+            rule.get("labels", {}).update(topology.as_dict_with_promql_labels())
+
+            if "expr" not in rule:
+                logger.error("Invalid alert rule %s: missing an 'expr' property.", path.name)
+                return None
+
+            # insert juju topology filters into a prometheus alert rule
+            rule["expr"] = topology.render(rule["expr"])
+            return rule
+
+
 def load_alert_rules_from_dir(
     dir_path: Union[str, Path], topology: JujuTopology, *, recursive: bool = False
 ) -> List[dict]:
@@ -563,35 +590,29 @@ def load_alert_rules_from_dir(
     """
     alerts = defaultdict(list)
 
-    for path in filter(Path.is_file, Path(dir_path).glob("**/*.rule" if recursive else "*.rule")):
+    def _group_name(path) -> str:
+        """Generate group name from path and topology.
+
+        The group name is made up of the relative path between the root dir_path, the file path,
+        and topology identifier.
+
+        Args:
+            path: path to rule file.
+        """
         relpath = os.path.relpath(os.path.dirname(path), dir_path)
 
         # Generate group name:
         #  - prefix, from the relative path of the rule file;
         #  - name, from juju topology
-        group_name = (
+        return (
             f"{'' if relpath == '.' else relpath.replace(os.path.sep, '_') + '_'}"
             f"{topology.identifier}_alerts"
         )
 
-        logger.debug("Reading alert rule from %s", path)
-        with path.open() as rule_file:
-            # Load a list of rules from file then add labels and filters
-            try:
-                rule = yaml.safe_load(rule_file)
-            except Exception as e:
-                logger.error("Failed to read alert rules from %s: %s", path.name, e)
-                continue
-            try:
-                # add "juju_" topology labels
-                rule.get("labels", {}).update(topology.as_dict_with_promql_labels())
-
-                # insert juju topology filters into a prometheus alert rule
-                rule["expr"] = topology.render(rule["expr"])
-            except KeyError:
-                logger.error("Invalid alert rule %s: missing an 'expr' property.", path.name)
-            else:
-                alerts[group_name].append(rule)
+    for path in filter(Path.is_file, Path(dir_path).glob("**/*.rule" if recursive else "*.rule")):
+        if rule := load_alert_rule_from_file(path, topology):
+            logger.debug("Reading alert rule from %s", path)
+            alerts[_group_name(path)].append(rule)
 
     # Gather all alerts into a list of groups since Prometheus
     # requires alerts be part of some group


lib/charms/prometheus_k8s/v0/prometheus_scrape.py
```diff
@@ -1182,15 +1220,54 @@ def _scrape_metadata(self) -> dict:
...
+        self._relation_name = relation_name
+        self.topology = JujuTopology.from_charm(charm)
+        self.dir_path = dir_path
+        self.recursive = recursive
+
+        events = self._charm.on[self._relation_name]
+        self.framework.observe(events.relation_joined, self._update_relation_data)
+        self.framework.observe(events.relation_changed, self._update_relation_data)
+        self.framework.observe(self._charm.on.leader_elected, self._update_relation_data)
+        self.framework.observe(self._charm.on.config_changed, self._update_relation_data)
```
mmanciop, line 1254: Todo here still relevant?

lib/charms/prometheus_k8s/v0/prometheus_scrape.py
```diff
@@ -474,6 +476,106 @@ def _sanitize_scrape_configuration(job) -> dict:
...
+    for path in filter(Path.is_file, Path(dir_path).glob("**/*.rule" if recursive else "*.rule")):
+        relpath = os.path.relpath(os.path.dirname(path), dir_path)
+
+        # Generate group name:
+        #  - prefix, from the relative path of the rule file;
+        #  - name, from juju topology
+        group_name = (
+            f"{'' if relpath == '.' else relpath.replace(os.path.sep, '_') + '_'}"
+            f"{topology.as_str_short_form()}_alerts"
+        )
```
Abuelodelanada, line 552: This kind of comments sometimes indicates that is worth extracting this lines into a small function, for instance: `_generate_group_name(path, topology)`

lib/charms/prometheus_k8s/v0/prometheus_scrape.py
```diff
@@ -474,6 +476,106 @@ def _sanitize_scrape_configuration(job) -> dict:
...
+                try:
+                    # add "juju_" topology labels
+                    rule.get("labels", {}).update(topology.as_dict_long_form())
+
+                    # insert juju topology filters into a prometheus alert rule
+                    rule["expr"] = topology.render(rule["expr"])
+                except KeyError:
+                    logger.error("Invalid alert rule %s: missing an 'expr' property.", path.name)
+                else:
+                    alerts[group_name].append(rule)
```
Abuelodelanada, line 571: The same here, maybe it's worth extracting this code into a custom function.

Besides, shouldn't we raise custom exceptions if `Exception` and `KeyError` is catched?

lib/charms/prometheus_k8s/v0/prometheus_scrape.py
```diff
@@ -1182,15 +1220,54 @@ def _scrape_metadata(self) -> dict:
...
-            "charm_name": f"{self._charm.meta.name}",
-        }
-        return metadata
+        return self.topology.as_dict()
+
+
+class RuleFilesProvider(Object):
+    """A 'prometheus_scrape' provider class for rule files only.
+
+    The RuleFilesProvider class if for sending rules to prometheus, unlike
```
Abuelodelanada, line 1231: The RuleFilesProvider class **is** for sending...

lib/charms/prometheus_k8s/v0/prometheus_scrape.py
```diff
@@ -1182,15 +1220,54 @@ def _scrape_metadata(self) -> dict:
...
+     (including rules). This is useful for providing "free standing" rules.
+
+    Args:
+        charm: a charm instance that has the `prometheus_scrape` interface on the "provides" side.
+        relation_name: must match metadata.yaml and have `prometheus_scrape` as interface.
+        dir_path: root directory for the collection of rule files.
+        recursive: whether or not to scan for rule files recursively.
+    """
+
+    def __init__(self, charm: CharmBase, relation_name: str, dir_path: str, recursive=True):
```
Abuelodelanada: What about having a default value for `relation_name`?

lib/charms/prometheus_k8s/v0/prometheus_scrape.py
```diff
@@ -1182,15 +1220,54 @@ def _scrape_metadata(self) -> dict:
...
+        super().__init__(charm, relation_name)
+        self._charm = charm
+        self._relation_name = relation_name
+        self.topology = JujuTopology.from_charm(charm)
+        self.dir_path = dir_path
+        self.recursive = recursive
```
Abuelodelanada, line 1256: If these properties are not going to be public, I would prefix them with  `_`

lib/charms/prometheus_k8s/v0/prometheus_scrape.py
```diff
@@ -474,6 +476,106 @@ def _sanitize_scrape_configuration(job) -> dict:
...
+    """Dataclass for storing and formatting juju topology information."""
+
+    model: str
+    model_uuid: str
+    application: str
+    charm_name: str
+
+    @staticmethod
+    def from_charm(charm):
+        """Factory method for creating the topology dataclass from a given charm."""
```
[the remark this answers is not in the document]
sed-i: I only refactored the existing.
Not sure what you mean: replace charm_name with (sanitized) unit name?

lib/charms/prometheus_k8s/v0/prometheus_scrape.py
```diff
@@ -474,6 +476,106 @@ def _sanitize_scrape_configuration(job) -> dict:
...
+        return JujuTopology(
+            model=charm.model.name,
+            model_uuid=charm.model.uuid,
+            application=charm.model.app.name,
+            charm_name=charm.meta.name,
+        )
+
+    def as_str_short_form(self) -> str:
```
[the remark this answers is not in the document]
sed-i: Should probably be changed to
```python
@property
def identifier(self) -> str:
```
?

lib/charms/prometheus_k8s/v0/prometheus_scrape.py
```diff
@@ -474,6 +476,106 @@ def _sanitize_scrape_configuration(job) -> dict:
...
+        """Format the topology information into a terse string."""
+        return f"{self.model}_{self.model_uuid}_{self.application}"
+
+    def as_str_long_form(self) -> str:
```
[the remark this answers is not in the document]
sed-i: Should probably be changed to
```python
@property
def promql_labels(self) -> str:
```
?

lib/charms/prometheus_k8s/v0/prometheus_scrape.py
```diff
@@ -474,6 +476,106 @@ def _sanitize_scrape_configuration(job) -> dict:
...
+            "juju_model_uuid": self.model_uuid,
+            "juju_application": self.application,
+        }
+
+    def render(self, template: str):
+        """Render a juju-topology template string with topology info."""
+        return template.replace("%%juju_topology%%", self.as_str_long_form())
+
+
+def load_alert_rules_from_dir(
```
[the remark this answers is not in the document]
sed-i: Good question. This function seems like it qualifies as a public interface. Not sure.

lib/charms/prometheus_k8s/v0/prometheus_scrape.py
sed-i: Not sure about this: relpath makes little sense outside of this function so I am more inclined towards an inner function, wdyt?

lib/charms/prometheus_k8s/v0/prometheus_scrape.py
sed-i: > it's worth extracting this code into a custom function

one option is to pass the `alerts` "by-ref":
`read_rule_file(alerts, group_name, path, rule_file, topology)`

another option is an inner function:
`read_rule_file(group_name, path, rule_file)`

wdyt?

> raise custom exceptions

Per discussion with @balbirthomas, and in line with the original code, file reading should proceed even if one of them is buggy.

lib/charms/prometheus_k8s/v0/prometheus_scrape.py
sed-i: added `_` to `recursive`. I wonder if `dir_path` and `topology` make sense as public.

PR #150, opened 2021-10-18
Prometheus Integration tests

balbirthomas:
This pull request adds Prometheus specific integration tests.


tests/integration/test_charm.py
```diff
@@ -5,19 +5,104 @@
...
+async def test_build_and_deploy_with_alternative_images(ops_test, prometheus_charm, oci_image):
     """Test that the Prometheus charm can be deployed successfully."""
-    prometheus_charm = await ops_test.build_charm(".")
-    resources = {"prometheus-image": "ubuntu/prometheus:latest"}
+    resources = {"prometheus-image": oci_image}
+    app_name = "prometheus-" + oci_image.split("/")[0]
 
+    await ops_test.model.deploy(prometheus_charm, resources=resources, application_name=app_name)
+    await ops_test.model.wait_for_idle(apps=[app_name], status="active")
+    await ops_test.model.block_until(lambda: len(ops_test.model.applications[app_name].units) > 0)
```
sed-i, line 28: Isn't this already captured by `wait_for_idle(..., status="active")`?

tests/integration/test_charm.py
```diff
@@ -5,19 +5,104 @@
...
+
+    assert ops_test.model.applications[app_name].units[0].workload_status == "active"
+
+    await check_prometheus_is_ready(ops_test, app_name, 0)
+
+    await ops_test.model.applications[app_name].remove()
+    await ops_test.model.reset()
```
sed-i, line 35: Why not carry forward the built app?

tests/integration/test_charm.py
```diff
@@ -5,19 +5,104 @@
...
+    await ops_test.model.wait_for_idle(apps=[prometheus_app_name], status="active")
+    config_with_relation = await get_prometheus_config(ops_test, prometheus_app_name, 0)
+    tester_job = get_job_config_for(tester_app_name, config_with_relation)
+    assert tester_job != {}
+
+    rules_with_relation = await get_prometheus_rules(ops_test, prometheus_app_name, 0)
+    tester_rules = get_rules_for(tester_app_name, rules_with_relation)
+    assert tester_rules != {}
+
+    await ops_test.model.applications[tester_app_name].remove()
```
sed-i, line 98: I would move this into a different test.
In pytest, tests are run sequentially and the ops_test plugin takes care of cleanup.

tests/integration/workload.py
```diff
@@ -0,0 +1,53 @@
+import logging
+
+import aiohttp
+
+logger = logging.getLogger(__name__)
+
+
+class Prometheus:
+    """A class that represents a running instance of Prometheus."""
```
sed-i, line 9: Could we somehow augment+use the existing `prometheus_server.py`?

sed-i (commented):
This is a good addition. See comments regarding design choices:
- charm-within-a-charm instead of avalanche
- build-and-remove instead of carry charms forward

tests/integration/conftest.py
```diff
@@ -0,0 +1,26 @@
...
+    """Ensure that the tester charm uses the current Prometheus library."""
+    library_path = "lib/charms/prometheus_k8s/v0/prometheus_scrape.py"
+    install_path = "tests/integration/prometheus-tester/" + library_path
+    shutil.copyfile(library_path, install_path)
+
+
+@pytest.fixture(scope="module")
+async def prometheus_charm(ops_test):
+    """Prometheus charm used for integration testing."""
+    charm = await ops_test.build_charm(".")
```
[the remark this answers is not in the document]
balbirthomas: As seen in the logs you will see the charm is built only once.

tests/integration/conftest.py
```diff
@@ -0,0 +1,26 @@
...
+    return charm
+
+
+@pytest.fixture(scope="module")
+async def prometheus_tester_charm(ops_test):
+    """A charm to integration test the Prometheus charm."""
+    charm_path = "tests/integration/prometheus-tester"
+    charm = await ops_test.build_charm(charm_path)
+    return charm
```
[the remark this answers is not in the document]
balbirthomas: 1. As mentioned above charm is built only once.
2. ditto

tests/integration/test_charm.py
balbirthomas: I want each test to be independent of the other.

tests/integration/test_charm.py
balbirthomas: As mentioned above. I would like each test to be independent of the other and want each test to start with a clean slate.

tests/integration/workload.py
balbirthomas: They are built for different purposes, use different packages , different request processing approaches and will diverge considerably as tests evolve.

tests/integration/workload.py
```diff
@@ -0,0 +1,53 @@
...
+    def __init__(self, host="localhost", port=9090):
+        """Utility to manage a Prometheus application.
+
+        Args:
+            host: Optional; host address of Prometheus application.
+            port: Optional; port on which Prometheus service is exposed.
+        """
+        self.base_url = f"http://{host}:{port}"
+
+    async def is_ready(self):
```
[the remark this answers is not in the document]
balbirthomas: Why not ?

tests/integration/prometheus-tester/src/charm.py
```diff
@@ -0,0 +1,111 @@
+#!/usr/bin/env python3
+# Copyright 2021 Canonical Ltd.
+# See LICENSE file for licensing details.
+
+"""A Charm to functionally test the Prometheus Operator."""
```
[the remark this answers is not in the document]
balbirthomas: This is just the scaffolding for future tests that provides a completely controlled way to implement functional tests. The tester charm can be instrumented in many different ways for additional tests as will be done in future PRs.

tests/integration/test_charm.py
```diff
@@ -5,19 +5,104 @@
...
+    get_prometheus_rules,
+    get_rules_for,
+)
 
 logger = logging.getLogger(__name__)
 
 
 @pytest.mark.abort_on_fail
-async def test_build_and_deploy_with_ubuntu_image(ops_test):
+@pytest.mark.parametrize("oci_image", ["ubuntu/prometheus:latest", "prom/prometheus:latest"])
```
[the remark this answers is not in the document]
balbirthomas: That is @simskij call. I am not aware if we are version fixing images for our charms.

tests/integration/helpers.py
```diff
@@ -0,0 +1,110 @@
...
+async def get_prometheus_config(ops_test, app_name, unit_num):
+    """Fetch Prometheus configuration.
+
+    Args:
+        ops_test: pytest-operator plugin
+        app_name: string name of Prometheus application
+        unit_num: integer number of a Prometheus juju unit
+
+    Returns:
+        Prometheus YALM configuration in string format.
```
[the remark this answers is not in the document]
balbirthomas: Ack. Will fix.

tests/integration/test_charm.py
balbirthomas: As far as I am aware it is not.

tests/integration/helpers.py
```diff
@@ -0,0 +1,110 @@
+import yaml
+from workload import Prometheus
+
+
+async def unit_address(ops_test, app_name, unit_num):
```
jnsgruk, line 5:
```suggestion
from pytest_operator.plugin import OpsTest

async def unit_address(ops_test: OpsTest, app_name: str, unit_num: int) -> str:
```

tests/integration/helpers.py
```diff
@@ -0,0 +1,110 @@
...
+        unit_num: integer number of a juju unit
+
+    Returns:
+        unit address as a string
+    """
+    status = await ops_test.model.get_status()
+    return status["applications"][app_name]["units"][f"{app_name}/{unit_num}"]["address"]
+
+
+async def check_prometheus_is_ready(ops_test, app_name, unit_num):
```
jnsgruk, line 20:
```suggestion
async def check_prometheus_is_ready(ops_test: OpsTest, app_name: str, unit_num: int) -> bool:
```

jnsgruk: (I won't add suggestions or all, but seems like adding type annotations to helper methods is a good idea throughout)

tests/integration/prometheus-tester/src/charm.py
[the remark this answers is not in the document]
jnsgruk: Yeh I don't see a particular issue with this, although if this tester charm becomes large, it may well be worth having it in its own repo?

tests/integration/test_charm.py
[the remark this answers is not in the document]
jnsgruk: The Kubeflow team tend to use the `upstream-image` parameter in their `metadata.yaml` pretty much specifically for enabling this kind of testing. We also don't really need to support both images, picking one (probably the `ubuntu` one) is fine.

tests/integration/test_charm.py
jnsgruk: Hmm, this does seem like a strange ordering -- we first check that the app is `active/idle`, then block on number of units being > 0? Surely this will never block?

tests/integration/test_charm.py
```diff
@@ -5,19 +5,104 @@
...
+async def test_build_and_deploy_prometheus_tester(ops_test, prometheus_tester_charm):
+    """Test that Prometheus tester charm can be deployed successfully."""
+    resources = {"prometheus-tester-image": "python:slim"}
+    app_name = "prometheus-tester"
+
+    await ops_test.model.deploy(
+        prometheus_tester_charm, resources=resources, application_name=app_name
+    )
+    await ops_test.model.wait_for_idle(apps=[app_name], status="active")
+    await ops_test.model.block_until(lambda: len(ops_test.model.applications[app_name].units) > 0)
```
jnsgruk, line 48: As above, unless these methods are misleading me, this seems like an odd ordering?

tests/integration/workload.py
jnsgruk: I'm sort of with @sed-i here - it seems odd to be testing a different `Prometheus` class than the one the charm uses?

tests/integration/helpers.py
balbirthomas: Ack. Will add type hints.

tests/integration/test_charm.py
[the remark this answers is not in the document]
simskij: Thought I posted a comment on this yesterday. I'll try to recollect what I wrote then.

The charm locks down the digest for the image on publish (see for instance https://charmhub.io/prometheus-k8s/resources/prometheus-image). 

I agree that we should pin it down to whatever is available in the published charm(s) for charms **other** than the one we're testing **from**. For instance, if we were to do an integration test with prometheus and grafana, from this repo, I'd strongly suggest we use the digest that is published for the grafana charm rather than `latest`.

For the charm under test on the other hand, I think it makes sense to use whatever digest would be set in the next revision if we were to publish the code we're testing, rather than what is already there.

tests/integration/test_charm.py
balbirthomas: I do find that `wait_for_idle` does not wait for units to come up. The doc string of the method only states that it waits for "applications in the model to settle into an idle state" with a timeout. It does not guarantee that the requested number of units will be active.

tests/integration/workload.py
balbirthomas: I think there is something lost in translation here. I am not testing this `Prometheus` class. This `Prometheus` class only provides a growing list of *asynchronous* methods to interact with the deployed Prometheus application. Future PRs will grow this class to query various aspects of Prometheus in ways that are only relevant for integration testing but not for actually managing the Prometheus instance.

tests/integration/test_charm.py
balbirthomas: I do find that reversing the ordering makes the test flaky.

tests/integration/test_charm.py
balbirthomas: I do find that reversing the ordering or removing the `block_until` requested units are up makes the tests flaky.

tests/integration/prometheus-tester/src/charm.py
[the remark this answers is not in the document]
balbirthomas: I don't think the tester charm will be any more complex than it is. The goal here is to have a single self contained repository that can functionally test this charm with an actually deployed Prometheus workload (as opposed to unit tests). The second motivation is to be able to probe the behaviour of this Prometheus charm under test more incisively than can be done with another charm who's codebase can not be instrumented in arbitrary ways.

tests/integration/test_charm.py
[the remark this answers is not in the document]
balbirthomas: I did not use `upstream-image` as I was asked to remove this in a previous code review. So should I remove the use upstream Prometheus image and add back `upstream-image` in `metadata.yaml` ?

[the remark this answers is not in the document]
mmanciop: I recall @jnsgruk had agreed using the upstream-image tag was useful both as docu for the end user, and for these testing scenarios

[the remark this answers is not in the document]
sed-i: If we're going with only one image then yes, using upstream image is easiest. I use it in alertmanager exactly for this.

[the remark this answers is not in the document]
balbirthomas: So only test with Ubuntu latest ?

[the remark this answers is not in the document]
mmanciop: There has been a lot of back-and-forth in PM land about what we can support, and it is not settled yet. As of now, the ROCK is the only thing we can _certainly_ support so let’s go with that.

tests/integration/test_charm.py
sed-i: App status is supposed to be set automatically to the " worst across all units".
Any chance charm code is touchig app status? Or maybe charm code sets unit status to active prematurely?

balbirthomas: The Prometheus charm only sets `ActiveStatus` for unit status (see `charm.py`). It does this either after reloading the prometheus configuration or setting a new pebble layer. I do not see how this can be considered "premature". Further these integrations tests (as yet) deploy only a single unit.

tests/integration/test_charm.py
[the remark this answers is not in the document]
balbirthomas: OK. Removed test with upstream Prometheus image retaining it only with ubuntu latest rock.

tests/integration/test_charm.py
```diff
@@ -5,19 +5,103 @@
...
 
 logger = logging.getLogger(__name__)
 
 
 @pytest.mark.abort_on_fail
-async def test_build_and_deploy_with_ubuntu_image(ops_test):
+async def test_build_and_deploy_with_alternative_images(ops_test, prometheus_charm):
     """Test that the Prometheus charm can be deployed successfully."""
-    prometheus_charm = await ops_test.build_charm(".")
     resources = {"prometheus-image": "ubuntu/prometheus:latest"}
```
sed-i, line 22:
```python
resources = {"prometheus-image": METADATA["resources"]["prometheus-image"]["upstream-source"]}
```

tests/integration/test_charm.py
sed-i: Looking at `_configure()` it seems that it may be possible that a unit status is set to active before startup sequence is complete. Not sure. For example:
1. config changed - > pebble ready -> (charm code changes status to active) -> (something else, e.g. leadership event) -> start
1. pebble ready -> (charm code changes status to active) -> config changed -> start
1. a subsequent call to `_configure` flips status to Blocked because `_is_valid_timespec` failed for some reason?

balbirthomas: Not sure what you are trying to say here.
1. If pebble is not ready when config changed executes `_configure()` returns from `can_connect()`. There is no use case for leadership or start here.
2. If pebble ready executes before config changed then a Prometheus configuration will be pushed and workload will be restarted or reloaded. So unit is marked as active which is valid.
3. If at any point Prometheus receives an invalid configuration it will set its status to blocked. This is as desired. This is certainly not the case in this the integration test in this review as may be easily checked by logs and even dumping the config.

sed-i: If case (2) is possible (haven't been paying attention if in recent juju versions pebble ready can fire before config changed), then it can:
`pebble ready -> (charm code changes status to active) -> config changed (status is automatically changed to maintenance or something else) -> _configure changes status back to active -> start`
However, [default idle period](https://github.com/juju/python-libjuju/blob/c68ee681995dc4b966ed14160d61a72f249c3b7b/juju/model.py#L2554) is 15 sec so that is probably not the case. :thinking:

sed-i: I do think however that if `await ops_test.model.wait_for_idle(apps=[app_name], status="active")` is not sufficient then your integration test found a bug in charm code. (Congratulations! :stuck_out_tongue_closed_eyes: )

balbirthomas: Again I am totally lost on what you are trying to say. Both pebble ready and config changed events are handled by the same method `_configure()`. If pebble ready runs before config changed then this method ensures that the workload is indeed "activated". When config changed runs after pebble ready, if nothing else has changed in the mean time the config changed method has no effect.

balbirthomas: > I do think however that if `await ops_test.model.wait_for_idle(apps=[app_name], status="active")` is not sufficient then your integration test found a bug in charm code. (Congratulations! stuck_out_tongue_closed_eyes )

Can you explain why
1. It is guaranteed that `ops_test.model.wait_for_idle` will ensure that the required number of units are deployed ?
2. Even if `ops_test.model.wait_for_idle` does guarantee that required number of units are deployed, why should blocking on the required number of units after that make any difference ?
3. How does this demonstrate any bug in the charm code ?

sed-i: Watching `juju status` during my deployments I see the app status only settles to idle-active when the startup sequence is complete. If units are being brought up I think the status is "executing" or "maintenance". So wait_for_idle, in this context, should _indirectly_ guarantee that all units are up. The wait period (default of 15 sec) is used to makes sure it is not a fluke.
afaiu, If you are getting idle-active for 15 sec before the unit is up then it may be a bug in charm code. Feel free to bring a counter example.

tests/integration/workload.py
sed-i: These methods do not need to be async because nothing else is done in the meanwhile: each and every one of them is awaited (see `asyncio.run`, `asyncio.gather`). Also, each and every one of them is a good fit for `prometheus_server.py`.

tests/integration/prometheus-tester/src/charm.py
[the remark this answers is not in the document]
sed-i: I still believe `prometheus-tester` should not be needed (should not exist):
- test scraping/remote-write + alerting with avalanche and test-specific rule file
- test relations etc. with real charms: deploy alertmanager and/or grafana as part of your integration test. if the test involves a lot of LMA charms, then it should probably be a bundle test (which are to be run anyway as part of integration tests).

metadata.yaml
```diff
@@ -38,3 +38,4 @@ resources:
   prometheus-image:
     type: oci-image
     description: Container image for Prometheus
+    upstream-source: docker.io/ubuntu/prometheus@sha256:03936a740851b8786c328e08803a70feaa97731714a8dbe4b709c44366699b0c
```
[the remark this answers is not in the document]
balbirthomas: The source of container image is made a little more explicit using `docker.io`.

tests/integration/helpers.py
```diff
@@ -0,0 +1,146 @@
...
+    Returns:
+        a dictionary (possibly empty) representing the rule group
+    """
+    for group in rule_groups:
+        if app_name in group["name"]:
+            return group
+    return {}
+
+
+def oci_image(metadata_file: str, image_name: str) -> str:
```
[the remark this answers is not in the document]
balbirthomas: The function is provided as a generic utility for use with multiple different container images and metadata files. Further it provides exceptions with meaningful error messages indicating the precise cause of the value error.

tests/integration/conftest.py
```diff
@@ -0,0 +1,26 @@
...
+
+import pytest
+
+
+@pytest.fixture(scope="module", autouse=True)
+def copy_prometheus_library_into_tester_charm(ops_test):
+    """Ensure that the tester charm uses the current Prometheus library."""
+    library_path = "lib/charms/prometheus_k8s/v0/prometheus_scrape.py"
+    install_path = "tests/integration/prometheus-tester/" + library_path
+    shutil.copyfile(library_path, install_path)
```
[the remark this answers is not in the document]
balbirthomas: As mentioned above it is one less thing for test code to handle.

tests/integration/prometheus-tester/src/charm.py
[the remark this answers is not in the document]
balbirthomas: As explained above use of a test charm that can be instrumented in arbitrary ways provides a more controlled environment which facilitates writing more incisive and diverse range of tests. This I hope to demonstrate in future PRs.

tests/integration/test_charm.py
balbirthomas: How does this suggest it is a bug in the Prometheus charm code rather than a misinterpretation of the Python libjuju API or a bug there or in Juju or pytest-operator ?

tests/integration/workload.py
balbirthomas: As discussed in standup this will be explored later.

metadata.yaml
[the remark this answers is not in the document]
mmanciop: `docker.io` is the "default" registry, but I like the clarity of spelling it out.

tests/integration/test_charm.py
sed-i: It looks like there may be a bug in python-libjuju
https://github.com/juju/python-libjuju/issues/574

Regardless, I would move the `block_until` above `wait_for_idle`, in which case the python-libjuju bug should not have any effect.

tests/integration/conftest.py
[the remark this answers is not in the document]
simskij: I agree that setup and teardown of needed test resources (including folders) should be done by the test script. While I get your intention, @balbirthomas, and agree that less concern is usually better, we don't want to fill our CVS with artifacts either. But it's a minor, so it's totally fine if you prefer addressing it in a separate PR/issue.

mmanciop (approved):

tests/integration/test_charm.py
sed-i: btw, since this is an integration tests PR, imo it can be merged if the tests are good, even if they are failing due to a _charm_ bug.

tests/integration/test_charm.py
simskij: This **might** be an issue in case a test fails prior to reset being called. Instead, I would suggest adding a fixture with the `@pytest.fixture(autouse=True)` decorator, which will run the reset prior to every test case.

sed-i: To my understanding, "model reset" should not be used in an integration test: after all, it's an integration test: the app should be able to handle whatever happens in whatever sequence. I think that "reset" is used here instead of creating another integration test file (which has the unpleasant overhead of rebuilding and redeploying everything).

tests/integration/prometheus-tester/src/charm.py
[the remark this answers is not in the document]
simskij: #155 raised to address @sed-i's concerns

tests/integration/test_charm.py
simskij: Will be addressed by https://github.com/canonical/prometheus-operator/issues/154

tests/integration/helpers.py
```diff
@@ -0,0 +1,146 @@
...
+        app_name: string name of Prometheus application
+        unit_num: integer number of a Prometheus juju unit
+
+    Returns:
+        True if Prometheus is responsive else False
+    """
+    host = await unit_address(ops_test, app_name, unit_num)
+    prometheus = Prometheus(host=host)
+    is_ready = await prometheus.is_ready()
+    assert is_ready
```
simskij, line 37: In general, I'd say that asserts should be done in the test case, and that this function should return a value for the test case to assert. This usually turns out to be a good practice later, as tests evolve and grow more complex, as no function except the test case itself will fail a test, making it easier to maintain. 

**If** you agree, please create an issue for addressing it later, after your vacation.

simskij (approved):
Once you've finished creating issues for any follow-up work that need to happen based on the discussions in the PR, I'm totally fine with having this merged.

tests/integration/conftest.py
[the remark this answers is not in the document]
balbirthomas: OK. Will create a clean up routine in next PR.

tests/integration/test_charm.py
balbirthomas: I have pointed this out in response to other comments of yours. No charm is built more than once. The redeployment is intentional. The point is that each test can be executed by itself. Not only is it good practice that tests are independent of each other but also it allows executing just a single test if so desired, for example during development. The Model reset ensures that there are no lingering artifacts from a previous test, which can happen if the previous test failed and did not clean up properly.

tests/integration/test_charm.py
balbirthomas: This is done.

tests/integration/helpers.py
balbirthomas: Created issue with items to be picked up in next PR : https://github.com/canonical/prometheus-operator/issues/156

tests/integration/conftest.py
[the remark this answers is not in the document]
balbirthomas: https://github.com/canonical/prometheus-operator/issues/156

tests/integration/test_charm.py
balbirthomas: > This **might** become an issue in case a test fails prior to reset being called, but since you've marked all of them with `abort_on_fail` that will never make any difference. However, I would suggest instead adding a fixture with the `@pytest.fixture(autouse=True)` decorator, which will run the reset prior to every test case. No need to act on it in this PR though, but could be nice to do in a future refactoring.

OK. will explore if `autouse` can be used so that it is executed at end of test rather. You make a good point regarding `abort_on_fail`. I think this could be removed now.  I would like to remove it if you agree. Or perhaps we can choose to selectively mark tests as abort on fail, for example only if the first two tests (deployment check of two charms) fail then abort on fail. I can pick this up in next PR after discussion in plenary when I am back.

balbirthomas: Add to the TODO list ticket https://github.com/canonical/prometheus-operator/issues/156

PR #157, opened 2021-11-09
fix in-place dict update

sed-i:
This PR fixes a bug introduced in #149:
`rule.get("labels", {}).update(topology.as_dict_with_promql_labels())` won't update `rule` if `"labels"` is not in there - the new empty dict is not referenced by `rule`.

```python
>>> d = {}; d.get("labels", {}).update(k="v"); print(d)
{}
>>> if "labels" not in d:
...     d["labels"] = {}
... 
>>> d["labels"].update(k="v"); print(d)
{'labels': {'k': 'v'}}
>>> 
```


mmanciop (changes requested):
I think this deserves a unit test

sed-i:
> I think this deserves a unit test

How could I forget :D 
Added!

mmanciop (approved):

PR #161, opened 2021-11-11
add support for dual rules format

sed-i:
This PR proposes the following changes to add support for official format (`groups:` etc.) in addition to lma-format (`alerts:...` level only):
- [x] `class AlertRules` is where aggregating all rules files takes place (optional: recursively).
- [x] in the case of lma-format alerts, the filename is used to generate the group name.

### DIFF
--- a/lib/charms/prometheus_k8s/v0/prometheus_scrape.py
+++ b/lib/charms/prometheus_k8s/v0/prometheus_scrape.py
@@ -708,7 +708,7 @@ class AlertRules:
 
         return alert_groups if alert_groups else []
 
-    def add_from_path(self, path: str, *, recursive: bool = False):
+    def add_path(self, path: str, *, recursive: bool = False):
         """Add rules from a dir path.
 
         All rules from files are aggregated into a data structure representing a single rule file.
@@ -720,9 +720,6 @@ class AlertRules:
 
         Raises:
             NotADirectoryError: if the provided path is invalid.
-
-        Returns:
-            self, so calls are chainable.
         """
         path = Path(path)  # type: Path
         if path.is_dir():
@@ -731,8 +728,6 @@ class AlertRules:
             self.alert_groups.extend(self._from_file(path.parent, path))
         else:
             raise InvalidAlertRulePathError(path, "path does not exist")
-
-        return self  # so calls are chainable
 
     def as_dict(self) -> dict:
         """Return standard alert rules file in dict representation."""
@@ -1308,22 +1303,20 @@ class MetricsEndpointProvider(Object):
         if not self._charm.unit.is_leader():
             return
 
-        rules_file = (
-            AlertRules(self.topology)
-            .add_from_path(self._alert_rules_path, recursive=False)
-            .as_dict()
-        )
+        alert_rules = AlertRules(self.topology)
+        alert_rules.add_path(self._alert_rules_path, recursive=False)
+        alert_rules_as_dict = alert_rules.as_dict()
 
         for relation in self._charm.model.relations[self._relation_name]:
             relation.data[self._charm.app]["scrape_metadata"] = json.dumps(self._scrape_metadata)
             relation.data[self._charm.app]["scrape_jobs"] = json.dumps(self._scrape_jobs)
 
-            if rules_file:
+            if alert_rules_as_dict:
                 # Update relation data with the string representation of the rule file.
                 # Juju topology is already included in the "scrape_metadata" field above.
                 # The consumer side of the relation uses this information to name the rules file
                 # that is written to the filesystem.
-                relation.data[self._charm.app]["alert_rules"] = json.dumps(rules_file)
+                relation.data[self._charm.app]["alert_rules"] = json.dumps(alert_rules_as_dict)
 
     def _set_unit_ip(self, _):
         """Set unit host address.
@@ -1412,17 +1405,15 @@ class RuleFilesProvider(Object):
         if not self._charm.unit.is_leader():
             return
 
-        rules_file = (
-            AlertRules(self.topology)
-            .add_from_path(self.dir_path, recursive=self._recursive)
-            .as_dict()
-        )
+        alert_rules = AlertRules(self.topology)
+        alert_rules.add_path(self.dir_path, recursive=self._recursive)
+        alert_rules_as_dict = alert_rules.as_dict()
 
-        if rules_file:
+        if alert_rules_as_dict:
             logger.info("Updating relation data with rule files from disk")
             for relation in self._charm.model.relations[self._relation_name]:
                 relation.data[self._charm.app]["alert_rules"] = json.dumps(
-                    rules_file,
+                    alert_rules_as_dict,
                     sort_keys=True,  # sort, to prevent unnecessary relation_changed events
                 )
 
--- a/lib/charms/prometheus_k8s/v0/prometheus_scrape.py
+++ b/lib/charms/prometheus_k8s/v0/prometheus_scrape.py
@@ -496,7 +496,7 @@ class JujuTopology:
         )
 
     @staticmethod
-    def from_relation_data(data):
+    def from_relation_data(data: dict):
         """Factory method for creating the topology dataclass from a relation data dict."""
         return JujuTopology(
             model=data["model"],
@@ -565,14 +565,14 @@ class InvalidAlertRulePathError(Exception):
 
 
 class AlertRules:
-    """Data store for prometheus alert rules.
+    """Utility class for amalgamating prometheus alert rule files and injecting juju topology.
 
-    Alert rule files can be read in both the official and the LMA format. All rules are combined in
-    a single data structure representing an amalgamated rules file.
+    Alert rule files can be read in both the official and the custom format (one alert per file).
+    All rules are combined in a single data structure representing an amalgamated rules file.
 
     The official Prometheus format is a YAML file conforming to the Prometheus documentation
     (https://prometheus.io/docs/prometheus/latest/configuration/alerting_rules/).
-    The LMA format is a subsection of the official YAML, having a single alert rule, effectively
+    The custom format is a subsection of the official YAML, having a single alert rule, effectively
     "one alert per file". This class supports either format.
 
     For the purpose of forwarding alert rules over relation data, all rules are aggregated into a
@@ -596,40 +596,43 @@ class AlertRules:
         self.alert_groups = []  # type: List[dict]
 
     @staticmethod
-    def _is_valid_official_format(rules_dict: dict):
+    def _is_official_format(rules_dict: dict) -> bool:
+        # TODO validate e.g. against schema using pydantic
         return "groups" in rules_dict
 
     @staticmethod
-    def _is_valid_lma_format(rules_dict: dict):
+    def _is_custom_format(rules_dict: dict) -> bool:
         # one alert rule per file
+        # TODO validate e.g. against schema using pydantic
         return set(rules_dict) >= {"alert", "expr"}
 
-    @staticmethod
-    def _from_file(root_path: str, file_path: Path, topology: JujuTopology) -> List[dict]:
+    def _from_file(self, root_path: Path, file_path: Path) -> List[dict]:
         """Read a rules file from path, injecting juju topology.
 
         Args:
-            root_path: path to the root rules folder (used only for generating group name)
-            file_path: path to a *.rule file.
-            topology: juju topology information.
+            root_path: full path to the root rules folder (used only for generating group name)
+            file_path: full path to a *.rule file.
+
+        Returns:
+            A list of dictionaries representing the rules file, if file is valid (the structure is
+            formed by `yaml.safe_load` of the file); an empty list otherwise.
         """
-        with file_path.open() as rule_file:
+        with file_path.open() as rf:
             # Load a list of rules from file then add labels and filters
             try:
-                from_yaml = yaml.safe_load(rule_file)
+                rule_file = yaml.safe_load(rf)
+
             except Exception as e:
                 logger.error("Failed to read alert rules from %s: %s", file_path.name, e)
                 return []
 
             # TODO validate input
-            if AlertRules._is_valid_official_format(from_yaml):
-                # looks like standard rules file
-                alert_groups = from_yaml["groups"]
-            elif AlertRules._is_valid_lma_format(from_yaml):
-                # looks like LMA format (one alert rule per file)
+            if AlertRules._is_official_format(rule_file):
+                alert_groups = rule_file["groups"]
+            elif AlertRules._is_custom_format(rule_file):
                 # convert to list of alert groups
                 # group name is made up from the file name
-                alert_groups = [{"name": file_path.stem, "rules": [from_yaml]}]
+                alert_groups = [{"name": file_path.stem, "rules": [rule_file]}]
             else:
                 # invalid/unsupported
                 logger.error("Invalid rules file: %s", file_path.name)
@@ -638,10 +641,9 @@ class AlertRules:
             # update rules with additional metadata
             for alert_group in alert_groups:
                 # update group name with topology and sub-path
-                alert_group["name"] = AlertRules._group_name(
-                    root_path,
+                alert_group["name"] = self._group_name(
+                    str(root_path),
                     str(file_path),
-                    topology,
                     alert_group["name"],
                 )
 
@@ -649,17 +651,14 @@ class AlertRules:
                 for alert_rule in alert_group["rules"]:
                     if "labels" not in alert_rule:
                         alert_rule["labels"] = {}
-                    alert_rule["labels"].update(topology.as_dict_with_promql_labels())
+                    alert_rule["labels"].update(self.topology.as_dict_with_promql_labels())
 
                     # insert juju topology filters into a prometheus alert rule
-                    alert_rule["expr"] = topology.render(alert_rule["expr"])
+                    alert_rule["expr"] = self.topology.render(alert_rule["expr"])
 
             return alert_groups
 
-    @staticmethod
-    def _group_name(
-        root_path: str, file_path: str, topology: JujuTopology, group_name: str
-    ) -> str:
+    def _group_name(self, root_path: str, file_path: str, group_name: str) -> str:
         """Generate group name from path and topology.
 
         The group name is made up of the relative path between the root dir_path, the file path,
@@ -668,8 +667,10 @@ class AlertRules:
         Args:
             root_path: path to the root rules dir.
             file_path: path to rule file.
-            topology: juju topology information.
             group_name: original group name to keep as part of the new augmented group name
+
+        Returns:
+            New group name, augmented by juju topology and relative path.
         """
         rel_path = os.path.relpath(os.path.dirname(file_path), root_path)
         rel_path = "" if rel_path == "." else rel_path.replace(os.path.sep, "_")
@@ -677,12 +678,11 @@ class AlertRules:
         # Generate group name:
         #  - name, from juju topology
         #  - suffix, from the relative path of the rule file;
+        group_name_parts = [self.topology.identifier, rel_path, group_name, "alerts"]
+        return "_".join(filter(None, group_name_parts))
 
-        return "{}_{}_{}_alerts".format(topology.identifier, rel_path, group_name)
-
-    @staticmethod
-    def _from_dir(dir_path: str, topology: JujuTopology, recursive: bool) -> List[dict]:
-        """Factory method for creating an instance from a path to a folder of rules.
+    def _from_dir(self, dir_path: Path, recursive: bool) -> List[dict]:
+        """Read all rule files in a directory.
 
         All rules from files for the same directory are loaded into a single
         group. The generated name of this group includes juju topology.
@@ -690,20 +690,20 @@ class AlertRules:
 
         Args:
             dir_path: directory containing *.rule files (alert rules without groups).
-            topology: juju topology information.
             recursive: flag indicating whether to scan for rule files recursively.
 
         Returns:
-            a list of prometheus alert rule groups.
+            a list of dictionaries representing prometheus alert rule groups, each dictionary
+            representing an alert group (structure determined by `yaml.safe_load`).
         """
         alert_groups = []  # type: List[dict]
 
         # Gather all alerts into a list of groups
-        paths = Path(dir_path).glob("**/*.rule" if recursive else "*.rule")
+        paths = dir_path.glob("**/*.rule" if recursive else "*.rule")
         for file_path in filter(Path.is_file, paths):
-            alert_groups_from_file = AlertRules._from_file(dir_path, file_path, topology)
+            alert_groups_from_file = self._from_file(dir_path, file_path)
             if alert_groups_from_file:
-                print("Reading alert rule from %s", file_path)
+                logger.debug("Reading alert rule from %s", file_path)
                 alert_groups.extend(alert_groups_from_file)
 
         return alert_groups if alert_groups else []
@@ -716,20 +716,26 @@ class AlertRules:
 
         Args:
             path: either a rules file or a dir of rules files.
-            recursive: whether to read files recursively or not.
+            recursive: whether to read files recursively or not (no impact if `path` is a file).
 
         Raises:
             NotADirectoryError: if the provided path is invalid.
+
+        Returns:
+            self, so calls are chainable.
         """
-        if Path(path).is_dir():
-            self.alert_groups.extend(AlertRules._from_dir(path, self.topology, recursive))
+        path = Path(path)  # type: Path
+        if path.is_dir():
+            self.alert_groups.extend(self._from_dir(path, recursive))
+        elif path.is_file():
+            self.alert_groups.extend(self._from_file(path.parent, path))
         else:
-            raise NotADirectoryError("The provided path does not exist: {}".format(path))
+            raise InvalidAlertRulePathError(path, "path does not exist")
 
         return self  # so calls are chainable
 
-    def as_rules_file_dict(self) -> dict:
-        """Return standard alert rules file in string representation."""
+    def as_dict(self) -> dict:
+        """Return standard alert rules file in dict representation."""
         return {"groups": self.alert_groups} if self.alert_groups else {}
 
 
@@ -859,28 +865,33 @@ class MetricsEndpointConsumer(Object):
 
         {"groups": [{"name": ...}, ...]}
 
+        per official prometheus documentation
+        https://prometheus.io/docs/prometheus/latest/configuration/alerting_rules/
+
         The value of the `groups` key is such that it may be used to generate
         a Prometheus alert rules file directly using `yaml.dump` but the
         `groups` key itself must be included as this is required by Prometheus.
 
         Returns:
-            A dictionary mapping the juju identifier of the source charm to its rules file.
+            A dictionary mapping the juju identifier of the source charm to its alert rules file.
         """
         alerts = {}  # type: Dict[str, dict] # mapping b/w juju identifiers and alert rule files
         for relation in self._charm.model.relations[self._relation_name]:
             if not relation.units:
                 continue
 
-            alert_rules = json.loads(relation.data[relation.app].get("alert_rules", "[]"))
+            alert_rules = json.loads(relation.data[relation.app].get("alert_rules", "{}"))
             if not alert_rules:
                 continue
 
             try:
-                topology_identifier, rules_file = alert_rules
-                # TODO validate rules_file
-                alerts[topology_identifier] = rules_file
+                scrape_metadata = json.loads(relation.data[relation.app]["scrape_metadata"])
+                topology = JujuTopology.from_relation_data(scrape_metadata)
 
-            except ValueError as e:
+                # TODO validate rules_file
+                alerts[topology.identifier] = alert_rules
+
+            except KeyError as e:
                 logger.error(
                     "Relation %s has invalid data : %s",
                     relation.id,
@@ -1300,7 +1311,7 @@ class MetricsEndpointProvider(Object):
         rules_file = (
             AlertRules(self.topology)
             .add_from_path(self._alert_rules_path, recursive=False)
-            .as_rules_file_dict()
+            .as_dict()
         )
 
         for relation in self._charm.model.relations[self._relation_name]:
@@ -1309,13 +1320,10 @@ class MetricsEndpointProvider(Object):
 
             if rules_file:
                 # Update relation data with the string representation of the rule file.
-                # Juju topology is included because the consumer side of the relation uses this
-                # information to name the rules file that is written to the filesystem.
-                # Although the topology identifier is included in all group names, it cannot be
-                # easily retrieved from there.
-                relation.data[self._charm.app]["alert_rules"] = json.dumps(
-                    [self.topology.identifier, rules_file]
-                )
+                # Juju topology is already included in the "scrape_metadata" field above.
+                # The consumer side of the relation uses this information to name the rules file
+                # that is written to the filesystem.
+                relation.data[self._charm.app]["alert_rules"] = json.dumps(rules_file)
 
     def _set_unit_ip(self, _):
         """Set unit host address.
@@ -1407,19 +1415,18 @@ class RuleFilesProvider(Object):
         rules_file = (
             AlertRules(self.topology)
             .add_from_path(self.dir_path, recursive=self._recursive)
-            .as_rules_file_dict()
+            .as_dict()
         )
 
         if rules_file:
             logger.info("Updating relation data with rule files from disk")
             for relation in self._charm.model.relations[self._relation_name]:
                 relation.data[self._charm.app]["alert_rules"] = json.dumps(
-                    [self.topology.identifier, rules_file],
+                    rules_file,
                     sort_keys=True,  # sort, to prevent unnecessary relation_changed events
                 )
 
 
-# TODO: update this class with changes made to relation data format
 class MetricsEndpointAggregator(Object):
     """Aggregate metrics from multiple scrape targets.
 
@@ -1532,12 +1539,14 @@ class MetricsEndpointAggregator(Object):
         """
         jobs = []  # list of scrape jobs, one per relation
         for relation in self.model.relations[self._target_relation]:
-            if targets := self._get_targets(relation):
+            targets = self._get_targets(relation)
+            if targets:
                 jobs.append(self._static_scrape_job(targets, relation.app.name))
 
         groups = []  # list of alert rule groups, one group per relation
         for relation in self.model.relations[self._alert_rules_relation]:
-            if unit_rules := self._get_alert_rules(relation):
+            unit_rules = self._get_alert_rules(relation)
+            if unit_rules:
                 appname = relation.app.name
                 rules = self._label_alert_rules(unit_rules, appname)
                 group = {"name": self._group_name(appname), "rules": rules}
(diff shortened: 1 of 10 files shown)


balbirthomas:
The requirements for this PR which seems to have started with [issue 153](https://github.com/canonical/prometheus-operator/issues/153) are not clear to me.

Could you please clarify :

1. Is it the requirement that recording and alerting rules should be combined into the same charm folder and use the same code flow for processing both ? Note the above discussion in [issue 153](https://github.com/canonical/prometheus-operator/issues/153) is making that case for supporting both ".rule" and ".rules" by mixing examples of both recording and alerting rules. At present the default folder name for "alert rules" is "prometheus_alert_rules" ! More importantly as far I can tell from the Prometheus alerting rules docs there are no groups specific parameters for alerting rules (such as shown in the example above for recording rules). The only thing an alerting rule group has a "group name" and a "list of rules" or am I missing something here ?

2. Group names using Juju topology are fairly unambiguous. What is the use case for one charm submitting multiple alert rule groups (the only difference between them being a different name) ?

3. In fact the Prometheus alert rules file can contain multiple groups in a single file. If the desire is to support the "full prometheus format" . Shouldn't we be supporting multiple alert rule groups in a single file ? Or is there some reason not to do this ? Do charms need to support the "full Prometheus format" or can we effect some simplification to the benefit of the system admins and charm writers ?

4.  What is the use case for recursive rules folders ?

sed-i:
## 1.

### (a)
> Is it the requirement that recording and alerting rules should be combined into the same charm folder

I think the charm lib should be able to pick up whatever tree structure the user devised. As you can see from the code, it is quite straightforward supporting arbitrary rules folder structure.

### (b)
> The only thing an alerting rule group has ...

True, but being future proof here is quite trivial.

### (c)
> At present the default folder name for "alert rules" is "prometheus_alert_rules"

We can keep it that way (or change it to `rules/prom` and `rules/loki`, doesn't matter). If the user ends up dumping there their entire collection, which may have recording rules as well, the lib should still be able to process I believe. Unless the kinds could be indistinguishable?

## 2.
> What is the use case for one charm submitting multiple alert rule groups

In case the user already has a collection of rule files, we want to be able to read them without any intervention by the user. Prom's official format says the user can do it, so we might as well assume the user did it.

## 3.
### (a)
> If the desire is to support the "full prometheus format"

That is my exact intention. As far as I can tell the current code supports that.

### (b)
> Do charms need to support the "full Prometheus format"

In case the user already has a collection of rule files, we want to be able to read them without any intervention by the user. Prom's official format says the user can do it, so we might as well assume the user did it.

## 4.
> What is the use case for recursive rules folders

An internal client has this in production.

balbirthomas:
## 1.
### (a)
> 
> > Is it the requirement that recording and alerting rules should be combined into the same charm folder
> 
> I think the charm lib should be able to pick up whatever tree structure the user devised. As you can see from the code, it is quite straightforward supporting arbitrary rules folder structure.

The question was **why** do we need to support reading of both recording rules **and** alerting rules **in the same folder** ? In other words why shouldn't we be more prescriptive here ?

Since you mentioned it, perhaps you may want to clarify **why** do we need to support arbitrary folder structure (instead of being prescriptive here) ? What do you mean by arbitrary - arbitrary nesting, arbitrary combination of different types of rules (alerting and recording) in arbitrarily nested folders with arbitrary folder names ?

### (b)

> True, but being future proof here is quite trivial.

The question is what benefit does this flexibility offer to the end user ? The future in this case happens to exactly what we make it to be. By not being prescriptive enough we are actually jeopardising that future by encouraging the creation of an ecosystem of charms that adopt arbitrary ways of doing things. This may not seem to be a problem now because it is "trivial" to implement such a feature **now** but will it be "trivial" to **continue supporting** arbitrary folder structures for all future use cases and for the rest of the lifetime of this charm ?

### (c)

> We can keep it that way (or change it to `rules/prom` and `rules/loki`, doesn't matter). If the user ends up dumping there their entire collection, which may have recording rules as well, the lib should still be able to process I believe. 

This is really a product decision and perhaps this discussion must be directed to the product specification and perhaps amend it accordingly. Also see 1(b) and 5 in this post regarding other concerns about trying to anticipate "If the user ends up dumping there their entire collection ...."

> Unless the kinds could be indistinguishable? 

So are you proposing that we should **now** allow arbitrary folder structure that allows combining recording rules and alert rules into a single folder and **later** bother about how we will distinguish them ?

## 2.
and
## 3.
### (b)

> In case the user already has a collection of rule files, we want to be able to read them without any intervention by the user. Prom's official format says the user can do it, so we might as well assume the user did it.

Which user are you referring to (Openstack ? ) ? As far as I am aware there are very few users for a Prometheus charm as yet.  

> Prom's official format says the user can do it, so we might as well assume the user did it.

It is not clear what the user can not do and what may we assume the user did.

I presume you mean the user has a collection of rule files in a git repository, correct ?  If so why can this not be done within the Alert Rules Charm ? I mean why can't the Alert rules charm support reading the rules from the git repository without "intervention by the user" and marshal it into the structure that the Prometheus charm expects . (also see 4 below)

## 4.

> > What is the use case for recursive rules folders
> 
> An internal client has this in production.

Which Internal client (Openstack) ? If so why is this use case not handled by the proxy or alert rules charm ? Since this is a use case specific to one client why should the Prometheus charm and its library which is also intended for wider distribution constrain its design based on the use cases of a single client ?

## 5
Seems like this PR is trying to anticipate product requirements and "future proof" the charm. Should we really be trying to anticipate the product requirements here ? Isn't such anticipation a common cause of over engineering that leads to convoluted code when the actual product requirements are discovered much later and just turn out to be different from what was anticipated ? Doesn't it makes more sense to provide the minimal feature set that exactly meets the product requirements for a first release and carefully grow the feature set precisely as new use cases arise and product requirements grow.

lib/charms/prometheus_k8s/v0/prometheus_scrape.py
```diff
@@ -297,9 +297,8 @@ def _on_scrape_targets_changed(self, event):
 import json
```
balbirthomas, line 297: Kindly note, I think this PR should be associated with an update to the "Code Overview" section of the Contributing.md docs. There is already a pending Github issue for such and update from your previous PR. I think the best way to keep docs updated is to include them in the PR that updates the code.

balbirthomas (changes requested):

lib/charms/prometheus_k8s/v0/prometheus_scrape.py
```diff
@@ -480,6 +479,8 @@ def _sanitize_scrape_configuration(job) -> dict:
 class JujuTopology:
     """Dataclass for storing and formatting juju topology information."""
 
+    VARIABLE: ClassVar = "%%juju_topology%%"
```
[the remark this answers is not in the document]
sed-i: Changed to STUB.
Removed variable type hint for py3.5.
#166 would obviate the %% stub, I think.

tests/unit/test_endpoint_provider.py
```diff
@@ -317,94 +309,84 @@ def expression_labels(expr):
...
+                    "rules": [expected_freestanding_rule],
+                },
+            ]
+        }
 
-    def test_group_name_prefixed_by_subdir_name(self):
-        expected_group_name = "nested_rules_dir_" + self.topology.identifier + "_alerts"
-        nested = list(filter(lambda group: expected_group_name == group["name"], self.rule_groups))
-        self.assertGreaterEqual(len(nested), 1)
+        self.assertEqual({}, DeepDiff(expected_rules_file, rules_file_dict, ignore_order=True))
```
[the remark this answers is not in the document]
sed-i: `assertDictEqual` is not an order-agnostic deepdiff. The following fails:
```python
self.assertDictEqual({1:[1,2]}, {1:[2,1]})
```

tests/unit/test_endpoint_provider.py
```diff
@@ -317,94 +309,84 @@ def expression_labels(expr):
...
-    #                              'summary': 'Prometheus target missing (instance '
-    #                                         '{{ $labels.instance }})'},
-    #              'expr': 'up{juju_model="MyModel", juju_model_uuid="MyUUID", '
-    #                      'juju_application="MyApp"} == 0',
-    #              'for': '0m',
-    #              'labels': {'juju_application': 'MyApp',
-    #                         'juju_model': 'MyModel',
-    #                         'juju_model_uuid': 'MyUUID',
-    #                         'severity': 'critical'}}]}]
+class Sandbox:
```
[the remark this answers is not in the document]
sed-i: Thanks! Docs added.

tox.ini
```diff
@@ -82,8 +83,8 @@ lma_bundle_dir = {envtmpdir}/lma-light-bundle
 deps =
     # deps from lma-bundle - these are needed here because will be running pytest on lma-bundle
     jinja2
-    #git+https://github.com/juju/python-libjuju.git
-    juju
+    git+https://github.com/juju/python-libjuju.git
```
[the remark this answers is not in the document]
sed-i: The diff will go away as soon as I rebase.

sed-i (commented):
The next commit addresses the easy fixes.
A refactoring of AlertRulesAggregator will follow.

tests/unit/test_endpoint_provider.py
```diff
@@ -317,94 +309,84 @@ def expression_labels(expr):
...
 
-    def test_at_least_one_group_per_file(self):
-        self.assertGreater(len(self.rule_groups), 1)
+    def test_non_recursive_is_default(self):
+        rules_file_dict = AlertRulesAggregator.from_path(
+            os.path.join(self.sandbox.root, "rules", "prom"), topology=self.topology
+        ).as_rules_file_dict()
+        self.assertEqual({}, rules_file_dict)
+
+    def test_alerts_in_both_formats_are_recursively_aggregated(self):
```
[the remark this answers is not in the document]
sed-i: - Added more tests (which found bugs :D thanks!).
- The line `self.assertEqual({}, DeepDiff(expected_rules_file, rules_file_dict, ignore_order=True))` is a "holistic structure validation" in lieu of a schema to test against.
- I am not in favor of expressions such as `assert "expr" in dict["groups"][0]["rules"][0]`.

lib/charms/prometheus_k8s/v0/prometheus_scrape.py
```diff
@@ -545,84 +546,203 @@ def as_dict_with_promql_labels(self):
...
+
+        Returns:
+            self, so calls are chainable.
+        """
+        if Path(path).is_dir():
+            self.alert_groups.extend(AlertRules._from_dir(path, self.topology, recursive))
+        else:
+            raise NotADirectoryError("The provided path does not exist: {}".format(path))
+
+        return self  # so calls are chainable
```
Abuelodelanada, line 741: If it's possible I would use guard clauses

```python
       if not Path(path).is_dir():
            raise NotADirectoryError("The provided path does not exist: {}".format(path))
            
        self.alert_groups.extend(AlertRules._from_dir(path, self.topology, recursive))    
        return self # so calls are chainable        
```

lib/charms/prometheus_k8s/v0/prometheus_scrape.py
```diff
@@ -541,85 +542,182 @@ def as_dict_with_promql_labels(self):
...
+                        if "labels" not in alert_rule:
+                            alert_rule["labels"] = {}
+                        alert_rule["labels"].update(topology.as_dict_with_promql_labels())
+
+                        # insert juju topology filters into a prometheus alert rule
+                        alert_rule["expr"] = topology.render(alert_rule["expr"])
+
+                return alert_groups
+
+    @staticmethod
```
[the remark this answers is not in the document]
balbirthomas: Here is one for example https://www.webucator.com/article/when-to-use-static-methods-in-python-never/.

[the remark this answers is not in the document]
balbirthomas: A quite by Guido Van Rossum from that article 

> We all know how limited static methods are. (They’re basically an accident — back in the Python 2.2 days when I was inventing new-style classes and descriptors, I meant to implement class methods but at first I didn’t understand them and accidentally implemented static methods first. Then it was too late to remove them and only provide class methods.

lib/charms/prometheus_k8s/v0/prometheus_scrape.py
```diff
@@ -545,84 +546,203 @@ def as_dict_with_promql_labels(self):
...
-            # insert juju topology filters into a prometheus alert rule
-            rule["expr"] = topology.render(rule["expr"])
-            return rule
+    def __init__(self, topology: JujuTopology):
+        self.topology = topology
+        self.alert_groups = []  # type: List[dict]
 
+    @staticmethod
+    def _is_valid_official_format(rules_dict: dict) -> bool:
+        # TODO validate against schema using e.g. pydantic
```
balbirthomas, line 600: Promtool already provides validation of alert rules. I am not sure if we should reimplement this functionality using pydantic. Besides schema validation for relation data is a planned feature in OF.  I do not think we should be using a TODO comment to suggest a solution that has not been discussed, agreed or spec'd out in the team or in a github ticket. Kindly create a github ticket where this can be discussed and alternatives considered. Even the TODO is probably not necessary.

lib/charms/prometheus_k8s/v0/prometheus_scrape.py
```diff
@@ -545,84 +546,203 @@ def as_dict_with_promql_labels(self):
...
+        self.alert_rules_absolute_path = alert_rules_absolute_path
+        self.message = message
 
-def load_alert_rule_from_file(path: Path, topology: JujuTopology) -> Optional[dict]:
-    """Load alert rule from a rules file.
+        super().__init__(self.message)
+
+
+class AlertRules:
+    """Data store for prometheus alert rules.
```
balbirthomas, line 568: Looking at the methods defined in this class (even though they are static methods) the responsibility of this class seems to be
1. Read alert rules from a directory
2. Ensure alert rules have topology labels
Hence it I am not sure if it correct to say the responsibility of this class is just to store alert rules.

lib/charms/prometheus_k8s/v0/prometheus_scrape.py
```diff
@@ -545,84 +546,203 @@ def as_dict_with_promql_labels(self):
...
+    def _is_valid_official_format(rules_dict: dict) -> bool:
+        # TODO validate against schema using e.g. pydantic
+        return "groups" in rules_dict
 
-def load_alert_rules_from_dir(
-    dir_path: Union[str, Path], topology: JujuTopology, *, recursive: bool = False
-) -> List[dict]:
-    """Load alert rules from rule files.
+    @staticmethod
+    def _is_valid_lma_format(rules_dict: dict) -> bool:
```
balbirthomas, line 604: Since this method currently is only trying to determine if the format is "lma format" wouldn't it be more appropriate to call it `_is_in_lma_format()` ? It is a private method after all and in the future if validation is added then the method name may be changed then or if required a new method created. 

(same question for `_is_valid_official_format()`.

mmanciop: Since we are having a discussion about whether we should have a different Product Name than "LMA", I do not find it appropriate to add to our code in an API :-0

lib/charms/prometheus_k8s/v0/prometheus_scrape.py
```diff
@@ -545,84 +546,203 @@ def as_dict_with_promql_labels(self):
...
+                logger.error("Failed to read alert rules from %s: %s", file_path.name, e)
+                return []
+
+            # TODO validate input
+            if AlertRules._is_valid_official_format(rule_file):
+                alert_groups = rule_file["groups"]
+            elif AlertRules._is_valid_lma_format(rule_file):
+                # convert to list of alert groups
+                # group name is made up from the file name
+                alert_groups = [{"name": file_path.stem, "rules": [rule_file]}]
```
balbirthomas, line 637: Am I correct in understanding this means for rules in LMA format each rule is put into a group of its own ? If so is this as per current product specification ? 
Note currently all rules in LMA format are put into a common group.

lib/charms/prometheus_k8s/v0/prometheus_scrape.py
```diff
@@ -545,84 +546,203 @@ def as_dict_with_promql_labels(self):
...
+                return []
+
+            # update rules with additional metadata
+            for alert_group in alert_groups:
+                # update group name with topology and sub-path
+                alert_group["name"] = AlertRules._group_name(
+                    root_path,
+                    str(file_path),
+                    topology,
+                    alert_group["name"],
```
balbirthomas, line 650: It is bit of a mind bender to see a method called `_group_name()` that accepts an `alert_group["name"]` and the return value is reassigned to `alert_group["name"]`. Wouldn't it be more intention revealing if the class structure and method names provide an API such as `self.rename_group()` or `self.rename_groups()` ?

lib/charms/prometheus_k8s/v0/prometheus_scrape.py
```diff
@@ -545,84 +546,203 @@ def as_dict_with_promql_labels(self):
...
+
+                    # insert juju topology filters into a prometheus alert rule
+                    alert_rule["expr"] = topology.render(alert_rule["expr"])
 
-    def _group_name(path) -> str:
+            return alert_groups
+
+    @staticmethod
+    def _group_name(
+        root_path: str, file_path: str, topology: JujuTopology, group_name: str
```
balbirthomas, line 666: `topology` is already a member of the `AlertRules` class. If this method was not static it would not need an additional topology argument thereby simplifying its function signature. It would perhaps be more readable to have a `rename_group` method that did not require the `group_name` argument either. Instead the `rename_group` method could read the group name from a alert rule group, change its name and write it back.

lib/charms/prometheus_k8s/v0/prometheus_scrape.py
```diff
@@ -545,84 +546,203 @@ def as_dict_with_promql_labels(self):
...
-        if rule := load_alert_rule_from_file(path, topology):
-            logger.debug("Reading alert rule from %s", path)
-            alerts[_group_name(path)].append(rule)
+        return alert_groups if alert_groups else []
 
-    # Gather all alerts into a list of groups since Prometheus
-    # requires alerts be part of some group
-    groups = [{"name": k, "rules": v} for k, v in alerts.items()]
-    return groups
+    def add_from_path(self, path: str, *, recursive: bool = False):
```
balbirthomas, line 720: It is not clear why this method is required. All this method is doing is validating its argument and invoking `_from_dir()`. It seems odd to have two methods for reading all alert rules from a directory with the only difference being - One is and instance method that invoke the other which is a private static method. Why not have just one public instance method that reads all alert rules from the directory. Another possibility is to make the directory path a constructor argument to `AlertRules` object which then only needs to provide methods to return the alert rules in different formats.

lib/charms/prometheus_k8s/v0/prometheus_scrape.py
```diff
@@ -545,84 +546,203 @@ def as_dict_with_promql_labels(self):
...
+        Raises:
+            NotADirectoryError: if the provided path is invalid.
+
+        Returns:
+            self, so calls are chainable.
+        """
+        if Path(path).is_dir():
+            self.alert_groups.extend(AlertRules._from_dir(path, self.topology, recursive))
+        else:
+            raise NotADirectoryError("The provided path does not exist: {}".format(path))
```
balbirthomas, line 739: Does this need a custom exception ? Would it be reasonable to use `ValueError` instead ?

lib/charms/prometheus_k8s/v0/prometheus_scrape.py
```diff
@@ -742,40 +840,34 @@ def alerts(self) -> dict:
...
+            alert_rules = json.loads(relation.data[relation.app].get("alert_rules", "[]"))
             if not alert_rules:
                 continue
 
             try:
-                for group in alert_rules["groups"]:
-                    alerts[group["name"]] = group
-            except KeyError as e:
+                topology_identifier, rules_file = alert_rules
+                # TODO validate rules_file
```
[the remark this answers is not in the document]
balbirthomas: Is there a Github ticket or product spec to discuss the requirements around validation of alert rules ? Please note Prometheus provides [a tool](https://www.prometheus.io/docs/prometheus/latest/configuration/unit_testing_rules/) to test alert rules.

lib/charms/prometheus_k8s/v0/prometheus_scrape.py
```diff
@@ -1320,17 +1418,22 @@ def _update_relation_data(self, _):
         if not self._charm.unit.is_leader():
             return
 
-        if alert_groups := load_alert_rules_from_dir(
-            self.dir_path, self.topology, recursive=self._recursive
-        ):
+        rules_file = (
+            AlertRules(self.topology)
+            .add_from_path(self.dir_path, recursive=self._recursive)
+            .as_dict()
```
balbirthomas, line 1424: 1. It is not very intuitive to see and `AlertRules` object being instantiated with a `topology` argument and then the rules themselves being added using a method. It is not clear why the `AlertRules` can not be instantiated with both a alert rules path and the topology.
2. It is not very intuitive to see an `AlertRules` converted to a `dict` and assigned to a variable called `rules_file` instead of just something like `alert_rules`. A `file` suffix is suggests that this is a file object.

lib/charms/prometheus_k8s/v0/prometheus_scrape.py
```diff
@@ -1320,17 +1418,22 @@ def _update_relation_data(self, _):
...
+        )
+
+        if rules_file:
             logger.info("Updating relation data with rule files from disk")
             for relation in self._charm.model.relations[self._relation_name]:
                 relation.data[self._charm.app]["alert_rules"] = json.dumps(
-                    {"groups": alert_groups},
+                    [self.topology.identifier, rules_file],
```
balbirthomas, line 1431: It is not clear why topology information needs to be provided here when it already available in the scrape metadata.

lib/charms/prometheus_k8s/v0/prometheus_scrape.py
sed-i: Docstrings:
- NotADirectoryError: Operation only works on directories.
- ValueError: Inappropriate argument value (of correct type).
- FileNotFoundError: File not found.

Seems to me that NotADirectoryError is better than a custom type or ValueError.

lib/charms/prometheus_k8s/v0/prometheus_scrape.py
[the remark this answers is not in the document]
sed-i: Captured in #170

lib/charms/prometheus_k8s/v0/prometheus_scrape.py
sed-i: That approach was already reverted in the previous commit, but I overlooked this line. Thanks!

lib/charms/prometheus_k8s/v0/prometheus_scrape.py
```diff
@@ -1317,17 +1389,18 @@ def _update_relation_data(self, _):
...
                     sort_keys=True,  # sort, to prevent unnecessary relation_changed events
                 )
 
 
+# TODO: update this class with changes made to relation data format
```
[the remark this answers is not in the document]
sed-i: Captured in #171

lib/charms/prometheus_k8s/v0/prometheus_scrape.py
```diff
@@ -541,85 +542,182 @@ def as_dict_with_promql_labels(self):
...
+        the "alert" and "expr" keys.
+      - alert rule (singular): a single dictionary that has the "alert" and "expr" keys.
     """
-    alerts = defaultdict(list)
 
-    def _group_name(path) -> str:
+    def __init__(self, alert_groups: List[dict] = None):
+        self.alert_groups = alert_groups if alert_groups else []
+
+    @staticmethod
```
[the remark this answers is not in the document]
sed-i: Converted to instance method. Adding `self` and dropping `topology` seems like a better choice.

lib/charms/prometheus_k8s/v0/prometheus_scrape.py
sed-i: Correct, for the custom format ("LMA format"), group name is made up from the filename, per (current) spec.

lib/charms/prometheus_k8s/v0/prometheus_scrape.py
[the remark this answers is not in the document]
sed-i: Converted to instance method. Adding `self` and dropping `topology` seems like a better choice.

lib/charms/prometheus_k8s/v0/prometheus_scrape.py
sed-i: This function does not rename anything though - it builds up a string, that ends up being used elsewhere.

lib/charms/prometheus_k8s/v0/prometheus_scrape.py
sed-i: I made a small addition. Take a look again?

lib/charms/prometheus_k8s/v0/prometheus_scrape.py
sed-i: Scratch that, we already have `InvalidAlertRulePathError`.

lib/charms/prometheus_k8s/v0/prometheus_scrape.py
sed-i: Agreed, but the code has changed a bit - there is an `else` now and no return.

lib/charms/prometheus_k8s/v0/prometheus_scrape.py
sed-i: Refactored.

lib/charms/prometheus_k8s/v0/prometheus_scrape.py
sed-i: Overwriting (`alert_group["name"]` value) is a common pattern.
This is now refactored a bit, take a look?

tests/unit/helpers.py
```diff
@@ -0,0 +1,34 @@
+# Copyright 2020 Canonical Ltd.
+# See LICENSE file for licensing details.
+
+import os
+import tempfile
+from typing import Tuple
+
+
+class Sandbox:
```
simskij, line 9: This name is not very descriptive of its use. Consider switching to `TempFolderSandbox` or something else that's more descriptive.

tests/unit/prometheus_alert_rules/nested_rules_dir/cpu_overuse.rule
```diff
@@ -1,8 +0,0 @@
-alert: CPUOverUseNested
```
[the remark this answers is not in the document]
simskij: I wonder this as well?

tests/unit/test_endpoint_consumer.py
```diff
@@ -364,11 +364,10 @@ def test_consumer_returns_alerts_indexed_by_group_name(self):
         self.harness.add_relation_unit(rel_id, "consumer/0")
         self.assertEqual(self.harness.charm._stored.num_events, 1)
 
-        alerts = self.harness.charm.prometheus_consumer.alerts()
-        self.assertEqual(len(alerts), 1)
-        for name, alert_group in alerts.items():
-            group = next((group for group in ALERT_RULES["groups"] if group["name"] == name), None)
-            self.assertDictEqual(alert_group, group)
+        rules_file = self.harness.charm.prometheus_consumer.alerts()
+        self.maxDiff = None
```
simskij, line 368:
```suggestion
```
debug artifact?

tests/unit/test_endpoint_provider.py
```diff
@@ -272,9 +277,7 @@ def test_a_bad_alert_expression_logs_an_error(self, _):
             self.harness.add_relation_unit(rel_id, "provider/0")
             messages = sorted(logger.output)
             self.assertEqual(len(messages), 1)
-            self.assertIn(
-                "Invalid alert rule missing_expr.rule: missing an 'expr' property", messages[0]
-            )
+            self.assertIn("Invalid rules file: missing_expr.rule", messages[0])
```
simskij, line 280: There is a loss of detail here as to why it failed. Why?

tests/unit/test_endpoint_provider.py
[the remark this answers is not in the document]
simskij: > I am not in favor of expressions such as `assert "expr" in dict["groups"][0]["rules"][0]`.

If they are needed to properly test the logic... well 😅 🤷🏽‍♂️

tox.ini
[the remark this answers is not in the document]
simskij: Then please do so we get an up-to-date view of what's going to change. 👍🏽

lib/charms/prometheus_k8s/v0/prometheus_scrape.py
```diff
@@ -545,84 +546,192 @@ def as_dict_with_promql_labels(self):
...
 
-def load_alert_rules_from_dir(
-    dir_path: Union[str, Path], topology: JujuTopology, *, recursive: bool = False
-) -> List[dict]:
-    """Load alert rules from rule files.
+    @staticmethod
+    def _is_custom_format(rules_dict: dict) -> bool:
+        # one alert rule per file
+        # TODO validate e.g. against schema using pydantic
+        return set(rules_dict) >= {"alert", "expr"}
```
simskij, line 607: Make these regular private class methods or free functions.

lib/charms/prometheus_k8s/v0/prometheus_scrape.py
```diff
@@ -545,84 +546,192 @@ def as_dict_with_promql_labels(self):
...
+                rule_file = yaml.safe_load(rf)
+
+            except Exception as e:
+                logger.error("Failed to read alert rules from %s: %s", file_path.name, e)
+                return []
+
+            # TODO validate input
+            if AlertRules._is_official_format(rule_file):
+                alert_groups = rule_file["groups"]
+            elif AlertRules._is_custom_format(rule_file):
```
simskij, line 632: official and custom seems like unnecessary signaling, especially as we want to promote the use of the "custom format".

lib/charms/prometheus_k8s/v0/prometheus_scrape.py
```diff
@@ -545,84 +546,192 @@ def as_dict_with_promql_labels(self):
...
-        if rule := load_alert_rule_from_file(path, topology):
-            logger.debug("Reading alert rule from %s", path)
-            alerts[_group_name(path)].append(rule)
+        # Gather all alerts into a list of groups
+        paths = dir_path.glob("**/*.rule" if recursive else "*.rule")
+        for file_path in filter(Path.is_file, paths):
+            alert_groups_from_file = self._from_file(dir_path, file_path)
+            if alert_groups_from_file:
+                logger.debug("Reading alert rule from %s", file_path)
+                alert_groups.extend(alert_groups_from_file)
```
simskij, line 707:
```suggestion
        for file_path in filter(Path.is_file, paths):
            alert_groups_from_file = self._from_file(dir_path, file_path)
            if not alert_groups_from_file:
                continue
            logger.debug("Reading alert rule from %s", file_path)
            alert_groups.extend(alert_groups_from_file)
```

simskij (changes requested):

lib/charms/prometheus_k8s/v0/prometheus_scrape.py
sed-i: In this simple case I think the guard decreases clarity.

lib/charms/prometheus_k8s/v0/prometheus_scrape.py
sed-i: What do you mean? Are you proposing to rename them e.g. to "prom format" and "charm format"?
I was not aware we want to promote the custom format.

sed-i:
Rebased again.

tox.ini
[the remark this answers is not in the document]
sed-i: Probably a merge/rebase error in one of the previous commits.

lib/charms/prometheus_k8s/v0/prometheus_scrape.py
```diff
@@ -545,84 +546,192 @@ def as_dict_with_promql_labels(self):
...
+
+        All rules from files are aggregated into a data structure representing a single rule file.
+        All group names are augmented with juju topology.
+
+        Args:
+            path: either a rules file or a dir of rules files.
+            recursive: whether to read files recursively or not (no impact if `path` is a file).
+
+        Raises:
+            NotADirectoryError: if the provided path is invalid.
```
balbirthomas, line 722:
```suggestion
            InvalidAlertRulePathError: if the provided path is invalid.
```

lib/charms/prometheus_k8s/v0/prometheus_scrape.py
```diff
@@ -545,84 +546,192 @@ def as_dict_with_promql_labels(self):
...
+            alert_groups_from_file = self._from_file(dir_path, file_path)
+            if alert_groups_from_file:
+                logger.debug("Reading alert rule from %s", file_path)
+                alert_groups.extend(alert_groups_from_file)
 
-    # Gather all alerts into a list of groups since Prometheus
-    # requires alerts be part of some group
-    groups = [{"name": k, "rules": v} for k, v in alerts.items()]
-    return groups
+        return alert_groups if alert_groups else []
```
balbirthomas, line 709:
```suggestion
        return alert_groups
```
since if expression is redundant given `alert_groups` is correctly initialized.

lib/charms/prometheus_k8s/v0/prometheus_scrape.py
```diff
@@ -545,84 +546,192 @@ def as_dict_with_promql_labels(self):
...
+        rel_path = "" if rel_path == "." else rel_path.replace(os.path.sep, "_")
 
         # Generate group name:
-        #  - prefix, from the relative path of the rule file;
         #  - name, from juju topology
-        prefix = "" if relpath == "." else relpath.replace(os.path.sep, "_") + "_"
-        return "{}{}_alerts".format(prefix, topology.identifier)
+        #  - suffix, from the relative path of the rule file;
+        group_name_parts = [self.topology.identifier, rel_path, group_name, "alerts"]
+        return "_".join(filter(None, group_name_parts))
```
balbirthomas, line 682:
```suggestion
        # filter to remove empty strings
        return "_".join(filter(None, group_name_parts))
```
since a "null filter" may be surprising to the reader

balbirthomas (approved):
Thank you for your cooperation. I have not yet had the time to test that lma-proxy-operator with this PR but I think it should be file since AFAIU the structure of relation data has not been changed.

tests/unit/test_endpoint_provider.py
simskij: What I mean is that the new message is worse. The previous one actually revealed what the user needs to do to fix the error.

simskij (approved):
Once the last remaining comments have been addressed, I'm fine with merging this as well.

tests/unit/test_endpoint_provider.py
sed-i: The message changed because validity is more complex now, per `_is_custom_alert_rule_format()`.
Format is referenced in the module level docstring which is rendered on charmhub.
When `_is_custom_alert_rule_format()` checks better for validity (e.g. pydantic against a schema) then it would be easier to provide the precise reason.

sed-i:
I believe all comments are addressed now. @simskij ?

lib/charms/prometheus_k8s/v0/prometheus_scrape.py
simskij: If by the custom format you mean single rule, then yes. But skip it for now. I'm ok with this.

simskij:
I'm content. Go ahead and merge!

PR #164, opened 2021-11-15
Use python-libjuju from github

balbirthomas:
This commit Python libjuju is used from github because of a
bug in Model.reset() method which has been fixed in github
but not in the current released version.


sed-i (approved):
I had tests failing for some fascade error, maybe it was fixed in main since.

simskij (approved):

PR #166, opened 2021-11-15
add support for injecting label-matchers at runtime

simskij:
This PR adds support for injecting label-matchers to promql expressions at runtime. Currently, this is implemented for alert rules, which are injected with the juju topology, replacing the previous relabeling mechanics of jinja with actual promql parsing and validation.

There are a couple of issues with the current promql-transform binary which are tracked in that repo:

- https://github.com/canonical/promql-transform/issues/1
- https://github.com/canonical/promql-transform/issues/2

Still left todo:
- [x] https://github.com/canonical/prometheus-k8s-operator/pull/166#pullrequestreview-866903808

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -59,8 +59,29 @@ $ charmcraft pack
 
 ### Deploy
 
+There are two ways of deploying the prometheus-k8s operator, with and without 
+promql-transform.  Deploying the charm without PromQL Transform means you won't 
+get any Juju topology labels  injected into your alert rule expressions.
+
+#### Without PromQL Transform
+
 ```bash
-$ juju deploy ./prometheus-k8s_ubuntu-20.04-amd64.charm --resource prometheus-image=ubuntu/prometheus:latest
+$ juju deploy \
+    ./prometheus-k8s_ubuntu-20.04-amd64.charm \
+    --resource prometheus-image=ubuntu/prometheus:latest
+```
+
+#### With PromQL Transform
+
+Place the binary of your selected promql-transform version in the root of the prometheus-k8s
+charm directory. Official binaries are available from the 
+[promql-transform repository](https://github.com/canonical/promql-transform).
+
+```bash
+$ juju deploy \
+    ./prometheus-k8s_ubuntu-20.04-amd64.charm \
+    --resource prometheus-image=ubuntu/prometheus:latest
+    --resource promql-transform-amd64=./promql-transform
 ```
 
 ## Linting
--- a/lib/charms/prometheus_k8s/v0/prometheus_remote_write.py
+++ b/lib/charms/prometheus_k8s/v0/prometheus_remote_write.py
@@ -14,13 +14,15 @@ import json
 import logging
 import os
 from collections import OrderedDict
+import platform
+import subprocess
 from pathlib import Path
 from typing import Dict, List, Optional, Union
 
 import yaml
 from ops.charm import CharmBase, HookEvent, RelationEvent, RelationMeta, RelationRole
 from ops.framework import EventBase, EventSource, Object, ObjectEvents
-from ops.model import Relation
+from ops.model import ModelError, Relation
 
 # The unique Charmhub library identifier, never change it
 LIBID = "f783823fa75f4b7880eb70f2077ec259"
@@ -900,6 +902,7 @@ class PrometheusRemoteWriteProvider(Object):
 
         super().__init__(charm, relation_name)
         self._charm = charm
+        self._transformer = PromqlTransformer(self._charm)
         self._relation_name = relation_name
         self._endpoint_schema = endpoint_schema
         self._endpoint_address = endpoint_address
@@ -1010,7 +1013,7 @@ class PrometheusRemoteWriteProvider(Object):
             try:
                 scrape_metadata = json.loads(relation.data[relation.app]["scrape_metadata"])
                 identifier = ProviderTopology.from_relation_data(scrape_metadata).identifier
-                alerts[identifier] = alert_rules
+                alerts[identifier] = self._transformer.apply_label_matchers(alert_rules)
             except KeyError as e:
                 logger.warning(
                     "Relation %s has no 'scrape_metadata': %s",
@@ -1035,3 +1038,88 @@ class PrometheusRemoteWriteProvider(Object):
                         logger.error("Alert rules were found but no usable labels were present")
 
         return alerts
+
+
+# Copy/pasted from prometheus_scrape.py
+class PromqlTransformer:
+    """Uses promql-transform to inject label matchers into alert rule expressions."""
+
+    _path = None
+    _disabled = False
+
+    @property
+    def path(self):
+        """Lazy lookup of the path of promql-transform."""
+        if self._disabled:
+            return None
+        if not self._path:
+            self._path = self._get_transformer_path()
+            if not self._path:
+                logger.debug("Skipping injection of juju topology as label matchers")
+                self._disabled = True
+        return self._path
+
+    def __init__(self, charm):
+        self._charm = charm
+
+    def apply_label_matchers(self, rules):
+        """Will apply label matchers to the expression of all alerts in all supplied groups."""
+        if not self.path:
+            return rules
+        for group in rules["groups"]:
+            rules_in_group = group.get("rules", [])
+            for rule in rules_in_group:
+                topology = {}
+                # if the user for some reason has provided juju_unit, we'll need to honor it
+                # in most cases, however, this will be empty
+                for label in [
+                    "juju_model",
+                    "juju_model_uuid",
+                    "juju_application",
+                    "juju_charm",
+                    "juju_unit",
+                ]:
+                    if label in rule["labels"]:
+                        topology[label] = rule["labels"][label]
+                rule["expr"] = self._apply_label_matcher(rule["expr"], topology)
+        return rules
+
+    def _apply_label_matcher(self, expression, topology):
+        if not topology:
+            return expression
+        if not self.path:
+            logger.debug(
+                "`promql-transform` unavailable. leaving expression unchanged: %s", expression
+            )
+            return expression
+        args = [str(self.path)]
+        args.extend(
+            ['--label-matcher={}="{}"'.format(key, value) for key, value in topology.items()]
+        )
+
+        args.extend(["{}".format(expression)])
+        # noinspection PyBroadException
+        try:
+            return self._exec(args)
+        except Exception as e:
+            logger.debug('Applying the expression failed: "{}", falling back to the original', e)
+            return expression
+
+    def _get_transformer_path(self) -> Optional[Path]:
+        arch = platform.processor()
+        arch = "amd64" if arch == "x86_64" else arch
+        res = "promql-transform-{}".format(arch)
+        try:
+            path = self._charm.model.resources.fetch(res)
+            os.chmod(path, 0o777)
+            return path
+        except NotImplementedError:
+            logger.debug("System lacks support for chmod")
+        except (NameError, ModelError):
+            logger.debug('No resource available for the platform "{}"'.format(arch))
+        return None
+
+    def _exec(self, cmd):
+        result = subprocess.run(cmd, check=False, stdout=subprocess.PIPE)
+        output = result.stdout.decode("utf-8").strip()
+        return output
--- a/lib/charms/prometheus_k8s/v0/prometheus_remote_write.py
+++ b/lib/charms/prometheus_k8s/v0/prometheus_remote_write.py
@@ -13,9 +13,9 @@ should use the `PrometheusRemoteWriteProducer`.
 import json
 import logging
 import os
-from collections import OrderedDict
 import platform
 import subprocess
+from collections import OrderedDict
 from pathlib import Path
 from typing import Dict, List, Optional, Union
 
--- a/lib/charms/prometheus_k8s/v0/prometheus_remote_write.py
+++ b/lib/charms/prometheus_k8s/v0/prometheus_remote_write.py
@@ -1011,19 +1011,9 @@ class PrometheusRemoteWriteProvider(Object):
                 continue
 
             try:
-<<<<<<< HEAD
                 scrape_metadata = json.loads(relation.data[relation.app]["scrape_metadata"])
                 identifier = ProviderTopology.from_relation_data(scrape_metadata).identifier
                 alerts[identifier] = self._transformer.apply_label_matchers(alert_rules)
-=======
-                # as this might very well contain multiple groups from different origins,
-                # and as such, with different topologies, we need to flatten the groups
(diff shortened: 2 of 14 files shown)


lib/charms/prometheus_k8s/v0/prometheus_scrape.py
```diff
@@ -1758,3 +1793,56 @@ def _relabel_configs(self) -> list:
...
+            raise MissingTransformResourceError(res)
+
+    def apply_label_matcher(self, expr: str, additional_label_filters: dict) -> str:
+        if self._path is None:
+            logger.debug(
+                "No `promql-transform` available, skipping the application of the label matchers."
+            )
+            return expr
+
+        os.chmod(self._path, 0o755)
```
sed-i, line 1831: Have to admit I do not readily recall what is 0o755. I'd [use stats](https://docs.python.org/3/library/os.html#os.chmod).

metadata.yaml
```diff
@@ -45,10 +45,15 @@ resources:
       Promql-transform binary for adding label selectors on amd64. Promql-transform is available
       as a stand-alone tool on https://github.com/canonical/promql-transform
     filename: promql-transform-amd64
+    # The version uploaded with the charm should be the one shipped as released in the
+    # https://github.com/canonical/promql-transform repository that matches the version
+    # of Prometheus shipped within the `prometheus-image` resource.
```
sed-i, line 50: This sounds a bit confusing

tox.ini
```diff
@@ -86,8 +86,6 @@ deps =
     juju
     pytest
     git+https://github.com/charmed-kubernetes/pytest-operator.git
-allowlist_externals =
-    git
```
sed-i, line 90: git is still used in the commands. Wouldn't removing this trigger a red tox warning?

tests/unit/test_endpoint_consumer.py
```diff
@@ -134,11 +134,15 @@ def version(self):
...
 
 
+@unittest.mock.patch("platform.processor", lambda: "x86_64")
 class TestEndpointConsumer(unittest.TestCase):
     def setUp(self):
         metadata_file = open("metadata.yaml")
         self.harness = Harness(EndpointConsumerCharm, meta=metadata_file)
+        self.harness.add_resource("promql-transform-amd64", "dummy resource")
         self.addCleanup(self.harness.cleanup)
+        self.maxDiff = None
```
sed-i, line 144: Debugging artifact?

lib/charms/prometheus_k8s/v0/prometheus_scrape.py
```diff
@@ -772,8 +794,22 @@ def alerts(self) -> dict:
...
+                    rules = group.get("rules", [])
+
+                    for rule in rules:
+                        expression = rule["expr"]
+                        rule["expr"] = transformation_provider.apply_label_matcher(
+                            expression,
+                            JujuTopology.from_relation_data(
+                                scrape_metadata
+                            ).as_dict_with_promql_labels(),
+                        )
```
sed-i, line 811: I noticed that the following line in `load_alert_rule_from_file()` was not removed as part of this PR:
```python
rule["expr"] = topology.render(rule["expr"])
```
I'm not sure why. Also all the example rule files should have `%%juju...%%` removed, no?

tox.ini
simskij: yes, you're correct. must have been a cut and paste error. my bad

tests/unit/test_endpoint_consumer.py
simskij: yes, thanks for pointing it out. removed

metadata.yaml
simskij: Agreed. Tried to simplify the language a bit.

lib/charms/prometheus_k8s/v0/prometheus_scrape.py
simskij: That would require multiple modes to be combined (`IXUSR`, `IXGRP` and `IXOTH` - maybe `IWUSR` as well?) Not sure whether that would be more or less graspable than googling "linux permission 755"? I'm OK with doing either, so feel free to pick which option you like the most.

lib/charms/prometheus_k8s/v0/prometheus_scrape.py
simskij: Yeah, it does nothing if the templating string is not present, but I kept it for backwards-compatibility with existing rules. @mmanciop is this an OK breaking change for you given that we're still in the `edge` channel?

CONTRIBUTING.md
```diff
@@ -59,8 +59,16 @@ $ charmcraft pack
...
+in the [promql-transform](https://github.com/canonical/promql-transform) repository that matches
+the version of Prometheus shipped within the `prometheus-image` resource.
+
+
 ```bash
-$ juju deploy ./prometheus-k8s_ubuntu-20.04-amd64.charm --resource prometheus-image=ubuntu/prometheus:latest
+$ juju deploy ./prometheus-k8s_ubuntu-20.04-amd64.charm \
+  --resource prometheus-image=ubuntu/prometheus:latest \
+  --resource promql-transformer-amd64=<path_to_promql_transform> \
+  --resource promql-transformer-arm64=<path_to_promql_transform>
```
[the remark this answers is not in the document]
simskij: They're not. Will update. Thanks! 👍🏽

lib/charms/prometheus_k8s/v0/prometheus_scrape.py
simskij: Regarding removing the template strings from, that would break the topology injection currently as we can't attach resources during testing. I could mock away the apply_label_selectors function and make it return the correct values though. I'll ping you once I have it in 👍🏽

lib/charms/prometheus_k8s/v0/prometheus_scrape.py
sed-i: ...or even `[stat.filemode](https://docs.python.org/3/library/stat.html#stat.filemode)`.
Yes I agree.

lib/charms/prometheus_k8s/v0/prometheus_scrape.py
```diff
@@ -1758,3 +1793,56 @@ def _relabel_configs(self) -> list:
...
+        res = "promql-transform-{}".format(arch)
+        try:
+            self._path = self._charm.model.resources.fetch(res)
+        except NameError:
+            raise MissingTransformResourceError(res)
+
+    def apply_label_matcher(self, expr: str, additional_label_filters: dict) -> str:
+        if self._path is None:
+            logger.debug(
+                "No `promql-transform` available, skipping the application of the label matchers."
```
sed-i, line 1827: Would mentioning the unmodified expression be useful here?
```suggestion
                "`promql-transform` is not available; leaving expression unchanged: %s.", expr
```

lib/charms/prometheus_k8s/v0/prometheus_scrape.py
sed-i: Would it make more sense to move the chmod into `_fetch_transformer_path`?

lib/charms/prometheus_k8s/v0/prometheus_scrape.py
```diff
@@ -1758,3 +1793,56 @@ def _relabel_configs(self) -> list:
...
+        po = subprocess.Popen(
+            args="{} {} {}".format(self._path, " ".join(args), "'{}'".format(expr)),
+            shell=True,
+            stdout=subprocess.PIPE,
+        )
+
+        return_code = po.wait()
+
+        if return_code:
+            raise ValueError("An error occurred while invoking `promql-transform`.")
```
sed-i, line 1846: since Popen [raises exceptions](https://docs.python.org/3.8/library/subprocess.html#exceptions), maybe this should be inside a try-except. `os.chmod` too.

tests/unit/test_transform.py
```diff
@@ -0,0 +1,64 @@
...
+    """Test that the promql-transform implementation works."""
+
+    def setUp(self):
+        self.harness = Harness(TransformProviderCharm, meta=META)
+        self.harness.set_model_name("transform")
+        self.addCleanup(self.harness.cleanup)
+        self.harness.add_resource("promql-transform-amd64", "dummy resource")
+        self.harness.begin()
+
+    # pylint: disable=protected-access
```
sed-i, line 40: `disable protected access` could probably be applied to everything under `tests/`. [Example](https://github.com/canonical/alertmanager-operator/blob/bf95b19feb05b21f945559950f2bc47132747648/pyproject.toml#L30)

CONTRIBUTING.md
```diff
@@ -59,8 +59,15 @@ $ charmcraft pack
...
+The file specified for the `promql-transformer` resource should be the one shipped as released
+in the [promql-transform](https://github.com/canonical/promql-transform) repository that matches
+the version of Prometheus shipped within the `prometheus-image` resource.
+
+
 ```bash
-$ juju deploy ./prometheus-k8s_ubuntu-20.04-amd64.charm --resource prometheus-image=ubuntu/prometheus:latest
+$ juju deploy ./prometheus-k8s_ubuntu-20.04-amd64.charm \
+  --resource prometheus-image=ubuntu/prometheus:latest \
+  --resource promql-transformer-amd64=<path_to_promql_transform>
```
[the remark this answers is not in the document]
mmanciop: This is a very interesting idea methinks

lib/charms/prometheus_k8s/v0/prometheus_scrape.py
simskij: Tests did not check for that, so nothing had to be changed, but kept one with topology template labels to make sure we have that present until we make a decision on whether to keep it around.

lib/charms/prometheus_k8s/v0/prometheus_scrape.py
simskij: I'll add it for Popen. Would you mind pointing me at something explaining what kind of exceptions chmod raises? Can't find anything on python.org (or anywhere else)

CONTRIBUTING.md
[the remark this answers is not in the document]
jnsgruk: You can certainly just prime the binary during build: https://github.com/canonical/istio-operators/blob/cea5c7aadc4082c8bf6260937f0d53f728737d0d/charms/istio-pilot/charmcraft.yaml#L12

lib/charms/prometheus_k8s/v0/prometheus_scrape.py
sed-i: I triggered them manually myself...
`FileNotFoundError`, `PermissionError` are the most obvious to me. Maybe `OSError` too? With these kind of funcs I see no reason not to have a bare exception (`except Exception as e`).

CONTRIBUTING.md
```diff
@@ -59,8 +59,15 @@ $ charmcraft pack
...
 
+The file specified for the `promql-transformer` resource should be the one shipped as released
+in the [promql-transform](https://github.com/canonical/promql-transform) repository that matches
+the version of Prometheus shipped within the `prometheus-image` resource.
+
+
 ```bash
-$ juju deploy ./prometheus-k8s_ubuntu-20.04-amd64.charm --resource prometheus-image=ubuntu/prometheus:latest
+$ juju deploy ./prometheus-k8s_ubuntu-20.04-amd64.charm \
+  --resource prometheus-image=ubuntu/prometheus:latest \
```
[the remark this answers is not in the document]
simskij: Did you really mean to change the prometheus-image name? If so, please elaborate cause I don't follow.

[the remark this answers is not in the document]
sed-i: Yes, I think the doc should reflect what is currently on charmhub. Whoever goes to the CONTRIBUTING is probably aware of the concept of tags and that they can omit/swap it for "latest".

sed-i (approved):

tests/unit/test_endpoint_consumer.py
```diff
@@ -138,6 +136,11 @@ class TestEndpointConsumer(unittest.TestCase):
     def setUp(self):
         metadata_file = open("metadata.yaml")
         self.harness = Harness(EndpointConsumerCharm, meta=metadata_file)
+
+        self.harness.add_resource(
+            "promql-transform-amd64",
```
balbirthomas, line 141: Wouldn't this break in the (possibly unlikely) event someone executes these tests on "arm" even though you are patching `platform.processor()` ? Though we can ignore this for beta of course.

balbirthomas (approved):

tests/unit/test_endpoint_consumer.py
simskij: That is correct. Right now the tests require amd64, which I'm totally fine with for now.

lib/charms/prometheus_k8s/v0/prometheus_remote_write.py
```diff
@@ -924,7 +927,9 @@ def alerts(self) -> dict:
 
             try:
                 for group in alert_rules["groups"]:
-                    alerts[group["name"]] = {"groups": [group]}
+                    alerts[group["name"]] = self._transformer.apply_label_matchers(
+                        {"groups": [group]}
```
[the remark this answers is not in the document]
simskij: That was my thought yeah. We might even have to go lower than that to make sure each unique set of topology labels get their own file on disk. I'm very open to the fact that I might have missed something though.

lib/charms/prometheus_k8s/v0/prometheus_remote_write.py
```diff
@@ -933,3 +938,86 @@ def alerts(self) -> dict:
...
+                    "juju_charm",
+                    "juju_unit",
+                ]:
+                    if label in rule["labels"]:
+                        topology[label] = rule["labels"][label]
+                rule["expr"] = self._apply_label_matcher(rule["expr"], topology)
+        return rules
+
+    def _apply_label_matcher(self, expression, topology):
+        if not self.path:
```
simskij, line 988:
```suggestion
        if not topology:
            return expression
        if not self.path:
```

mmanciop (changes requested):
No mention of the promql-resource in CONTRIBUTING.md

simskij (commented):
Promql-transform injects additional quotes into the alert rule expression, effectively disabling all rules, as they wont match anything.

simskij:
> Promql-transform injects additional quotes into the alert rule expression, effectively disabling all rules, as they wont match anything.

Fixed in latest commit

CONTRIBUTING.md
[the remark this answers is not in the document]
simskij: Split into https://github.com/canonical/prometheus-k8s-operator/issues/216

sed-i (approved):

balbirthomas (approved):

PR #169, opened 2021-11-22
Fix incorrect charm name key in Juju topology object

balbirthomas:
A previous commit introduced a bug in accessing the charm name
relation data key in the Juju Topology object.

This commit fixes the issue.


rbarry82 (approved):

rbarry82:
I'm surprised there's no testcases around this

sed-i (approved):

PR #179, opened 2021-12-01
Minor updates and code cleanups

balbirthomas:
- Updated Library overview docs to reflect recent changes made to the interface
- Set unit names in relation data
- Updated unit tests to reflect unit names are present in relation data
- Correct docstrings of individual methods

Deferred to future PRs.
- Review of test cases and perhaps add more unit tests
- Other code cleanups (such as refactor of rules provider and related objects for which github tickets have been created).

### DIFF
--- a/lib/charms/prometheus_k8s/v0/prometheus_scrape.py
+++ b/lib/charms/prometheus_k8s/v0/prometheus_scrape.py
@@ -287,7 +287,7 @@ the YAML structure of Prometheus [scrape configuration]
 (https://prometheus.io/docs/prometheus/latest/configuration/configuration/#scrape_config).
 
 Units of consumer charm advertise their address over unit relation
-data using the `prometheus_scrape_host` key. While the
+data using the `prometheus_scrape_host_address` key. While the
 `scrape_metadata`, `scrape_jobs` and `alert_rules` keys in application
 relation data provide eponymous information.
 """
@@ -941,11 +941,14 @@ class MetricsEndpointConsumer(Object):
             A dictionary that maps unit names to unit addresses for
             the specified relation.
         """
-        return {
-            unit.name: relation.data[unit].get("prometheus_scrape_host")
-            for unit in relation.units
-            if relation.data[unit].get("prometheus_scrape_host")
-        }
+        hosts = {}
+        for unit in relation.units:
+            unit_name = relation.data[unit].get("prometheus_scrape_host_name")
+            host_address = relation.data[unit].get("prometheus_scrape_host_address")
+            if unit_name and host_address:
+                hosts.update({unit_name: host_address})
+
+        return hosts
 
     def _labeled_static_job_config(self, job, job_name_prefix, hosts, scrape_metadata) -> dict:
         """Construct labeled job configuration for a single job.
@@ -1322,8 +1325,11 @@ class MetricsEndpointProvider(Object):
         event is actually needed.
         """
         for relation in self._charm.model.relations[self._relation_name]:
-            relation.data[self._charm.unit]["prometheus_scrape_host"] = str(
+            relation.data[self._charm.unit]["prometheus_scrape_host_address"] = str(
                 self._charm.model.get_binding(relation).network.bind_address
+            )
+            relation.data[self._charm.unit]["prometheus_scrape_host_name"] = str(
+                self._charm.model.unit.name
             )
 
     @property
--- a/lib/charms/prometheus_k8s/v0/prometheus_scrape.py
+++ b/lib/charms/prometheus_k8s/v0/prometheus_scrape.py
@@ -483,20 +483,20 @@ class JujuTopology:
         self.application = application
         self.charm_name = charm_name
 
-    @staticmethod
-    def from_charm(charm):
+    @classmethod
+    def from_charm(cls, charm):
         """Factory method for creating the topology dataclass from a given charm."""
-        return JujuTopology(
+        return cls(
             model=charm.model.name,
             model_uuid=charm.model.uuid,
             application=charm.model.app.name,
             charm_name=charm.meta.name,
         )
 
-    @staticmethod
-    def from_relation_data(data: dict):
+    @classmethod
+    def from_relation_data(cls, data: dict):
         """Factory method for creating the topology dataclass from a relation data dict."""
-        return JujuTopology(
+        return cls(
             model=data["model"],
             model_uuid=data["model_uuid"],
             application=data["application"],
--- a/lib/charms/prometheus_k8s/v0/prometheus_scrape.py
+++ b/lib/charms/prometheus_k8s/v0/prometheus_scrape.py
@@ -1344,8 +1344,7 @@ class MetricsEndpointProvider(Object):
            A list of dictionaries, where each dictionary specifies a
            single scrape job for Prometheus.
         """
-        default_job = [{"metrics_path": "/metrics"}]
-        return self._jobs if self._jobs else default_job
+        return self._jobs if self._jobs else [DEFAULT_JOB]
 
     @property
     def _scrape_metadata(self) -> dict:
--- a/lib/charms/prometheus_k8s/v0/prometheus_scrape.py
+++ b/lib/charms/prometheus_k8s/v0/prometheus_scrape.py
@@ -3,8 +3,8 @@
 """## Overview.
 
 This document explains how to integrate with the Prometheus charm
-for the purposes of providing a metrics endpoint to Prometheus. It
-also explains how alternative implementations of the Prometheus charm
+for the purpose of providing a metrics endpoint to Prometheus. It
+also explains how alternative implementations of the Prometheus charms
 may maintain the same interface and be backward compatible with all
 currently integrated charms. Finally this document is the
 authoritative reference on the structure of relation data that is
@@ -14,7 +14,7 @@ provide a scrape target for Prometheus.
 ## Provider Library Usage
 
 This Prometheus charm interacts with its scrape targets using its
-charm library. Charms seeking to expose a metric endpoints for the
+charm library. Charms seeking to expose metric endpoints for the
 Prometheus charm, must do so using the `MetricsEndpointProvider`
 object from this charm library. For the simplest use cases, using the
 `MetricsEndpointProvider` object only requires instantiating it,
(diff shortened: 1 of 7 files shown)


sed-i:
Might be worthwhile to throw [codespell](https://github.com/canonical/operator-template/pull/9/files) into tox lint env?

lib/charms/prometheus_k8s/v0/prometheus_scrape.py
```diff
@@ -293,10 +287,9 @@ def _on_scrape_targets_changed(self, event):
 (https://prometheus.io/docs/prometheus/latest/configuration/configuration/#scrape_config).
 
 Units of consumer charm advertise their address over unit relation
-data using the `prometheus_scrape_host` key. While the
+data using the `prometheus_scrape_host_address` key. While the
```
mmanciop, line 290: @balbirthomas please remind me why we did this? It's going to come _really_ in handy with `ingress`, but I am not sure we had that in mind at the time.

tests/unit/test_endpoint_consumer.py
```diff
@@ -179,7 +184,12 @@ def setup_charm_relations(self, multi=False):
             )
             self.harness.add_relation_unit(rel_id, "other-consumer/0")
             self.harness.update_relation_data(
-                rel_id, "other-consumer/0", {"prometheus_scrape_host": "2.2.2.2"}
+                rel_id,
+                "other-consumer/0",
+                {
+                    "prometheus_scrape_host_address": "2.2.2.2",
+                    "prometheus_scrape_host_name": "other-consumer/0",
```
mmanciop, line 191: Bad name, it is a `unit_name`, not a `host_name`

mmanciop (changes requested):
I do not agree with delivering the unit name over a field called `host_name`. It is going to cause a lot of WTF all around :-)

lib/charms/prometheus_k8s/v0/prometheus_scrape.py
balbirthomas: When the Prometheus job specification use wildcard hosts (`*:8080`) then `MetricsEndpointConsumer` needs the IP addresses of all the units. There were two options at that time 1) `MetricsEndpointConsumer` should gather the IP address from relation event data 2) `MetricsEndpointProvider` units forward this data in unit relation data. I am not sure if I remember the precise details why we went down route 2 but I think at that time there was lot of discussion on how this should be done. The consensus in the team was for the provider charm in this case to forward the IP address through unit relation data. If I remember correctly this was the pattern Jeremy was recommending. I do remember creating a [ticket](https://github.com/canonical/operator/issues/560) for `ops` also. I think Graylog charm (at least Jeremy's branch) and other charms may also be doing the same. Happy to investigate if you think this needs changing.

tests/unit/test_endpoint_consumer.py
balbirthomas: Ok will change.

balbirthomas:
Manual test of unit names in cross model relations.
Screenshot from 2021-12-02 14-31-13

lib/charms/prometheus_k8s/v0/prometheus_scrape.py
```diff
@@ -953,11 +941,14 @@ def _relation_hosts(self, relation) -> dict:
...
         """
-        return {
-            unit.name: relation.data[unit].get("prometheus_scrape_host")
-            for unit in relation.units
-            if relation.data[unit].get("prometheus_scrape_host")
-        }
+        hosts = {}
+        for unit in relation.units:
+            unit_name = relation.data[unit].get("prometheus_scrape_unit_name")
+            host_address = relation.data[unit].get("prometheus_scrape_unit_address")
```
mmanciop, line 947: If we do a fallback to the previous label name we do not break compat with the previous versions of this lib.
Also, the name of the variable is wrong :-)

lib/charms/prometheus_k8s/v0/prometheus_scrape.py
```diff
@@ -953,11 +941,14 @@ def _relation_hosts(self, relation) -> dict:
...
             the specified relation.
         """
-        return {
-            unit.name: relation.data[unit].get("prometheus_scrape_host")
-            for unit in relation.units
-            if relation.data[unit].get("prometheus_scrape_host")
-        }
+        hosts = {}
+        for unit in relation.units:
+            unit_name = relation.data[unit].get("prometheus_scrape_unit_name")
```
mmanciop, line 946: This is not handling the case in which the other side is not adding the unit name, which is the case for every charm out there atm :-)

I feel we are being a touch too cavalier about backwards compatibility here.

mmanciop (changes requested):
Getting there, but there is no good reason to break backwards compatibility like this, even in a v0 version :-)

lib/charms/prometheus_k8s/v0/prometheus_scrape.py
balbirthomas: You mean leave the name as `prometheus_scrape_host` ?

lib/charms/prometheus_k8s/v0/prometheus_scrape.py
balbirthomas: This change does break compatibility with previous version. Charms will need to updated to the current version of the charm library. If ùnit_name` is not set then that scrape target is not scrapped. You mean we should allow scrapping of charms that do not set unit names ? If so in this case should the unit names be taken from the event as was previously done ?

mmanciop: I mean that, if the consumer is talking with an older version of the producer that sends relation data, it should not break. In such a case, keeping the old behavior as fallback is fine.

lib/charms/prometheus_k8s/v0/prometheus_scrape.py
balbirthomas: OK. I presume we do intend to deprecate `prometheus_scrape_host` at some point. I will put a TODO to this effect.

lib/charms/prometheus_k8s/v0/prometheus_scrape.py
balbirthomas: Done.

lib/charms/prometheus_k8s/v0/prometheus_scrape.py
balbirthomas: Done.

mmanciop (approved):

lib/charms/prometheus_k8s/v0/prometheus_scrape.py
```diff
@@ -1142,7 +1240,7 @@ def _resolve_dir_against_charm_path(charm: CharmBase, *path_elements: str) -> st
 
 
 class MetricsEndpointProvider(Object):
-    """Construct a metrics provider for a Prometheus charm."""
+    """An metrics endpoint for Prometheus."""
```
simskij, line 1243:
```suggestion
    """A metrics endpoint for Prometheus."""
```

lib/charms/prometheus_k8s/v0/prometheus_scrape.py
```diff
@@ -1155,16 +1253,20 @@ def __init__(
 
         If your charm exposes a Prometheus metrics endpoint, the
         `MetricsEndpointProvider` object enables your charm to easily
-        communicate to a consumer how to reach that metrics endpoint.
+        communicate how to reach that metrics endpoint.
 
-        A charm instantiating this object has the metrics endpoints
-        of each of its units scraped by the related Prometheus
+        By default a charm instantiating this object has the metrics
```
simskij, line 1258:
```suggestion
        By default, a charm instantiating this object has the metrics
```

simskij (changes requested):

lib/charms/prometheus_k8s/v0/prometheus_scrape.py
balbirthomas: done.

lib/charms/prometheus_k8s/v0/prometheus_scrape.py
balbirthomas: done.

lib/charms/prometheus_k8s/v0/prometheus_scrape.py
```diff
@@ -1359,15 +1462,20 @@ def _scrape_metadata(self) -> dict:
 
 
 class PrometheusRulesProvider(Object):
-    """A 'prometheus_scrape' provider class for rule files only.
+    """Forward only rules to Prometheus.
```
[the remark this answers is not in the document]
balbirthomas: done.

lib/charms/prometheus_k8s/v0/prometheus_scrape.py
```diff
@@ -1361,16 +1461,21 @@ def _scrape_metadata(self) -> dict:
...
-    """A 'prometheus_scrape' provider class for rule files only.
+class PrometheusRulesProvider(Object):
+    """Forward rules to Prometheus.
 
-    This class is for sending rules to prometheus, unlike :class:`MetricsEndpointProvider`, which
-    is used for configuring a scrape target (including rules).
-    This is useful for providing "free standing" rules.
+    This object may be used to forward rules to Prometheus. At present it only supports
+    forwarding alert rules. This is unlike :class:`MetricsEndpointProvider`, which
+    is used for forwarding both scrape targets and associated alert rules. This object
```
sed-i, line 1469:
```suggestion
    forwarding alert rules (support for recording rules is not yet implemented).
    In contrast to :class:`MetricsEndpointProvider`, which is used for forwarding both
    scrape job and associated alert rules, this object is only for forwarding 
    free-standing rules. This object
```

balbirthomas: I am fine with most of this suggestion with the only minor exception of the use of the phrase "free-standing rules". I would rather not invent such phraseology unless it is coming from a product specification. Are you ok with that ?

mmanciop (dismissed):

simskij (approved):

lib/charms/prometheus_k8s/v0/prometheus_scrape.py
simskij: I agree with you on this @balbirthomas. The less internal nomenclature we introduce, the better - at least from an onboarding perspective. In that case, maybe changing `free-standing rules` to `rules from the lma-rules charm` would make sense?

mmanciop: Well we are going to rename the `lma-rules` charm, sooo... :D

mmanciop (approved):

PR #192, opened 2022-01-05
add kubectl tests

sed-i:
This PR adds kubectl itests for spotting the various edge cases @mmanciop is repeatedly finding manually.

### DIFF
--- a/tests/integration/test_kubectl_delete.py
+++ b/tests/integration/test_kubectl_delete.py
@@ -36,15 +36,14 @@ async def test_deploy_from_local_path(ops_test, prometheus_charm):
 @pytest.mark.abort_on_fail
 async def test_kubectl_delete_pod(ops_test, prometheus_charm):
     pod_name = f"{app_name}-0"
+
     cmd = [
-        "sudo",
-        "microk8s.kubectl",
-        "delete",
-        "pod",
-        "-n",
-        ops_test.model_name,
-        pod_name,
+        "sg",
+        "microk8s",
+        "-c",
+        " ".join(["microk8s.kubectl", "delete", "pod", "-n", ops_test.model_name, pod_name]),
     ]
+
     logger.debug(
         "Removing pod '%s' from model '%s' with cmd: %s", pod_name, ops_test.model_name, cmd
     )
--- a/tests/integration/test_kubectl_delete.py
+++ b/tests/integration/test_kubectl_delete.py
@@ -39,8 +39,8 @@ async def test_deploy_from_local_path(ops_test, prometheus_charm):
 @pytest.mark.abort_on_fail
 async def test_kubectl_delete_pod(ops_test, prometheus_charm):
     pod_name = f"{app_name}-0"
-    log.info("Removing pod '%s' from model '%s'", pod_name, ops_test.model_name)
     cmd = [
+        "sudo",
         "microk8s.kubectl",
         "delete",
         "pod",
@@ -48,6 +48,7 @@ async def test_kubectl_delete_pod(ops_test, prometheus_charm):
         ops_test.model_name,
         pod_name,
     ]
+    log.info("Removing pod '%s' from model '%s' with cmd: %s", pod_name, ops_test.model_name, cmd)
 
     retcode, stdout, stderr = await ops_test._run(*cmd)
     assert retcode == 0, f"kubectl failed: {(stderr or stdout).strip()}"
--- a/tests/integration/test_kubectl_delete.py
+++ b/tests/integration/test_kubectl_delete.py
@@ -13,7 +13,7 @@ from helpers import (  # type: ignore[attr-defined]
     check_prometheus_is_ready,
 )
 
-log = logging.getLogger(__name__)
+logger = logging.getLogger(__name__)
 
 METADATA = yaml.safe_load(Path("./metadata.yaml").read_text())
 app_name = METADATA["name"]
@@ -23,7 +23,7 @@ resources = {"prometheus-image": METADATA["resources"]["prometheus-image"]["upst
 @pytest.mark.abort_on_fail
 async def test_deploy_from_local_path(ops_test, prometheus_charm):
     """Deploy the charm-under-test."""
-    log.debug("deploy local charm")
+    logger.debug("deploy local charm")
 
     async with IPAddressWorkaround(ops_test):
         await ops_test.model.deploy(
@@ -45,11 +45,13 @@ async def test_kubectl_delete_pod(ops_test, prometheus_charm):
         ops_test.model_name,
         pod_name,
     ]
-    log.debug("Removing pod '%s' from model '%s' with cmd: %s", pod_name, ops_test.model_name, cmd)
+    logger.debug(
+        "Removing pod '%s' from model '%s' with cmd: %s", pod_name, ops_test.model_name, cmd
+    )
 
     retcode, stdout, stderr = await ops_test._run(*cmd)
     assert retcode == 0, f"kubectl failed: {(stderr or stdout).strip()}"
-    log.info(stdout)
+    logger.debug(stdout)
     await ops_test.model.block_until(lambda: len(ops_test.model.applications[app_name].units) > 0)
     await ops_test.model.wait_for_idle(apps=[app_name], status="active", timeout=1000)
     await check_prometheus_is_ready(ops_test, app_name, 0)
--- a/tests/integration/test_kubectl_delete.py
+++ b/tests/integration/test_kubectl_delete.py
@@ -48,7 +48,7 @@ async def test_kubectl_delete_pod(ops_test, prometheus_charm):
         "Removing pod '%s' from model '%s' with cmd: %s", pod_name, ops_test.model_name, cmd
     )
 
-    retcode, stdout, stderr = await ops_test._run(*cmd)
+    retcode, stdout, stderr = await ops_test.run(*cmd)
     assert retcode == 0, f"kubectl failed: {(stderr or stdout).strip()}"
     logger.debug(stdout)
     await ops_test.model.block_until(lambda: len(ops_test.model.applications[app_name].units) > 0)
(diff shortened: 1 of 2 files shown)


tests/integration/test_kubectl_delete.py
```diff
@@ -0,0 +1,58 @@
...
+        )
+        await ops_test.model.wait_for_idle(apps=[app_name], status="active", timeout=1000)
+        await check_prometheus_is_ready(ops_test, app_name, 0)
+
+
+@pytest.mark.abort_on_fail
+async def test_kubectl_delete_pod(ops_test, prometheus_charm):
+    pod_name = f"{app_name}-0"
+    cmd = [
+        "sudo",
```
sed-i, line 43: I have sudo here for now because I wasn't able to correctly escape quotation marks for `sg -c`:
```python
cmd = [                                                                                        
    "sg",                                                                                      
    "-c",                                                                                      
    "microk8s",                                                                                
    r'"{}"'.format(                                                                            
        " ".join(                                                                              
            [                                                                                  
                "microk8s.kubectl",                                                            
                "delete",                                                                      
                "pod",                                                                         
                "-n",                                                                          
                ops_test.model_name,                                                           
                pod_name,                                                                      
            ]                                                                                  
        )                                                                                      
    ),                                                                                         
] 
```

mmanciop:
Cool stuff! But I object the implications pf defining these “edge cases”. Upgrades are a main concern for an operator.

mmanciop (commented):
The logic LGTM, but I have not made due diligence for the functional verification

tests/integration/test_kubectl_delete.py
```diff
@@ -0,0 +1,58 @@
...
+
+METADATA = yaml.safe_load(Path("./metadata.yaml").read_text())
+app_name = METADATA["name"]
+
+
+@pytest.mark.abort_on_fail
+async def test_deploy_from_local_path(ops_test, prometheus_charm):
+    """Deploy the charm-under-test."""
+    resources = {"prometheus-image": METADATA["resources"]["prometheus-image"]["upstream-source"]}
+    log.info("deploy local charm")
```
simskij, line 26:
```suggestion
    log.debug("deploy local charm")
```

let's try to keep the test logs clean 👍🏽

tests/integration/test_kubectl_delete.py
```diff
@@ -0,0 +1,58 @@
...
+
+    async with IPAddressWorkaround(ops_test):
+        await ops_test.model.deploy(
+            prometheus_charm, application_name=app_name, resources=resources
+        )
+        await ops_test.model.block_until(
+            lambda: len(ops_test.model.applications[app_name].units) > 0
+        )
+        await ops_test.model.wait_for_idle(apps=[app_name], status="active", timeout=1000)
```
simskij, line 35: good candidate for gathering. i also have a vague recollection of reading somewhere that `wait_for_idle` now also blocks until there is at least one unit that is idle?

tests/integration/test_kubectl_delete.py
```diff
@@ -0,0 +1,58 @@
...
+        ops_test.model_name,
+        pod_name,
+    ]
+    log.info("Removing pod '%s' from model '%s' with cmd: %s", pod_name, ops_test.model_name, cmd)
+
+    retcode, stdout, stderr = await ops_test._run(*cmd)
+    assert retcode == 0, f"kubectl failed: {(stderr or stdout).strip()}"
+    log.info(stdout)
+    await ops_test.model.block_until(lambda: len(ops_test.model.applications[app_name].units) > 0)
+    await ops_test.model.wait_for_idle(apps=[app_name], status="active", timeout=1000)
```
simskij, line 57: same comment as above.

tests/integration/test_kubectl_delete.py
```diff
@@ -0,0 +1,58 @@
...
+        "pod",
+        "-n",
+        ops_test.model_name,
+        pod_name,
+    ]
+    log.info("Removing pod '%s' from model '%s' with cmd: %s", pod_name, ops_test.model_name, cmd)
+
+    retcode, stdout, stderr = await ops_test._run(*cmd)
+    assert retcode == 0, f"kubectl failed: {(stderr or stdout).strip()}"
+    log.info(stdout)
```
simskij, line 55:
```suggestion
    log.debug(stdout)
```

tests/integration/test_kubectl_delete.py
```diff
@@ -0,0 +1,58 @@
...
+log = logging.getLogger(__name__)
+
+METADATA = yaml.safe_load(Path("./metadata.yaml").read_text())
+app_name = METADATA["name"]
+
+
+@pytest.mark.abort_on_fail
+async def test_deploy_from_local_path(ops_test, prometheus_charm):
+    """Deploy the charm-under-test."""
+    resources = {"prometheus-image": METADATA["resources"]["prometheus-image"]["upstream-source"]}
```
simskij, line 25: `upstream_source` could be extracted after `app_name` and then just used here to avoid cluttering the function body.

tests/integration/test_kubectl_delete.py
```diff
@@ -0,0 +1,58 @@
...
+    cmd = [
+        "sudo",
+        "microk8s.kubectl",
+        "delete",
+        "pod",
+        "-n",
+        ops_test.model_name,
+        pod_name,
+    ]
+    log.info("Removing pod '%s' from model '%s' with cmd: %s", pod_name, ops_test.model_name, cmd)
```
simskij, line 51:
```suggestion
    log.debug("Removing pod '%s' from model '%s' with cmd: %s", pod_name, ops_test.model_name, cmd)
```

simskij (changes requested):

tests/integration/test_kubectl_delete.py
sed-i: Moved to global scope

tests/integration/test_kubectl_delete.py
sed-i: Done

tests/integration/test_kubectl_delete.py
sed-i: Done

tests/integration/test_kubectl_delete.py
sed-i: I think the output of `_run` qualifies for `info`. What's the criterion?

tests/integration/test_kubectl_delete.py
sed-i: Correct. Removed `block_until`.

tests/integration/test_kubectl_delete.py
simskij: i don't think we have clear criteria. personally, i'd say: logs should be completely empty unless someone turns on debugging or something goes wrong.

sed-i: Done

tests/integration/test_kubectl_delete.py
rbarry82:
```
cmd = [
  "sg",
  "microk8s",
  "-c",
  " ".join(
    [
      "microk8s.kubectl",
      "delete",
      "pod",
      "-n",
      ops_test.model_name,
      pod_name
    ]
  )
]

File: tests/integration/test_kubectl_delete.py [hunk 8]
```diff
@@ -0,0 +1,56 @@
...
+async def test_deploy_from_local_path(ops_test, prometheus_charm):
+    """Deploy the charm-under-test."""
+    logger.debug("deploy local charm")
+
+    async with IPAddressWorkaround(ops_test):
+        await ops_test.model.deploy(
+            prometheus_charm, application_name=app_name, resources=resources
+        )
+        await ops_test.model.wait_for_idle(apps=[app_name], status="active", timeout=1000)
+        await check_prometheus_is_ready(ops_test, app_name, 0)
```
Abuelodelanada, line 33: We are avoiding inner functions ;-)

Abuelodelanada (approved):

tests/integration/test_kubectl_delete.py
```diff
@@ -0,0 +1,56 @@
...
+        "microk8s",
+        "-c",
+        " ".join(["microk8s.kubectl", "delete", "pod", "-n", ops_test.model_name, pod_name]),
+    ]
+
+    logger.debug(
+        "Removing pod '%s' from model '%s' with cmd: %s", pod_name, ops_test.model_name, cmd
+    )
+
+    retcode, stdout, stderr = await ops_test._run(*cmd)
```
dstathis, line 51: Not a blocker by any means but I think you should not use a private method like this. subprocess will work just as well and will never change.

rbarry82: It's also [available as `.run()`](https://github.com/charmed-kubernetes/pytest-operator/blob/f1dccecaaa05162b02173c16e841cf57aafd3928/pytest_operator/plugin.py#L222). It does inject environment variables for Juju, but we probably don't really need those, and `asyncio.create_subprocess_exec` may be enough on its own considering we're not actually interacting with Juju in this test.

dstathis (approved):
This comment is out of scope for this PR, but, I want to mention it anyway. `IPAddressWorkaround` should really be called something like `set_update_status_interval` as it has nothing to do with IP addresses.

tests/integration/test_kubectl_delete.py
sed-i: Thanks for pointing this out. This is a copy paste from old "idiomatic" pytest, and now
`_run = run  # backward compatibility alias`
I will change this to `run`.

tests/integration/test_kubectl_delete.py
Abuelodelanada: I do not know why I saw a `def` here

Abuelodelanada (approved):

dstathis (approved):

PR #199, opened 2022-01-12
Get Prometheus to monitor itself and export a dashboard

rbarry82:
The default Prometheus config monitors itself anyway, but a
dashboard must be sent back to Grafana for visibility to make this
useful. Grafana "includes" a Prometheus dashboard, but it does not
include Juju topology, and isn't as complete as the one included
here.


src/grafana_dashboards/prometheus-k8s_rev1.json.tmpl
```diff
@@ -0,0 +1,2972 @@
+{
```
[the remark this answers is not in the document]
sed-i: Is there juju topology in this file?

sed-i (dismissed):

metadata.yaml
```diff
@@ -24,6 +24,8 @@ containers:
 provides:
```
mmanciop, line 24: I expected this PR to also include a provided `metrics-endpoint` relation using the `prometheus_scrape` relation interface, so that Grafana Agent can monitor Prometheus (scraping the same endpoints Prometheus scrapes). Having Grafana Agent scrape Prometheus is more resilient in terms of Prometheus failures than Prometheus monitoring itself alone, e.g., when the Prometheus pod is in a crash loop.

mmanciop (changes requested):

mmanciop:
The PR is unfortunately borked: the job that self-scrapes Prometheus is NOT adding the Juju topology labels and that breaks ofc the dashboard:

Screenshot from 2022-01-12 11-56-53

Screenshot from 2022-01-12 11-55-05

mmanciop (changes requested):
The lack of Juju topology labels in the self-scrape job borks the dashboard, see comments. Also, we need to talk about exposing an actual self-monitoring relation via the `prometheus_scrape` interface.

mmanciop:
Also, what about PromQL alerting rules?

src/grafana_dashboards/prometheus-k8s_rev1.json.tmpl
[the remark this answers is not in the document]
rbarry82: Topology dropdowns are injected by the Grafana dashboard library itself. Similarly, the filename is derived from topology when it actually hits Grafana

rbarry82:
Topology is now added to self-monitoring, ad well as the relation for Grafana-agent. Alert rules next

mmanciop:
> Topology is now added to self-monitoring, ad well as the relation for Grafana-agent. Alert rules next

It’s soooo cool :-)

metadata.yaml
```diff
@@ -24,6 +24,10 @@ containers:
...
   grafana-source:
     interface: grafana_datasource
+  grafana-dashboard:
+    interface: grafana_dashboard
+  metrics-endpoint:
```
mmanciop, line 29: @simskij another self-monitoring relation to consider naming out of convention

simskij:
[9 votes for `self-*`](https://discourse.charmhub.io/t/how-should-we-call-relations-for-monitoring-cos-charms/5950), let's go with that and only use it for the ones that also provide that relation organically, ie. `self-metrics-endpoint` for Prometheus, `self-logs` for Loki, `self-logs` for Grafana Agent, and `self-dashboards` for Grafana itself.

metadata.yaml
simskij, line 29:
```suggestion
  self-metrics-endpoint:
```

simskij (changes requested):

simskij:
Closing this as it's already present through @Abuelodelanada's PR

PR #200, opened 2022-01-12
Change relation name from `receive-remote-write` to `prometheus-remote-write`

simskij:


simskij:
We'll stick with the relation name for now.

PR #201, opened 2022-01-12
Streamline remote write

sed-i:
This PR is intended to fix the rough edges of #147, specifically
- the structure of the alerts relation data
- missing observer for receive_remote_write_relation_changed

In addition, a new integration test is introduced to confirm that metrics and alerts from avalanche are reaching prometheus over the remote_write relation. (itest will fail until juju 2.9.23 is released to stable due to a [pebble bug](https://github.com/canonical/pebble/pull/96).)

Restructuring the alert data to accomodate use-cases for grafana-agent would be done in a separate PR, if deemed necessary (tracked in #202).


src/charm.py
```diff
@@ -220,8 +221,10 @@ def _command(self) -> str:
             args.append(f"--web.external-url={external_url}")
 
         # enable remote write if an instance of the relation exists
-        if self.model.relations[DEFAULT_REMOTE_WRITE_RELATION_NAME]:
-            args.append("--enable-feature=remote-write-receiver")
+        # FIXME for some reason this does not work
+        # FIXME if user changes relation name this will break anyway
+        # if self.model.relations[DEFAULT_REMOTE_WRITE_RELATION_NAME]:
+        #     args.append("--enable-feature=remote-write-receiver")
```
simskij, line 227: User will not change the relation name on this side of the charm, so that should be fine, but instead of `DEFAULT_REMOTE_WRITE_RELATION_NAME` it should prolly set the relation name to a var on self and check for that.

src/charm.py
```diff
@@ -211,6 +211,7 @@ def _command(self) -> str:
             "--web.enable-lifecycle",
             "--web.console.templates=/usr/share/prometheus/consoles",
             "--web.console.libraries=/usr/share/prometheus/console_libraries",
+            "--enable-feature=remote-write-receiver",
```
simskij, line 214: Once we've got it working smoothly, this needs to be removed again.

lib/charms/prometheus_k8s/v0/prometheus_remote_write.py
```diff
@@ -923,9 +923,15 @@ def alerts(self) -> dict:
                 continue
 
             try:
+                # remote write doesn't have scrape metadata; need to get topology from somewhere
+                # scrape_metadata = json.loads(relation.data[relation.app]["scrape_metadata"])
+                # topology = JujuTopology.from_relation_data(scrape_metadata)
+                topology = JujuTopology(
+                    "dont", "really", "know", "yet"
```
simskij, line 930: Just a note to self that we need to change this to something proper before merging.

src/charm.py
sed-i: You fixed it :) resolving

lib/charms/prometheus_k8s/v0/prometheus_remote_write.py
sed-i: You fixed it :) resolving

rbarry82:
There's a typo in the name of one of the integration tests -- file renamed as `grfana`

rbarry82 (approved):

simskij (approved):

PR #205, opened 2022-01-18
static analysis

sed-i:
This PR improves the coverage of the static analysis.
This was very much called for after I had to restart integration tests multiple times because of mistakes easily detectable by mypy.

- Add CodeQL
- Add proper py3.5 check
- Use codespell on lib as well
- setenv MYPYPATH correctly per testenv

### DIFF
--- a/src/charm.py
+++ b/src/charm.py
@@ -7,6 +7,7 @@
 import logging
 import os
 import re
+from typing import Optional, Dict
 
 import yaml
 from charms.alertmanager_k8s.v0.alertmanager_dispatch import AlertmanagerConsumer
@@ -268,7 +269,7 @@ class PrometheusCharm(CharmBase):
         if not (matched := re.match(r"[1-9][0-9]*[ymwdhs]", timeval)):
             self.unit.status = BlockedStatus(f"Invalid time spec : {timeval}")
 
-        return matched
+        return bool(matched)
 
     def _prometheus_global_config(self) -> dict:
         """Construct Prometheus global configuration.
@@ -292,7 +293,7 @@ class PrometheusCharm(CharmBase):
         Returns:
             a dictionary consisting of the alerting configuration for Prometheus.
         """
-        alerting_config = {}
+        alerting_config = {}  # type: Dict[str, list]
 
         alertmanagers = self.alertmanager_consumer.get_cluster_info()
 
@@ -329,10 +330,10 @@ class PrometheusCharm(CharmBase):
             "scheme": "http",
             "static_configs": [{"targets": [f"localhost:{self._port}"]}],
         }
-        prometheus_config["scrape_configs"].append(default_config)
+        prometheus_config["scrape_configs"].append(default_config)  # type: ignore
         scrape_jobs = self.metrics_consumer.jobs()
         for job in scrape_jobs:
-            prometheus_config["scrape_configs"].append(job)
+            prometheus_config["scrape_configs"].append(job)  # type: ignore
 
         return yaml.dump(prometheus_config)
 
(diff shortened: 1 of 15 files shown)


src/charm.py
```diff
@@ -369,7 +370,7 @@ def _external_hostname(self) -> str:
         return self.config["web_external_url"] or f"{self.app.name}"
 
     @property
-    def _remote_write_address(self) -> str:
+    def _remote_write_address(self) -> Optional[str]:
```
balbirthomas, line 373: This is bad practice for the following reason -  An empty string is false in a boolean context. Hence the return type does not need to be an optional string, it can always be a string. This ensures every method always has a single, unique, well defined return type. Failure to ensure single unique return types in general is bad practice because it prevents composition of functions i.e f(g(x)) etc.

tests/unit/test_charm.py
```diff
@@ -2,14 +2,15 @@
...
+from typing import Any, Dict
 from unittest.mock import patch
 
 import yaml
 from ops.testing import Harness
 
 from charm import PROMETHEUS_CONFIG, PrometheusCharm
 
-MINIMAL_CONFIG = {}
+MINIMAL_CONFIG: Dict[str, Any] = {}
```
balbirthomas, line 13: This is misleading because the values of the mapping is not really "Any" but actually a well defined but complex nested structure that would be horrendous to read if fully spec'd. For this reason I think it should be removed. If necessary we may remove MINIMAL_CONFIG itself since it seems to be empty and is probably a carry over after various refactorings.

src/charm.py
simskij: I agree on this one.

tests/unit/test_charm.py
simskij: Then make sure it's not used and then delete it 👍🏽

src/charm.py
rbarry82: As do I. But this means it should return `""`, not `None`, and the constructor for `PrometheusRemoteWriteProvider` should take `endpoint_address: Optional[str] = ""`. 

It doesn't otherwise change the logic.

sed-i: To emphasize Ryan's comment: this change is correct as far as typing goes because of what is returned:
```
return self._external_hostname if self.model.get_relation("ingress") else None
```
If you'd like, I can change it to `else ""`.
However, in this particular case, returning None is correct semantically and "" is wrong semantically: if there is no ip it is most likely you would not want to compose.

tests/unit/test_charm.py
sed-i: Removed.

src/charm.py
dstathis: Python is dynamically typed. Embrace it!

dstathis (approved):

src/charm.py
balbirthomas: This is a false understanding of "Dynamic Typing". Dynamic Typing is effectively employed when you use alternatives types that have the same/similar interface (behaviour) a.k.a duck typing. However if you pass a `NoneType` instead of a `string`, these two types do not have the same interface/behaviour. If your return type was an `Iterable` and you returned a either a list or a string that would still be okay though and an ideal design.

balbirthomas: To elaborate on my points above consider what happens if some other methods that invokes this one writes something like this
```
if len(returned_address) > 0:
    blah blah blah 
```
If you always returned a string this would not blow up. It does blow up because you return `NoneType`.
On the other hand if the invoking method did
```
if returned_address:
    blah blah blah
```
It is indeed fine even with NoneType but it will also be fine with empty string instead of NoneType.
Which solution "makes your code easy to use".

sed-i: Captured in https://github.com/canonical/prometheus-k8s-operator/issues/206 as this change would be outside the scope of this PR.

rbarry82 (approved):

balbirthomas (approved):

PR #207, opened 2022-01-19
Change return type of _remote_write_address to strictly string

### LINKED ISSUE #206: Change return type of `_remote_write_address` to strictly string
> This is bad practice for the following reason -  An empty string is false in a boolean context. Hence the return type does not need to be an optional string, it can always be a string. This ensures every method always has a single, unique, well defined return type. Failure to ensure single unique return types in general is bad practice because it prevents composition of functions i.e f(g(x)) etc.

_Originally posted by @balbirthomas in https://github.com/canonical/prometheus-k8s-operator/pull/205#discussion_r787556937_

> this means it should return `""`, not `None`, and the constructor for `PrometheusRemoteWriteProvider` should take `endpoint_address: Optional[str] = ""`. 
>
> It doesn't otherwise change the logic.

_Originally posted by @rbarry82 in https://github.com/canonical/prometheus-k8s-operator/pull/205#discussion_r787637435_

> To elaborate on my points above consider what happens if some other methods that invokes this one writes something like this
>```
>if len(returned_address) > 0:
>    blah blah blah 
>```
>If you always returned a string this would not blow up. It does blow up because you return `NoneType`.
>On the other hand if the invoking method did
>```
>if returned_address:
>    blah blah blah
>```
>It is indeed fine even with NoneType but it will also be fine with empty string instead of NoneType.
>Which solution "makes your code easy to use".

_Originally posted by @balbirthomas in https://github.com/canonical/prometheus-k8s-operator/pull/205#discussion_r788077614_

sed-i:
Discussion takes place in #206. Discussion is a continuation of [these comments](https://github.com/canonical/prometheus-k8s-operator/pull/205#discussion_r787556937) from a previous pull request.

(At the moment I think this should not be merged and the None kept.)

Fixes #206

### DIFF
--- a/lib/charms/prometheus_k8s/v0/prometheus_remote_write.py
+++ b/lib/charms/prometheus_k8s/v0/prometheus_remote_write.py
@@ -14,7 +14,7 @@ import json
 import logging
 import os
 from pathlib import Path
-from typing import Dict, List, Optional, Union
+from typing import Dict, List, Union
 
 import yaml
 from ops.charm import CharmBase, RelationEvent, RelationMeta, RelationRole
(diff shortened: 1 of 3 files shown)


balbirthomas (dismissed):
A huge thank you.

rbarry82 (approved):

balbirthomas (approved):

PR #209, opened 2022-01-24
allow emptying out alerts

sed-i:
This PR removes an "is not empty" guard before updating relation data with alerts.
This is needed for when a relation is removed, or in git-sync when config changes.
This change also guarantees there will always be a "alert_rules" key in relation data.


dstathis (approved):

mmanciop:
> This PR removes an "is not empty" guard before updating relation data with alerts. This is needed for when a relation is removed, or in git-sync when config changes.

When a relation is "removed", as in `relation_broken`, there is no need of removing its data: the Juju controller throws it all away, and if the same relation is established again between the same two Juju applications, it gets a new ID and starts with empty data bags. Or what am I missing @sed-i ?

I buy the use-case for `git-sync` though.

balbirthomas (approved):

PR #211, opened 2022-01-27
Bring prometheus_remote_write up to speed

rbarry82:
Merge in the classes/changes in JujuTopology and AlertRules. Also
ensure that `PrometheusremoteWriteProvider.alerts()` handles missing
metadata in the same way as `prometheus_scrape`

Bring `LIBPATCH` back down to 1. This hasn't been published yet,
and Charmcraft requires that LIBPATCH numbers go sequentially, with
no missing numbers allowed.

### DIFF
--- a/lib/charms/prometheus_k8s/v0/prometheus_scrape.py
+++ b/lib/charms/prometheus_k8s/v0/prometheus_scrape.py
@@ -566,7 +566,7 @@ class JujuTopology:
     def identifier(self) -> str:
         """Format the topology information into a terse string."""
         # This is odd, but may have `None` as a model key
-        return "_".join([str(val) for val in self.as_dict().values()])
+        return "_".join([str(val) for val in self.as_dict().values()]).replace("/", "_")
 
     @property
     def promql_labels(self) -> str:
--- a/lib/charms/prometheus_k8s/v0/prometheus_remote_write.py
+++ b/lib/charms/prometheus_k8s/v0/prometheus_remote_write.py
@@ -13,11 +13,12 @@ should use the `PrometheusRemoteWriteProducer`.
 import json
 import logging
 import os
+from collections import OrderedDict
 from pathlib import Path
-from typing import Dict, List, Union
+from typing import Dict, List, Optional, Union
 
 import yaml
-from ops.charm import CharmBase, RelationEvent, RelationMeta, RelationRole
+from ops.charm import CharmBase, HookEvent, RelationEvent, RelationMeta, RelationRole
 from ops.framework import EventBase, EventSource, Object, ObjectEvents
 from ops.model import Relation
 
@@ -29,7 +30,7 @@ LIBAPI = 0
 
 # Increment this PATCH version before using `charmcraft publish-lib` or reset
 # to 0 if you are raising the major API version
-LIBPATCH = 3
+LIBPATCH = 1
 
 
 logger = logging.getLogger(__name__)
@@ -96,7 +97,20 @@ class JujuTopology:
 
     STUB = "%%juju_topology%%"
 
-    def __init__(self, model: str, model_uuid: str, application: str, charm_name: str):
+    def __new__(cls, *args, **kwargs):
+        """Reject instantiation of a base JujuTopology class. Children only."""
+        if cls is JujuTopology:
+            raise TypeError("only children of '{}' may be instantiated".format(cls.__name__))
+        return object.__new__(cls)
+
+    def __init__(
+        self,
+        model: str,
+        model_uuid: str,
+        application: str,
+        unit: Optional[str] = "",
+        charm_name: Optional[str] = "",
+    ):
         """Build a JujuTopology object.
 
         A `JujuTopology` object is used for storing and transforming
@@ -112,16 +126,18 @@ class JujuTopology:
             model: a string name of the Juju model
             model_uuid: a globally unique string identifier for the Juju model
             application: an application name as a string
+            unit: a unit name as a string
             charm_name: name of charm as a string
         """
         self.model = model
         self.model_uuid = model_uuid
         self.application = application
         self.charm_name = charm_name
+        self.unit = unit
 
     @classmethod
     def from_charm(cls, charm):
-        """Factory method for creating the `JujuTopology` dataclass from a given charm.
+        """Factory method for creating `JujuTopology` children from a given charm.
 
         Args:
             charm: a `CharmBase` object for which the `JujuTopology` has to be constructed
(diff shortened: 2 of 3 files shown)


sed-i (approved):

dstathis (dismissed):

rbarry82:
@simskij @sed-i can you check the avalanche tests?

IMO, this should just be removed and integrated into the avalanche charm instead to verify functionality. This is way too fragile to belong here.

Pulling a charm which appears to be published from a non-`main` branch (considering that `main` does not have a remote-write relation) and an `edge` channel before our libraries have settled out is bad. Leaving them as `xfail` for now is probably ideal

This works with Grafana Agent remote write, which is an actual use case for COS. Load generation could go into the bundle, or test prometheus against avalanche in its own repo, or whatever, but grafana-agent should be sufficient to cover the integration use cases, and at least everyone is more or less familiar with how that works, whereas `label_kkkkk_0` is completely meaningless to me even looking at the avalanche charm code.

Where does it come from? What does it mean? Why `kkkkk`? Is this configurable? Does it change every time? I looked for it in the avalanche charm repo once I found the right branch, but didn't find anything. Who do I ask?

These are a bunch of questions which don't fit at all into the problem domain or mental model of COS. There are enough moving pieces to keep track of for all of us. It's probably possible for most of us to look at `grafana-agent` tests and figure out what the assigned topology labels should be, etc.

sed-i:
@rbarry82 
- Indeed, itests now take even longer than before because charmcraft 1.4.0 takes much longer to build the first time, and for itests, every time is the first time.
- I can and will improve docs around `label_kkkkk_0` when I get to it, thanks for pointing out.
- Validating remote-write with grafana-agent seems fine to me. I used avalanche at the time because I am familiar with it. There was no intention to load test, and the number of metrics etc. is set to minimum for that reason. But regardless grafana-agent is probably more straightforward.

lib/charms/prometheus_k8s/v0/prometheus_scrape.py
```diff
@@ -566,7 +566,7 @@ def from_relation_data(cls, data: dict):
     def identifier(self) -> str:
         """Format the topology information into a terse string."""
         # This is odd, but may have `None` as a model key
```
sed-i, line 568: Can you add to the comment what you wrote in the commit message? Pebble creating subpath... (?)

sed-i (approved):

rbarry82:
> @rbarry82
> 
> * Indeed, itests now take even longer than before because charmcraft 1.4.0 takes much longer to build the first time, and for itests, every time is the first time.
> * I can and will improve docs around `label_kkkkk_0` when I get to it, thanks for pointing out.
> * Validating remote-write with grafana-agent seems fine to me. I used avalanche at the time because I am familiar with it. There was no intention to load test, and the number of metrics etc. is set to minimum for that reason. But regardless grafana-agent is probably more straightforward.

I suppose that I'm not really worried about the runtime  -- it was definitely noticeable running the tests, but memoizing it when we get around to a testing harness solves that.

It's more that it re-raises the idea that we need to come to an agreement on integration tests. The pyramid we spoke about _initially_ proposed having very simple charms ala `prometheus-tester` and `grafana-tester` which exercise the actual running charm in a manner very similar to the avalanche "always firing charm". Changes in the `-tester` charms would make it immediately obvious that something changed in a way which may break API compatibility, but they're complex enough to perform "real" operations.

Pulling from charmhub (which `grafana-agent` is also doing) can be problematic in multiple ways:
* Avalanche is frustrating here, and I wanted you or @simskij to take a look because it's essentially a black box from a testing perspective. Right now, it fails, but it's not obvious to me whether it _should_ be failing or not. In the case of, say, the proxy, or our charms which interact with external teams, ensuring that a patch doesn't break compatibility with redis or somethng is great. As a very rough, never-been-published library, prometheus_remote_write doesn't carry that kind of contract. Adding them back in the future once they hit `stable` or get close is great.
* For now, it's an open question as to "should I build a new copy of grafana-agent locally, update the library, modify the test to point there instead, and if it passes, push a new version to charmhub and re-trigger the tests? If I do that, there's not an obvious indication in _this_ repo that I modified the other side of the test, and you have no way of knowing without deconstructing the charm whether I just hardcoded the tests to pass. If the integration tests can be modified to pass from outside the repo and successfully gate, then what good are they?
* There's also a chicken-and-egg problem where grafana-agent depends on prometheus, and prometheus depends on grafana-agent. An intentional, approved change for the API happens. But your patch to either one will fail the test, and the only way to break the deadlock is to force merging on one of the repos with failing tests, then re-trigger the other side, and this problem only expands as the web of dependencies grow if we put them in each repo.
* Pulling from charmhub provides an out-of-band way to get test to turn green, but, as the avalanche test conveniently demonstrates, there's no guarantees that what's on charmhub has any correlation to what's in `main`, and it doesn't necessarily have to have a correlation to what's in _any repo_. I could easily rebuild avalanche and push it right now to get the test to pass, then forget to submit a PR because it's late, and the next time you write a patch it suddenly fails again with no way to actually bisect changes other than reverse engineering things with a `pdb` breakpoint to stop the tests and poking around at what's in the juju data bags+charm variables, because they were never in source control.

Sure, we could brainstorm ways around this, like adding hashes to verify the integrity of the charms under test, but that adds additional pieces to a Rube Goldberg machine, and some part of the pipeline will inevitably lock up and need manual intervention.

I could keep going with reasons why. This is the reason for my "integration tests with other COS charms should happen in a separate repo dedicated to that purpose, which triggers 'canary' jobs through a webhook when a change/PR happens on any other repo in a list, and runs a 'full' suite on a timer which checks to see whether there were changes since the last run, and fires if there have been". Because nobody wants to wait 3 hours for CI when the PR is `bump LIBPATCH`.

Per-repo integration testers for our own charms should be, essentially, overbuilt mocks to make sure that the actual workload API endpoints look ok, too. Bundle/multi-repo/"suite"/whatever testers should have configurable run options (`ref_from_webhook|channel_from_config|local_charm_from_cmdline_opt`). 

If designed that way, we can break the circular coupling of "`grafana-agent` adds prometheus from charmhub && `prometheus` adds grafana-agent from charmhub, so changing either deadlocks integration tests without merging with failing gates or pushing unmerged code to charmhub.

Sorry, long comment, and I'm not _actually_ frustrated or angry about avalanche, it's just an abject demonstration of putting your hand back on a hot stove after being burned. We know exactly what the problems with designing integration tests of tightly coupled components with loose dependencies are, and it's this, but it made it in anyway and now we're re-learning lessons we already knew.

Let's just leave them as `xfail` and we can sort of the rest later.

balbirthomas (approved):

PR #213, opened 2022-01-28
Add the changes NRPE exporter needs to prometheus_scrape

rbarry82:
Sync up this library, too. It was missed in the initial proxy work, and resurrecting it from git reflow was not fun.

Essentially, add a public method to invoke building a scrape target, and add the ability to add custom fields to the scrape job.


balbirthomas (approved):

balbirthomas:
If waiting for idle in the integration tests is timing out I guess we can kick start it again and see if it passes but my experience has been using block until is more robust.

balbirthomas (approved):

sed-i (approved):

PR #215, opened 2022-01-28
Replace ingress relation with ingress-unit

johnsca:
Supports the Traefik ingress with the split out per-unit ingress relation, `ingress-unit`.


johnsca:
Replaced with #218 due to typo in branch name.

PR #218, opened 2022-01-31
Replace ingress relation with ingress-unit

johnsca:
Supports the Traefik ingress with the split out per-unit ingress relation, `ingress-unit`.


mmanciop:
Discontinued due to change in scope: the charm we integrate with is going to be https://github.com/canonical/traefik-k8s-operator

PR #219, opened 2022-02-01
OPENG-588: Do not override provider metrics labels

balbirthomas:
This commit ensures that metrics labels from the provider are not
overriden by Prometheus. Since the honor labels attribute for all
jobs is enfored by the MetricsEndpointConsumer it can not itself
be overriden by any MetricsEndpointProvider.

### DIFF
--- a/src/charm.py
+++ b/src/charm.py
@@ -333,6 +333,7 @@ class PrometheusCharm(CharmBase):
         prometheus_config["scrape_configs"].append(default_config)  # type: ignore
         scrape_jobs = self.metrics_consumer.jobs()
         for job in scrape_jobs:
+            job["honor_labels"] = True
             prometheus_config["scrape_configs"].append(job)  # type: ignore
 
         return yaml.dump(prometheus_config)
--- a/tests/unit/test_charm.py
+++ b/tests/unit/test_charm.py
@@ -1,6 +1,7 @@
 # Copyright 2020 Canonical Ltd.
 # See LICENSE file for licensing details.
 
+import json
 import unittest
 from unittest.mock import patch
 
@@ -9,8 +10,13 @@ from ops.testing import Harness
 
 from charm import PROMETHEUS_CONFIG, PrometheusCharm
 
-SAMPLE_ALERTING_CONFIG = {
-    "alertmanagers": [{"static_configs": [{"targets": ["192.168.0.1:9093"]}]}]
+RELATION_NAME = "metrics-endpoint"
+DEFAULT_JOBS = [{"metrics_path": "/metrics"}]
+SCRAPE_METADATA = {
+    "model": "provider-model",
+    "model_uuid": "abcdef",
+    "application": "provider",
+    "charm_name": "provider-charm",
 }
 
 
@@ -137,6 +143,24 @@ class TestCharm(unittest.TestCase):
         config = container.pull(PROMETHEUS_CONFIG)
         prometheus_scrape_config = scrape_config(config, "prometheus")
         self.assertIsNotNone(prometheus_scrape_config, "No default config found")
+
+    def test_honor_labels_is_always_set_in_scrape_configs(self):
+        rel_id = self.harness.add_relation(RELATION_NAME, "provider")
+        self.harness.add_relation_unit(rel_id, "provider/0")
+        self.harness.update_relation_data(
+            rel_id,
+            "provider",
+            {
+                "scrape_metadata": json.dumps(SCRAPE_METADATA),
+                "scrape_jobs": json.dumps(DEFAULT_JOBS),
+            },
+        )
+
+        prometheus_scrape_config = yaml.safe_load(self.harness.charm._prometheus_config())
+        for job in prometheus_scrape_config["scrape_configs"]:
+            if job["job_name"] != "prometheus":
+                self.assertIn("honor_labels", job)
+                self.assertTrue(job["honor_labels"])
 
     @patch("prometheus_server.Prometheus.reload_configuration")
     def test_configuration_reload(self, trigger_configuration_reload):


simskij (dismissed):

lib/charms/prometheus_k8s/v0/prometheus_scrape.py
```diff
@@ -346,6 +346,7 @@ def _on_scrape_targets_changed(self, event):
     "label_limit",
     "label_name_length_limit",
     "label_value_lenght_limit",
+    "honor_labels",
```
mmanciop, line 349: @balbirthomas, we had agreed NOT to allow Providers to set this for now.

mmanciop (changes requested):

lib/charms/prometheus_k8s/v0/prometheus_scrape.py
mmanciop: E.g., here: https://github.com/canonical/prometheus-k8s-operator/blob/f35ff29e037dddfe826b1901d81ecb6248c1a562/src/charm.py#L335

mmanciop: I find that the proposed solution in the PR violates big times the element of least surprise: we allow to set a `honor_labels` key in the relation, then we override whatever value comes in in the `MetricsEndpointConsumer`. No. Much better to hard-code it in the `charm.py` of Prometheus and NOT allow the `honor_labels` key in the relation. That way, providers are not tricked into thinking they can control that.

balbirthomas: Ok. Let me give this another thought. I agree with the issue of "least surprise". I put it in allowed labels because we use it to sanitize jobs. But I can ensure the label is injected after job sanitization. I think it already is but let me double check.

balbirthomas: Removed honor_label from allowed keys. Had to modify one test which was checking that the keys returned by `jobs()` method were a strict subset of `ALLOWED_KEYS`. Please re-review.

simskij: Does that really address Michele's point about `prometheus-scrape-config` and such?

balbirthomas: Sorry missed that. Will look again.

balbirthomas: The honor labels is not set by the charm itself. Also added unit test to check it is always set.

src/charm.py
```diff
@@ -327,12 +327,14 @@ def _prometheus_config(self) -> str:
...
             "metrics_path": "/metrics",
             "honor_timestamps": True,
+            "honor_labels": True,
             "scheme": "http",
             "static_configs": [{"targets": [f"localhost:{self._port}"]}],
         }
         prometheus_config["scrape_configs"].append(default_config)  # type: ignore
         scrape_jobs = self.metrics_consumer.jobs()
         for job in scrape_jobs:
+            job["honor_labels"] = True
```
sed-i, line 337: Should the provider be allowed to set it to False?
Also I'd add a comment here that this is a deviation from prometheus defaults.

tests/unit/test_charm.py
```diff
@@ -138,6 +144,23 @@ def test_default_scrape_config_is_always_set(self):
...
+            "provider",
+            {
+                "scrape_metadata": json.dumps(SCRAPE_METADATA),
+                "scrape_jobs": json.dumps(DEFAULT_JOBS),
+            },
+        )
+
+        prometheus_scrape_config = yaml.safe_load(self.harness.charm._prometheus_config())
+        for job in prometheus_scrape_config["scrape_configs"]:
+            self.assertIn("honor_labels", job)
```
sed-i, line 161: This line could probably be removed - KerError would fail the test anyway.

tests/unit/test_charm.py
```diff
@@ -138,6 +144,23 @@ def test_default_scrape_config_is_always_set(self):
         prometheus_scrape_config = scrape_config(config, "prometheus")
         self.assertIsNotNone(prometheus_scrape_config, "No default config found")
 
+    def test_honor_labels_is_always_set_in_scrape_configs(self):
```
sed-i, line 147: Missing test cases:
- honor_labels is preset to False
- jobs spec is empty

sed-i (dismissed):

src/charm.py
```diff
@@ -327,12 +327,14 @@ def _prometheus_config(self) -> str:
             "scrape_timeout": "5s",
             "metrics_path": "/metrics",
             "honor_timestamps": True,
+            "honor_labels": True,
```
mmanciop, line 330: This is not needed. The default self-monitoring job comes with no labels we override in its scraping.

src/charm.py
mmanciop: No, see https://github.com/canonical/prometheus-k8s-operator/pull/219#discussion_r796530679 :-)

tests/unit/test_charm.py
mmanciop: > honor_labels is preset to False

Quite literally cannot happen.

mmanciop (dismissed):

tests/unit/test_charm.py
sed-i: Prometheus charm (consumer side) technically has no control over this, so it very much _can_ happen.

tests/unit/test_charm.py
balbirthomas: I see no harm in it being a two step process -- Assert key exists then Assert key has expected value.

src/charm.py
balbirthomas: I will confess I knew that. I was being lazy in writing the test case. By adding this line here I could make my test case simple. Without this line I would need extra code in my test case to exclude the self-monitoring job.

Ping here, if you think there is cause for concern here that I have not yet comprehended and I will stop being lazy.

tests/unit/test_charm.py
balbirthomas: Because jobs even on the provider side are sanitized using `ALLOWED_KEYS` ( kudos @mmanciop ) the only way the relation data can have `honor[labels]=False` is if relation data is manipulated directly by the provider charm rather than using the library. If there are the types of cases we would like to protect against then I claim I have already won our bug bounty competition for all the charms we ever wrote. :-)

rbarry82 (approved):

sed-i (dismissed):

src/charm.py
simskij: Yes please, fix it. 👍🏽

balbirthomas: Ok. Will refactor test to only check for honor labels key being set only in non-self monitoring jobs.

sed-i (approved):

balbirthomas:
@mmanciop Could you please re-review since merging is blocked.

PR #220, opened 2022-02-02
unit should block if reload_configuration fails

sed-i:
For now, this PR contains only a unit test demonstrating how a unit is in Active state even though reload configuration failed:

```
Falsifying example: test_unit_is_blocked_if_reload_configuration_fails(
    is_leader=False,
    num_units=1,
    self=<test_charm_status.TestActiveStatus testMethod=test_unit_is_blocked_if_reload_configuration_fails>,
)
```

This is somewhat of an edge case because this is unlikely to happen during startup, and is inconclusive (yet) if could happen on scaledown, but charm code should arguably be resilient to this edge case as well.


balbirthomas:
There can be multiple reasons why reload configuration fails. Should we block in all of those cases ? For example if reload configuration fails because connectivity with workload transiently interrupted, this is likely to self-heal if event is deferred.

sed-i:
> There can be multiple reasons why reload configuration fails. Should we block in all of those cases ? For example if reload configuration fails because connectivity with workload transiently interrupted, this is likely to self-heal if event is deferred.

True. But that's how the charms behaves now, iiuc. If we change the charm then we'd drop this test.

mmanciop (changes requested):
The change seems OK but I would also want to see this as an integration test using some bad configs that come through the relation. I am skeptical of the accuracy of the overall behavior, when we test it by mocking the `reload_configuration` method.

balbirthomas:
@sed-i could you please close this PR if it is no longer relevant.

balbirthomas:
@sed-i could you please close this PR if it is no longer relevant.

tests/unit/test_charm_status.py
```diff
@@ -0,0 +1,102 @@
...
+    In some cases (e.g. Ingress conflicts) the charm should go into blocked state.
+    """
+
+    def setUp(self) -> None:
+        self.app_name = "prometheus-k8s"
+
+    @patch_network_get(private_address="1.1.1.1")
+    @patch("charm.KubernetesServicePatch", lambda x, y: None)
+    @given(st.booleans(), st.integers(1, 5))
+    def test_unit_is_active_if_deployed_without_config_or_relations(self, is_leader, num_units):
```
balbirthomas, line 31: - Should a Prometheus charm block if it has no relations ?
- The Prometheus charm always sets a default config which makes it scrape itself.

In view of the above two concerns it is not clear to me what this method (test case) is testing for ? In particular what is new about this test case compared to the ones that already exist.

simskij (changes requested):

tox.ini
```diff
@@ -84,6 +84,7 @@ deps =
     responses
     -r{toxinidir}/requirements.txt
     deepdiff
+    hypothesis
```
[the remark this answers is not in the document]
sed-i: unittest an pytest are each a "framework" in a very different sense from the way hypothesis is a "framework". See references section [here](https://discourse.charmhub.io/t/unit-test-charmed-operators-with-hypothesis/5602).

mmanciop (dismissed):

src/charm.py
```diff
@@ -152,12 +141,16 @@ def _configure(self, _):
                 logger.error(f"Failed to replan; pebble plan: {container.get_plan().to_dict()}")
                 raise e
 
-        if (
```
balbirthomas, line 155: By removing these lines and not adding anything equivalent you are changing the behaviour of this charm. Actually you are reverting the behaviour to what existed before these lines were added. So I think it may be important to understand why these lines were added and why it is ok to now remove them. This is not entirely clear to me.

tests/unit/test_charm_status.py
```diff
@@ -0,0 +1,109 @@
...
+        """
+        # without the try-finally, if any assertion fails, then hypothesis would reenter without
+        # the cleanup, carrying forward the units that were previously added
+        self.harness = Harness(PrometheusCharm)
+        self.peer_rel_id = self.harness.add_relation("prometheus-peers", self.app_name)
+
+        try:
+            self.assertEqual(self.harness.model.app.planned_units(), 1)
+
+            # GIVEN any number of units present
```
balbirthomas, line 46: It is not clear to me what is the significance of varying the number of units in this test. Prometheus only sets its IP address into the peer relation. I do not see how this affects its reload behaviour.

tests/unit/test_charm_status.py
```diff
@@ -0,0 +1,109 @@
...
+# Copyright 2021 Canonical Ltd.
+# See LICENSE file for licensing details.
+
+import logging
+import unittest
+from unittest.mock import patch
+
+import hypothesis.strategies as st
+from helpers import patch_network_get
+from hypothesis import assume, given
```
balbirthomas, line 11: Change (`hypothesis`) still required here.

src/charm.py
```diff
@@ -132,18 +132,7 @@ def _configure(self, _):
...
-            prometheus_server = Prometheus(web_route_prefix=external_url.path)
-
-            reloaded = prometheus_server.reload_configuration()
-            if not reloaded:
-                logger.error("Prometheus failed to reload the configuration")
-                self.unit.status = BlockedStatus(CORRUPT_PROMETHEUS_CONFIG_MESSAGE)
-                return
-            logger.info("Prometheus configuration reloaded")
-        else:
+        if current_services != new_layer.services:
```
balbirthomas, line 135: When the pebble plan has changed, which should happen primarily due changes in the command line arguments, then shouldn't we be restarting Prometheus rather than reloading its configuration.

tests/unit/test_charm_status.py
```diff
@@ -0,0 +1,109 @@
...
+    def setUp(self) -> None:
+        self.app_name = "prometheus-k8s"
+
+    @patch_network_get(private_address="1.1.1.1")
+    @patch("charm.KubernetesServicePatch", lambda x, y: None)
+    @given(st.booleans(), st.integers(1, 5))
+    def test_unit_is_active_when_deployed_in_isolation(self, is_leader, num_units):
+        """Scenario: Unit is deployed without any user-provided config or regular relations.
+
+        The charm should end up in ActiveState when deployed in isolation with defaults only.
```
balbirthomas, line 36: 1. The purpose of this test is not clear to me. The method name suggests this test is checking that Prometheus is active when it is "deployed_in_isolation". The docstring suggests that test is checking if Prometheus is active "without any user-provided config or regular relations". The body of the test varies the number of deployed peer units from 1 through 5. Shouldn't the number of units be 1 if we are trying to "deploy in isolation" ?
2. It is not clear to me if and what "deploying in isolation" has anything to do with Prometheus reload behaviour .
3. If there is something new here, could [this existing test](https://github.com/canonical/prometheus-k8s-operator/blob/main/tests/unit/test_charm.py#L184) be modified to fulfil that requirement ?

tests/unit/test_charm_status.py
```diff
@@ -0,0 +1,109 @@
...
+                self.assertGreater(plan.to_dict().items(), {}.items())
+
+        finally:
+            # cleanup to prep for reentry by hypothesis' strategy
+            self.harness.cleanup()
+
+    @patch_network_get(private_address="1.1.1.1")
+    @patch("charm.KubernetesServicePatch", lambda x, y: None)
+    @given(st.booleans(), st.integers(1, 5))
+    def test_unit_is_blocked_if_reload_configuration_fails(self, is_leader, num_units):
```
balbirthomas, line 73: 1. Couldn't this test can be trivially reimplemented without using `hypothesis` and only using `unittest`. This can be done by patching the `reload_configuration()` method to return false just as it is done in this case. 
2. I fail to see how `num_units` is relevant to this test. 
3. Even though `is_leader` may appear to be relevant to this test it actually is not because setting a pebble plan restarting/reloading configuration is something all units of all charms do and should do regardless of if they are leader or not. Hence it is not clear to me how leadership status is relevant to this test.

balbirthomas (dismissed):

src/charm.py
```diff
@@ -152,12 +141,16 @@ def _configure(self, _):
...
+        external_url = urlparse(self._external_url)
+        prometheus_server = Prometheus(web_route_prefix=external_url.path)
+        reloaded = prometheus_server.reload_configuration()
+        if not reloaded:
+            logger.error("Prometheus failed to reload the configuration")
+            self.unit.status = BlockedStatus(CORRUPT_PROMETHEUS_CONFIG_MESSAGE)
             return
 
+        logger.info("Prometheus configuration reloaded")
         self.unit.status = ActiveStatus()
```
simskij, line 154: This changes the behavior significantly. Previously, the only blocked status we would "auto-heal" from was the Prometheus config being corrupt. Now, as long as the config is not corrupt, any other reason for blocks would also be discarded. Is this intentional?

simskij: Further, why is a subsequent reload needed after a replan? Replan will restart all services which changed layers. To my knowledge, the reason why we had a conditional check deciding whether to replan or reload, is to avoid unnecessary downtime in the case where a reload is sufficient.

I first thought that it might be because you wanted a trigger for detecting invalid configs, but as far as I've been told (thanks @balbirthomas for that clarification), Prometheus will exit with an error status if you try to launch it with a broken config?

simskij (changes requested):

sed-i (commented):
@balbirthomas @simskij At the moment, phrasing tests with hypothesis helps me understand the charm. Once we agree on the problem and on the solution I will remove hypothesis from the PR.

src/charm.py
simskij: > except the rare occasion when layer changes

This is the case I want to guard for, as it would otherwise mean it first is restarted, and then additionally reloaded

src/charm.py
```diff
@@ -152,12 +141,16 @@ def _configure(self, _):
                 logger.error(f"Failed to replan; pebble plan: {container.get_plan().to_dict()}")
                 raise e
```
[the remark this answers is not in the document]
balbirthomas: Not sure. This may be a question for Michele given [this commit](https://github.com/canonical/prometheus-k8s-operator/commit/2913c91505653e96c8958457243c74f8f03d08ec).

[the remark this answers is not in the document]
sed-i: @mmanciop any reason not to change this raise to block & return?

src/charm.py
```diff
@@ -133,32 +134,39 @@ def _configure(self, _):
...
-        ):
-            return
+            except ChangeError as e:
+                logger.error(
+                    "Failed to replan; pebble plan: %s; %s", container.get_plan().to_dict(), str(e)
+                )
+                self.unit.status = BlockedStatus(REPLAN_FAILED_MESSAGE)
+                return
 
-        self.unit.status = ActiveStatus()
```
sed-i, line 161: @balbirthomas iiuc, the existing code would incorrectly set the status to Active if the current status is e.g. Maintenance.

src/charm.py
sed-i: Done.

src/charm.py
```diff
@@ -133,32 +134,40 @@ def _configure(self, _):
...
+        # Auto-heal only from statuses managed by this method
+        if self.unit.status in [
+            WaitingStatus("Waiting for Pebble ready"),
+            BlockedStatus(CORRUPT_PROMETHEUS_CONFIG_MESSAGE),
+            BlockedStatus(REPLAN_FAILED_MESSAGE),
+        ]:
+            self.unit.status = ActiveStatus()
```
sed-i, line 170: I'm not sure about this approach for auto-heal:
- we probably want to heal only from statuses set by this method
- this exact approach would not work for variable messages e.g. `BlockedStatus(f"Invalid time spec : {timeval}")`
- for some reason `begin_with_initial_hooks` ends up with `Maintenance` (which the existing code masks) -- could this be a harness bug?

Both approaches -- touch status only in _configure, or auto-heal only from own statuses -- seem inadequate.

@balbirthomas @simskij

sed-i: With this new change of mine (only heal from own statuses) the itests fail too because they're [stuck on Maintenance or Unknown status](https://github.com/canonical/prometheus-k8s-operator/runs/5576228211?check_suite_focus=true#step:4:1449).

This is because Juju doesn't change it to Active on its own:

```shell
mkdir blank
charmcraft init -p blank
cd blank
# comment out: self.unit.status = ActiveStatus()
charmcraft pack
juju deploy ./blank_ubuntu-20.04-amd64.charm --resource httpbin-image=kennethreitz/httpbin
```

and the charm will go into Maintenance and at the end of startup change to unknown:

```python
# https://github.com/canonical/operator/blob/45e1e5a9e4e99492d09423436b8378f955c4f0a3/ops/model.py#L885
class UnknownStatus(StatusBase):
    """The unit status is unknown.
    A unit-agent has finished calling install, config-changed and start, but the
    charm has not called status-set yet.
    """
```

So I added Maintenance to the self healing list and now tests pass. Two questions:
1. Do we want this pattern?
2. Why some itests were stuck on Maintenance while others on Unknown?

mmanciop: (2) suggests that the units are doing different things. Do you see a difference between the status of the leader unit and that of the followers maybe?

src/charm.py
```diff
@@ -133,24 +133,32 @@ def _configure(self, _):
...
                 self.unit.status = BlockedStatus(CORRUPT_PROMETHEUS_CONFIG_MESSAGE)
                 return
+
             logger.info("Prometheus configuration reloaded")
+
         else:
+            # Layer changed - replan.
             container.add_layer(self._name, new_layer, combine=True)
             try:
+                # If a config is invalid then prometheus would exist immediately.
```
balbirthomas, line 151:
```suggestion
                # If a config is invalid then prometheus would exit immediately.
```

tests/unit/test_charm_status.py
```diff
@@ -0,0 +1,82 @@
...
+            self.harness.begin_with_initial_hooks()
+
+            # WHEN no config is provided or relations created
+
+            # THEN the unit goes into active state
+            self.assertIsInstance(self.harness.charm.unit.status, ActiveStatus)
+
+            # AND pebble plan is not empty
+            plan = self.harness.get_container_pebble_plan(self.harness.charm._name)
+            self.assertGreater(plan.to_dict().items(), {}.items())
```
balbirthomas, line 55:
```suggestion
            self.assertTrue(plan.to_dict())
```

tests/unit/test_charm_status.py
```diff
@@ -0,0 +1,82 @@
...
+            self.harness.begin_with_initial_hooks()
+
+            # WHEN no config is provided or relations created
+
+            # THEN the unit goes into blocked state
+            self.assertIsInstance(self.harness.charm.unit.status, BlockedStatus)
+
+            # AND pebble plan is not empty
+            plan = self.harness.get_container_pebble_plan(self.harness.charm._name)
+            self.assertGreater(plan.to_dict().items(), {}.items())
```
balbirthomas, line 82:
```suggestion
            self.assertTrue(plan.to_dict())
```

balbirthomas (approved):

PR #224, opened 2022-02-02
Load both .rule and .rules files

sed-i:
Since the introduction of the dual rule format (#161 - one rule per file as well as groups of alert rules), two file extensions emerged: `.rule` and `.rules`.
This PR adds `.rule` to the files being globbed and loaded from disk.

### DIFF
--- a/lib/charms/prometheus_k8s/v0/prometheus_scrape.py
+++ b/lib/charms/prometheus_k8s/v0/prometheus_scrape.py
@@ -866,7 +866,7 @@ class AlertRules:
         Returns:
             List of files in `dir_path` that have one of the suffixes specified in `suffixes`.
         """
-        all_files_in_dir = Path(dir_path).glob("**/*" if recursive else "*")
+        all_files_in_dir = dir_path.glob("**/*" if recursive else "*")
         return list(filter(lambda f: f.is_file() and f.suffix in suffixes, all_files_in_dir))
 
     def _from_dir(self, dir_path: Path, recursive: bool) -> List[dict]:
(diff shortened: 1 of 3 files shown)


rbarry82:
I mean, looks ok, but why are there different extensions for:

```
- matchers: [ 'model="foo", environment="bar"}' ]
```

vs
```
- matchers: [ model=foo, environment=bar ]
````

**Review by rbarry82 — DISMISSED** [review]

File: lib/charms/prometheus_k8s/v0/prometheus_scrape.py [hunk 1]
```diff
@@ -852,6 +852,23 @@ def _group_name(self, root_path: str, file_path: str, group_name: str) -> str:
...
         return "_".join(filter(None, group_name_parts))
 
+    @classmethod
+    def _multi_suffix_glob(
+        cls, dir_path: Path, suffixes: List[str], recursive: bool = True
+    ) -> list:
+        """Helper function for getting all files in a directory that have a matching suffix.
+
+        Args:
+            dir_path: path to the director to glob from.
```
Abuelodelanada, line 862: director -> directory

Abuelodelanada (approved):

lib/charms/prometheus_k8s/v0/prometheus_scrape.py
sed-i: Thanks! Fixed

Abuelodelanada (approved):

rbarry82 (approved):

Issue #223, opened 2022-02-02
`_apply_label_matcher` creates an unwarranted file on filesytsem


sed-i:
The test `test_consumer_returns_alerts_rules_file` calls 
```
rules_file = self.harness.charm.prometheus_consumer.alerts()
```
which eventually calls `_apply_label_matcher` results in a file named `0.12` created in the charms folder, with the following contents:
```
process_cpu_seconds_total{juju_application="consumer-tester",juju_model="None",juju_model_uuid="f2c1b2a6-e006-11eb-ba80-0242ac130004"}
```
"0.12" happens to be the string after ">" in the alert rule. Seems like this is executed as a shell script.
```
./tests/unit/test_endpoint_consumer.py
69:                    "expr": "process_cpu_seconds_total > 0.12",
```

rbarry82:
```python
result = subprocess.run(" ".join(cmd), check=False, stdout=subprocess.PIPE, shell=True)
```

**shell=True**

🚨🚨 wait until it's passed a rule with `expr: pass; rm -rf /` 🚨🚨

PR #225, opened 2022-02-02
verify and solve #223 with unit tests and disabling shell mode

simskij:
solves #223

### DIFF
--- a/lib/charms/prometheus_k8s/v0/prometheus_remote_write.py
+++ b/lib/charms/prometheus_k8s/v0/prometheus_remote_write.py
@@ -1095,7 +1095,7 @@ class PromqlTransformer:
             return expression
         args = [str(self.path)]
         args.extend(
-            ['--label-matcher {}="{}"'.format(key, value) for key, value in topology.items()]
+            ["--label-matcher={}={}".format(key, value) for key, value in topology.items()]
         )
 
         args.extend(["{}".format(expression)])
@@ -1104,8 +1104,6 @@ class PromqlTransformer:
             return self._exec(args)
         except Exception as e:
             logger.debug('Applying the expression failed: "{}", falling back to the original', e)
-            raise e
-
             return expression
 
     def _get_transformer_path(self) -> Optional[Path]:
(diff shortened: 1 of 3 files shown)


sed-i (approved):

lib/charms/prometheus_k8s/v0/prometheus_remote_write.py
```diff
@@ -1104,8 +1104,6 @@ def _apply_label_matcher(self, expression, topology):
             return self._exec(args)
         except Exception as e:
             logger.debug('Applying the expression failed: "{}", falling back to the original', e)
```
Abuelodelanada, line 1106: Is it possible to catch a more specific exception?

rbarry82 (approved):

lib/charms/prometheus_k8s/v0/prometheus_scrape.py
```diff
@@ -2222,8 +2222,6 @@ def _apply_label_matcher(self, expression, topology):
             return self._exec(args)
         except Exception as e:
             logger.debug('Applying the expression failed: "{}", falling back to the original', e)
```
Abuelodelanada, line 2224: Is it possible to catch a more specific exception?

lib/charms/prometheus_k8s/v0/prometheus_remote_write.py
simskij: good question! regardless of what happened, we want it to catch and return the original, so I think it's appropriate in this case.

lib/charms/prometheus_k8s/v0/prometheus_scrape.py
simskij: see previous comment

lib/charms/prometheus_k8s/v0/prometheus_remote_write.py
sed-i: `_exec` can raise the combination of `subprocess.run` and `result.stdout.decode` which is quite a bit, and I guess it doesn't matter why it failed.

rbarry82: `decode` is really unlikely to fail unless the LOCALE has been changed inside a pod somehow.

```python
try:
  subprocess.run(...
  # do something
```

Is probably a sufficient guard. I'm more surprised that the linters aren't yelling about the bare `Exception`

simskij:
```
# noinspection PyBroadException
```

they would have, but I've suppressed them for that part as it, as I said, is intentional. 👍🏽

PR #229, opened 2022-02-12
Catch uncaught exception

### LINKED ISSUE #228: Spurious logs in Prometheus
Sometimes we get:

```
unit-prometheus-1: 10:55:29 ERROR unit.prometheus/1.juju-log metrics-endpoint:14: Uncaught exception while in charm code:
Traceback (most recent call last):
  File "/var/lib/juju/agents/unit-prometheus-1/charm/venv/urllib3/connectionpool.py", line 449, in _make_request
    six.raise_from(e, None)
  File "<string>", line 3, in raise_from
  File "/var/lib/juju/agents/unit-prometheus-1/charm/venv/urllib3/connectionpool.py", line 444, in _make_request
    httplib_response = conn.getresponse()
  File "/usr/lib/python3.8/http/client.py", line 1347, in getresponse
    response.begin()
  File "/usr/lib/python3.8/http/client.py", line 307, in begin
    version, status, reason = self._read_status()
  File "/usr/lib/python3.8/http/client.py", line 268, in _read_status
    line = str(self.fp.readline(_MAXLINE + 1), "iso-8859-1")
  File "/usr/lib/python3.8/socket.py", line 669, in readinto
    return self._sock.recv_into(b)
socket.timeout: timed out

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/var/lib/juju/agents/unit-prometheus-1/charm/venv/requests/adapters.py", line 440, in send
    resp = conn.urlopen(
  File "/var/lib/juju/agents/unit-prometheus-1/charm/venv/urllib3/connectionpool.py", line 785, in urlopen
    retries = retries.increment(
  File "/var/lib/juju/agents/unit-prometheus-1/charm/venv/urllib3/util/retry.py", line 550, in increment
    raise six.reraise(type(error), error, _stacktrace)
  File "/var/lib/juju/agents/unit-prometheus-1/charm/venv/urllib3/packages/six.py", line 770, in reraise
    raise value
  File "/var/lib/juju/agents/unit-prometheus-1/charm/venv/urllib3/connectionpool.py", line 703, in urlopen
    httplib_response = self._make_request(
  File "/var/lib/juju/agents/unit-prometheus-1/charm/venv/urllib3/connectionpool.py", line 451, in _make_request
    self._raise_timeout(err=e, url=url, timeout_value=read_timeout)
  File "/var/lib/juju/agents/unit-prometheus-1/charm/venv/urllib3/connectionpool.py", line 340, in _raise_timeout
    raise ReadTimeoutError(
urllib3.exceptions.ReadTimeoutError: HTTPConnectionPool(host='localhost', port=9090): Read timed out. (read timeout=2.0)

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "./src/charm.py", line 378, in <module>
    main(PrometheusCharm)
  File "/var/lib/juju/agents/unit-prometheus-1/charm/venv/ops/main.py", line 426, in main
    _emit_charm_event(charm, dispatcher.event_name)
  File "/var/lib/juju/agents/unit-prometheus-1/charm/venv/ops/main.py", line 142, in _emit_charm_event
    event_to_emit.emit(*args, **kwargs)
  File "/var/lib/juju/agents/unit-prometheus-1/charm/venv/ops/framework.py", line 276, in emit
    framework._emit(event)
  File "/var/lib/juju/agents/unit-prometheus-1/charm/venv/ops/framework.py", line 736, in _emit
    self._reemit(event_path)
  File "/var/lib/juju/agents/unit-prometheus-1/charm/venv/ops/framework.py", line 783, in _reemit
    custom_handler(event)
  File "/var/lib/juju/agents/unit-prometheus-1/charm/lib/charms/prometheus_k8s/v0/prometheus_scrape.py", line 976, in _on_metrics_provider_relation_changed
    self.on.targets_changed.emit(relation_id=rel_id)
  File "/var/lib/juju/agents/unit-prometheus-1/charm/venv/ops/framework.py", line 276, in emit
    framework._emit(event)
  File "/var/lib/juju/agents/unit-prometheus-1/charm/venv/ops/framework.py", line 736, in _emit
    self._reemit(event_path)
  File "/var/lib/juju/agents/unit-prometheus-1/charm/venv/ops/framework.py", line 783, in _reemit
    custom_handler(event)
  File "./src/charm.py", line 134, in _configure
    reloaded = self._prometheus_server.reload_configuration()
  File "/var/lib/juju/agents/unit-prometheus-1/charm/src/prometheus_server.py", line 38, in reload_configuration
    response = post(url, timeout=self.api_timeout)
  File "/var/lib/juju/agents/unit-prometheus-1/charm/venv/requests/api.py", line 117, in post
    return request('post', url, data=data, json=json, **kwargs)
  File "/var/lib/juju/agents/unit-prometheus-1/charm/venv/requests/api.py", line 61, in request
    return session.request(method=method, url=url, **kwargs)
  File "/var/lib/juju/agents/unit-prometheus-1/charm/venv/requests/sessions.py", line 529, in request
    resp = self.send(prep, **send_kwargs)
  File "/var/lib/juju/agents/unit-prometheus-1/charm/venv/requests/sessions.py", line 645, in send
    r = adapter.send(request, **kwargs)
  File "/var/lib/juju/agents/unit-prometheus-1/charm/venv/requests/adapters.py", line 532, in send
    raise ReadTimeout(e, request=request)
requests.exceptions.ReadTimeout: HTTPConnectionPool(host='localhost', port=9090): Read timed out. (read timeout=2.0)
unit-prometheus-1: 10:55:29 ERROR juju.worker.uniter.operation hook "metrics-endpoint-relation-changed" (via hook dispatching script: dispatch) failed: exit status 1
```

The problem is that we are not catching the `requests.exceptions.ReadTimeout` exception in the method `reload_configuration` from `prometheus_server.Prometheus`.

As far as I understand there is another aspect that can also be improved in this method:

```python
    def reload_configuration(self) -> bool:
        """Send a POST request to to hot-reload the config.
        This reduces down-time compared to restarting the service.
        Returns:
          True if reload succeeded (returned 200 OK); False otherwise.
        """
        url = f"{self.base_url}/-/reload"
        try:
            response = post(url, timeout=self.api_timeout)

            if response.status_code == 200:
                return True
        except (ConnectionError, ConnectTimeout) as e:
            logger.debug("config reload error via %s: %s", url, str(e))

        return False
```
If Prometheus is under heavy load it is possible that the request will insume more time than `self.api_timeout`, but this do not imply that Prometheus is down... it only implies that is slow.
We should implement a `reload_configuration` method that reties the request.

Abuelodelanada:
This PR fixes #228 

The `ReadTiemout` exception was not catched

---
~~To reload Prometheus configuration we were using `response = post(url, timeout=self.api_timeout)`. This is nice and simple but the problem is that networks are unreliable or even Prometheus is under heavy load and cannot respond before `self.api_timeout`.
So it's a good idea to wrap these network calls with retries to avoid "false negatives" .~~ 

~~Now we are using the `Session` and `Retry` objects from the `request` module, with the following [retry setup](https://urllib3.readthedocs.io/en/latest/reference/urllib3.util.html?highlight=Retry#urllib3.util.Retry):~~

~~- retries: 5~~
~~- backoff_factor: 0.1~~
  ~~- _A backoff factor to apply between attempts after the second try (most errors are resolved immediately by a second try without a delay). urllib3 will sleep for:_ `{backoff factor} * (2 ** ({number of total retries} - 1))`~~
  
  ~~Some good examples of this implementation can be found [here](https://stackoverflow.com/questions/23267409/how-to-implement-retry-mechanism-into-python-requests-library),  [here](https://stackoverflow.com/questions/58724052/urllib3-how-to-get-response-when-maxretryerror-is-thrown), and [here](https://www.peterbe.com/plog/best-practice-with-retries-with-requests)~~

### DIFF
--- a/src/prometheus_server.py
+++ b/src/prometheus_server.py
@@ -37,7 +37,9 @@ class Prometheus:
           True if reload succeeded (returned 200 OK); False otherwise.
         """
         url = f"{self.base_url}/-/reload"
-        errors = list(range(400, 452)) + list(range(500, 512))
+        # http status codes see:
+        # https://www.iana.org/assignments/http-status-codes/http-status-codes.xhtml
+        http_errors_codes = list(range(400, 452)) + list(range(500, 512))
         retries = 5
         try:
             s = Session()
@@ -46,7 +48,7 @@ class Prometheus:
                 read=retries,
                 connect=retries,
                 backoff_factor=0.1,
-                status_forcelist=errors,
+                status_forcelist=http_errors_codes,
             )
             s.mount("http://", HTTPAdapter(max_retries=retry))
             response = s.post(url)
--- a/src/prometheus_server.py
+++ b/src/prometheus_server.py
@@ -5,11 +5,8 @@
 
 import logging
 
-from requests import Session, get
-from requests.adapters import HTTPAdapter
-from requests.exceptions import ConnectionError, ConnectTimeout, HTTPError, ReadTimeout
-from requests.packages.urllib3.exceptions import MaxRetryError  # type: ignore
-from requests.packages.urllib3.util.retry import Retry  # type: ignore
+from requests import get, post
+from requests.exceptions import ConnectionError, ConnectTimeout, ReadTimeout
 
 logger = logging.getLogger(__name__)
 
@@ -37,26 +34,12 @@ class Prometheus:
           True if reload succeeded (returned 200 OK); False otherwise.
         """
         url = f"{self.base_url}/-/reload"
-        # http status codes see:
-        # https://www.iana.org/assignments/http-status-codes/http-status-codes.xhtml
-        http_errors_codes = list(range(400, 452)) + list(range(500, 512))
-        retries = 5
         try:
-            s = Session()
-            retry = Retry(
-                total=retries,
-                read=retries,
-                connect=retries,
-                backoff_factor=0.1,
-                status_forcelist=http_errors_codes,
-            )
-            s.mount("http://", HTTPAdapter(max_retries=retry))
-            response = s.post(url)
-            response.raise_for_status()
+            response = post(url, timeout=self.api_timeout)
 
             if response.status_code == 200:
                 return True
-        except (ConnectionError, ConnectTimeout, ReadTimeout, HTTPError, MaxRetryError) as e:
+        except (ConnectionError, ConnectTimeout, ReadTimeout) as e:
             logger.error("config reload error via %s: %s", url, str(e))
 
         return False
--- a/src/prometheus_server.py
+++ b/src/prometheus_server.py
@@ -5,8 +5,11 @@
 
 import logging
 
-from requests import get, post
-from requests.exceptions import ConnectionError, ConnectTimeout
+from requests import Session, get
+from requests.adapters import HTTPAdapter
+from requests.exceptions import ConnectionError, ConnectTimeout, HTTPError, ReadTimeout
+from requests.packages.urllib3.exceptions import MaxRetryError  # type: ignore
+from requests.packages.urllib3.util.retry import Retry  # type: ignore
 
 logger = logging.getLogger(__name__)
 
@@ -34,13 +37,25 @@ class Prometheus:
           True if reload succeeded (returned 200 OK); False otherwise.
         """
         url = f"{self.base_url}/-/reload"
+        errors = list(range(400, 452)) + list(range(500, 512))
+        retries = 5
         try:
-            response = post(url, timeout=self.api_timeout)
+            s = Session()
+            retry = Retry(
+                total=retries,
+                read=retries,
+                connect=retries,
+                backoff_factor=0.1,
+                status_forcelist=errors,
+            )
+            s.mount("http://", HTTPAdapter(max_retries=retry))
+            response = s.post(url)
+            response.raise_for_status()
 
             if response.status_code == 200:
                 return True
-        except (ConnectionError, ConnectTimeout) as e:
-            logger.debug("config reload error via %s: %s", url, str(e))
+        except (ConnectionError, ConnectTimeout, ReadTimeout, HTTPError, MaxRetryError) as e:
+            logger.error("config reload error via %s: %s", url, str(e))
 
         return False
 


src/prometheus_server.py
```diff
@@ -34,13 +37,25 @@ def reload_configuration(self) -> bool:
           True if reload succeeded (returned 200 OK); False otherwise.
         """
         url = f"{self.base_url}/-/reload"
```
sed-i, line 39: iirc, @mmanciop encourages to use SIGHUP (container.exec kill -HUP?) instead of http api, for DoS reasons.

sed-i (approved):

src/prometheus_server.py
mmanciop: Not really DoS, but it is less complex. Although it is tricky to validate whether it worked.

rbarry82: Well, I did actually write that Pebble log streamer…

mmanciop: If we can detect the outcome of the reload very reliably with it, it is worth considering.

src/prometheus_server.py
```diff
@@ -5,8 +5,11 @@
 
 import logging
 
-from requests import get, post
-from requests.exceptions import ConnectionError, ConnectTimeout
+from requests import Session, get
```
dstathis, line 8: Is get used?

dstathis (approved):

src/prometheus_server.py
Abuelodelanada: Yes, in line 74:

```python
        try:
            response = get(url, timeout=self.api_timeout)
```

src/prometheus_server.py
Abuelodelanada: > If we can detect the outcome of the reload very reliably with it, it is worth considering.

Is this possible @rbarry82 ? If so, do we have a charm implementing this?

src/prometheus_server.py
```diff
@@ -34,13 +37,25 @@ def reload_configuration(self) -> bool:
           True if reload succeeded (returned 200 OK); False otherwise.
         """
         url = f"{self.base_url}/-/reload"
+        errors = list(range(400, 452)) + list(range(500, 512))
```
balbirthomas, line 40: These HTTP status codes feel like magic numbers. Maybe a comment here would be worthwhile. Perhaps a more descriptive variable names such as `http_errors` may would also be more revealing.

src/prometheus_server.py
```diff
@@ -34,13 +37,25 @@ def reload_configuration(self) -> bool:
...
+            )
+            s.mount("http://", HTTPAdapter(max_retries=retry))
+            response = s.post(url)
+            response.raise_for_status()
 
             if response.status_code == 200:
                 return True
-        except (ConnectionError, ConnectTimeout) as e:
-            logger.debug("config reload error via %s: %s", url, str(e))
+        except (ConnectionError, ConnectTimeout, ReadTimeout, HTTPError, MaxRetryError) as e:
```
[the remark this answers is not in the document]
Abuelodelanada: The main issue we are trying to fix (#228) is that `ReadTimeout` exception is not catched. Based on that we are also improving the original implementation

It is true that trying to reload the configuration several times does not completely eliminate the possibility of the configuration reload failing.
However, if those attempts are made at increasing time intervals, I think we can be pretty sure that if we keep getting ConnectionError, ConnectTimeout, ReadTimeout, HTTPError exceptions, it means that there is a problem with the API.

We could have explored the possibility of sending a SIGHUP signal, but that would have led to a bigger refactor and I think it's beyond the scope of the issue we're trying to solve here.

sed-i (dismissed):

src/prometheus_server.py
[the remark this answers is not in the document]
balbirthomas: If the goal as you say is to catch all possible exceptions then I do not see why add the additional logic to retry. I would have expected if reloading of Prometheus configuration fails then the Charm must be informed and it (the charm) should take the appropriate action. Is that already happening ? If this is not already happening, isn't this the more critical bug that needs to be fixed.

balbirthomas (approved):

sed-i (approved):

PR #230, opened 2022-02-14
Integrate ingress-unit with Prometheus

mmanciop:
This commit introduces support for the ingress-unit capabilities provided by the [Traefik Ingress charm]((https://github.com/canonical/traefik-k8s-operator) as of the https://github.com/canonical/traefik-k8s-operator/pull/3 PR.

This change also removes `web.external-url` as valid config option, which becomes redundant as the external URL is going to be provided by the ingress charm over the relation.


mmanciop:
I am thinking, actually, that we should not break backwards compatibility with `web.external-url`, but rather put the Prometheus charm in blocked state if both that option and the `ingress` relation are present.

Issue #188, opened 2021-12-15
MetricsEndpointConsumer does not support `upgrade_charm`


mmanciop:
The `MetricsEndpointConsumer` does not have support for the `upgrade_charm` event, which leads to Prometheus not noticing the scrape job and alert rules changes when developing in local and updating Prometheus (e.g., for bug fixes in Prometheus operator itself).

I am unsure whether the bug would trigger in production, e.g., when the release changes the Prometheus container image and churns the pods. I _think_ in that case we will get `relation_joined` and it would work, but it is not a scenario we test in integration tests atm.

simskij:
Test if there **is** still a bug, if there is a bug; we need a test that verifies it and then fix the bug.

balbirthomas:
I am unable to reproduce this bug hence closing it. Also it is the `MetricsEndpointProvider` that updates alert rules on upgrade charm.

PR #232, opened 2022-02-25
Added a MetricsEndpointProvider upgrade integration test

balbirthomas:
This commit adds and integration test to check the upgrade
workflow for MetricsEndpoint Provider.

This commit also fixes a bug in the integration test helper
method "get_rules_for". Previously each MetricsEndpointProvider
charm only submitted a single alert rule group. However the
logic of gathering alert rules by MetricsEndpointProvider
was refactored to allow returing more than one group. This
refactor failed to refactor the "get_rules_for" method which
only returned the first matching rule.


tests/integration/helpers.py
```diff
@@ -187,6 +191,43 @@ def initial_workload_is_ready(ops_test, app_names) -> bool:
...
+    Args:
+        rule: a string containing Prometheus alert rule in YAML format.
+        name: a string name of alert rule file
+    """
+    rules_path = Path(TESTER_ALERT_RULES_PATH).joinpath(name)
+    with rules_path.open(mode="w") as f:
+        f.write(rule)
+
+
+def remove_tester_alert_rule_file(name):
```
sed-i, line 206: Can we avoid modifying the source tree at all?
If not, maybe add this to auto cleanup?

tests/integration/test_upgrade_charm.py
```diff
@@ -28,12 +52,77 @@ async def test_deploy_from_edge_and_upgrade_from_local_path(ops_test, prometheus
...
+
+async def test_metrics_endpoint_provider_can_be_upgraded(ops_test, prometheus_tester_charm):
+    """Ensure scrape alert rules can be updated.
+
+    This test upgrades the metrics provider charm and checks that
+    updates to alert rules are propagated correctly.
+    """
+    await ops_test.model.deploy(
+        prometheus_tester_charm, resources=tester_resources, application_name=TESTER_APP_NAME
+    )
```
sed-i, line 80: I think the tester should be deployed and related to prom _before_ prom gets refreshed from local path (line 63).

tests/integration/test_upgrade_charm.py
```diff
@@ -28,12 +52,77 @@ async def test_deploy_from_edge_and_upgrade_from_local_path(ops_test, prometheus
...
+        except RetryError:
+            pass
+
+    # Add new alert rule, rebuild and refresh prometheus tester charm
+    write_tester_alert_rule_file(MISSING_TARGET_RULE, "target_missing.rule")
+    tester_charm = await rebuild_prometheus_tester(ops_test)
+    await ops_test.model.applications[TESTER_APP_NAME].refresh(
+        path=tester_charm, resources=tester_resources
+    )
+    remove_tester_alert_rule_file("target_missing.rule")
```
sed-i, line 111: Since the extra rule is needed only for packing, maybe create a context manager for it?
```python
with write_tester_alert_rules(...) as tester_charm:
    # ...
```

tests/integration/helpers.py
balbirthomas: I do not see how I can inject an alert rule through the mechanism provided by the Prometheus charm (`promtheus_alert_rules` folder) without "modifying the source" tree and without more complicated code logic.

tests/integration/test_upgrade_charm.py
balbirthomas: The purpose of the new test is check that MetricsEndpointProvider upgrade path works since the bug described in the ticket was that new alert rules were not being set after upgrade. If modifications are required to the previously existing test for MetricsEndpointConsumer then we can create a separate ticket and modify as part of that ticket.

tests/integration/test_upgrade_charm.py
balbirthomas: I find it simpler to just provide explicit methods to write and remove the file instead of an "enter" and "exit" method. A context manager would be useful if we need to write multiple files in different contexts, for which I do not see a use case at present. Also I find the phrase "with write_tester_alert_rules() as tester_charm" odd to read.

tests/integration/test_upgrade_charm.py
sed-i: I was not aware of that. Can you please add the bug description or crossref to the ticket?

balbirthomas: In meeting with @mmanciop the discussion concluded it would be useful to have a integration test to check that alert rules are updated on charm upgrade and that the alert rules check was sufficient (i.e without scrape job check) . The alert rules are updated by `MetricsEndpointProvider` **not** `MetricsEndpointConsumer`. The only issue on github was #188, which I have now clarified and closed.

balbirthomas: Thanks for your time @sed-i . As discussed I have
- Created #234 to enhance the existing `MetricsEnpointConsumer` test in the next commit.
- On refactoring the `MetricsEndpointConsumer` test will be renamed to emphasise it is testing the Consumer charm.
- Also as discussed I will retain both tests in the same module `test_upgrade_charm.py` and document that this module test upgrade of both Provider and Consumer charm.

tests/integration/test_upgrade_charm.py
balbirthomas: Resolving as discussed.

tests/integration/test_upgrade_charm.py
```diff
@@ -28,12 +52,77 @@ async def test_deploy_from_edge_and_upgrade_from_local_path(ops_test, prometheus
...
         )
-        await ops_test.model.wait_for_idle(apps=[app_name], status="active", timeout=TIMEOUT)
-        await check_prometheus_is_ready(ops_test, app_name, 0)
+        await ops_test.model.wait_for_idle(
+            apps=[PROMETHEUS_APP_NAME], status="active", timeout=TIMEOUT
+        )
+        await check_prometheus_is_ready(ops_test, PROMETHEUS_APP_NAME, 0)
+
+
+async def test_metrics_endpoint_provider_can_be_upgraded(ops_test, prometheus_tester_charm):
```
sed-i, line 72:
```suggestion
async def test_upgrading_rules_provider_also_updates_rule_files(ops_test, prometheus_tester_charm):
```

sed-i (dismissed):

simskij (dismissed):

tests/integration/test_upgrade_charm.py
balbirthomas: I am fine with changing this title. I just don't want to trigger and test run and wait for re-reviews. I hope it is ok if I just merge as is now and will change the title as part of refactor of other test. Feel free to remind me if I fail to do so.

simskij (approved):

dstathis (approved):

PR #239, opened 2022-03-04
Remove unit from alerts

dstathis:


tests/unit/test_endpoint_provider.py
```diff
@@ -411,6 +411,14 @@ def test_alerts_in_both_formats_are_recursively_aggregated(self):
 
         self.assertEqual({}, DeepDiff(expected_rules_file, rules_file_dict, ignore_order=True))
 
+    def test_unit_not_in_alert_labels(self):
```
balbirthomas, line 414: I would have expected a couple of more unit tests here, because the behaviours we are concerned with are

- Does `MetricsEndpointProvider` **not add** `juju_unit` to alert rules expressions and labels ?
- Does `MetricsEndpointProvider` **retain** `juju_unit` if the are present in alert rule expressions and labels ?
- Does `MetricsEndpointAggregator` **not add** `juju_unit` to alert rules expressions and labels ?
- Does `MetricsEndpointAggregator` **retain** `juju_unit` if the are present in alert rule expressions and labels ?

balbirthomas:
@dstathis It would help in review and for posterity if your *commit message* included
- A description of the bug
- An explanation why this commit is the right fix for the problem.

rbarry82 (dismissed):

tests/unit/test_endpoint_provider.py
dstathis: I'm not sure if this will strip out existing unit tags. @mmanciop What is the expected behavior here?

lib/charms/prometheus_k8s/v0/prometheus_scrape.py
```diff
@@ -634,6 +633,9 @@ def as_promql_label_dict(self):
             "juju_{}".format(key): val
             for key, val in self.as_dict(rename_keys={"charm_name": "charm"}).items()
         }
+        # unit should never be specified in alert rules as the apply to the entire application
```
sed-i, line 636: Typo in the comment?
Can you elaborate more?

lib/charms/prometheus_k8s/v0/prometheus_scrape.py
```diff
@@ -585,16 +585,15 @@ def from_relation_data(cls, data: dict):
     def identifier(self) -> str:
         """Format the topology information into a terse string."""
         # This is odd, but may have `None` as a model key
-        return "_".join([str(val) for val in self.as_dict().values()]).replace("/", "_")
+        return "_".join([str(val) for val in self.as_promql_label_dict().values()]).replace(
+            "/", "_"
+        )
```
sed-i, line 590: Breaking change for `identifier`?

lib/charms/prometheus_k8s/v0/prometheus_scrape.py
```diff
@@ -585,16 +585,15 @@ def from_relation_data(cls, data: dict):
...
 
     @property
     def promql_labels(self) -> str:
         """Format the topology information into a verbose string."""
         return ", ".join(
-            [
-                'juju_{}="{}"'.format(key, value)
-                for key, value in self.as_dict(rename_keys={"charm_name": "charm"}).items()
-            ]
+            ['{}="{}"'.format(key, value) for key, value in self.as_promql_label_dict().items()]
```
sed-i, line 596: This looks significantly different. Is this a breaking change?

lib/charms/prometheus_k8s/v0/prometheus_scrape.py
```diff
@@ -683,9 +685,7 @@ def scrape_identifier(self):
         """Format the topology information into a scrape identifier."""
         # This is used only by Metrics[Consumer|Provider] and does not need a
         # unit name, so only check for the charm name
-        return "juju_{}_prometheus_scrape".format(
-            "_".join([self.model, self.model_uuid[:7], self.application, self.charm_name])  # type: ignore
-        )
+        return "juju_{}_prometheus_scrape".format(self.identifier)
```
sed-i, line 688: This looks significantly different. Is this a breaking change?

lib/charms/prometheus_k8s/v0/prometheus_scrape.py
mmanciop: The problem is that by setting the unit in the alert rule label, we are also making the PromQL transformer add it to the filters it injects in the expression. Since the leader sets the alerts in the relation, the unit marked in the alert rules with the current logic is always the leader's. And since we filter the expression in the alert rules using the `juju_unit="<leader_unit_name>"`, the alert rules are evaluated ONLY for the leader.

lib/charms/prometheus_k8s/v0/prometheus_scrape.py
mmanciop: @sed-i if you are asking this question, we may not have sufficient unit tests. Care to validate that?

tests/unit/test_endpoint_provider.py
mmanciop: If the alert rule does come with unit tags (which might be needed for some corner cases of the cos-configuration-repository charm), we need to keep that label.

lib/charms/prometheus_k8s/v0/prometheus_scrape.py
dstathis: It should be the same except without unit.

lib/charms/prometheus_k8s/v0/prometheus_scrape.py
dstathis: It's the same except without unit. `as_promql_label_dict` applies `"juju_" to the values.

lib/charms/prometheus_k8s/v0/prometheus_scrape.py
dstathis: I believe it does not truncate the UUID. @mmanciop would you like to continue truncating the UUID? Otherwise the change should not break anything that I am aware of.

tests/unit/test_endpoint_provider.py
dstathis: A unit test has been added for the case of existing labels

lib/charms/prometheus_k8s/v0/prometheus_scrape.py
mmanciop: If we truncate the id, we need to use a different matcher (~=) in promql-transform

mmanciop: I am actually curious to know whether truncating the model_uuid would break the itests. @simskij ?

dstathis: @mmanciop I believe this only affects the file name.

tests/unit/test_endpoint_provider.py
```diff
@@ -544,3 +553,141 @@ def test_deeply_nested(self):
...
+        self.harness.set_leader(True)
+        rel_id = self.harness.add_relation("metrics-endpoint", "prom")
+        self.harness.add_relation_unit(rel_id, "prom/0")
+        self.harness.begin_with_initial_hooks()
+
+        # THEN relation data does not contain juju_unit in expr or labels
+        relation = self.harness.charm.model.get_relation("metrics-endpoint")
+        alert_rules = json.loads(relation.data[self.harness.charm.app].get("alert_rules"))
+        for group in alert_rules["groups"]:
+            with self.subTest(group=group["name"]):
```
balbirthomas, line 621: This is rather deeply nested code. The 4 level nesting is a result of using `subTests`. I struggle to see the value of using `subTest` here. If test conditions fail for any group this test should just fail.

lib/charms/prometheus_k8s/v0/prometheus_scrape.py
balbirthomas: I think we should eliminate work towards eliminating the proliferation of methods `promql_labels`, `as_promql_labels_dict` and retain only a single method `as_dict`. Hence it is may be a better idea to do something like
```
identifiers = topology.as_dict()
identifiers.pop('juju_unit')
return "_".join([str(val) for val in identifiers.values()]).replace("/", "_")
```

lib/charms/prometheus_k8s/v0/prometheus_scrape.py
balbirthomas: I think the right think to do here is eliminate the `scrape_identifier()` method using only the `topology.identifier()` method and ensuring that it truncates UUID.

tests/unit/test_endpoint_provider.py
```diff
@@ -544,3 +553,141 @@ def test_deeply_nested(self):
             ]
         }
         self.assertDictEqual(expected_rules_file, rules_file_dict_read)
+
+
+def provider_with_rules_factory(rules: Dict[str, str]):
```
balbirthomas, line 558: We now have multiple ways of creating customized test Provider Charms in our unit tests

On one hand we have two four line methods that use alert rules from file 
- `class CustomizableEndpointProviderCharm`
- `def customize_endpoint_provider`

On the other hand we have 
- The `helper.py` module that offers the `TempFolderSandbox` which writes alert rules (taken from this unittest script) into files on disk
- The `def provider_with_rules_factory()` method that uses `TempFolderSandbox` class to create alert rules and instantiate a Provider charm.

At least within one charm's code base should we be consistent in the way we write the unit tests ? Do you think there is value in doing this ? Would it make our code easier to understand ?

tests/unit/test_endpoint_provider.py
```diff
@@ -544,3 +553,141 @@ def test_deeply_nested(self):
...
+        provider_charm = provider_with_rules_factory(
+            {
+                "with_topology.rule": yaml.safe_dump(
+                    {"alert": "with_topology", "expr": "up{%%juju_topology%%} == 0"}
+                )
+            }
+        )
+
+        # WHEN relation data is generated
+        self.harness = Harness(provider_charm, meta=provider_charm.metadata_yaml)
```
balbirthomas, line 611: It is unusual to instantiate the Harness in each test of the same `unittest` class. By departing from standard convention we make our tests that bit harder to understand.

tests/unit/test_endpoint_provider.py
```diff
@@ -544,3 +553,141 @@ def test_deeply_nested(self):
...
+                self, jobs=JOBS, alert_rules_path=alert_rules_path
+            )
+
+    return ProviderCharm
+
+
+class TestAlertRulesDoNotContainUnitNumber(unittest.TestCase):
+    """Feature: Consumer lib must not add a unit number to rule filters or labels.
+
+    Background: In lieu of SDI or pydantic, schema verification is done in unit tests.
```
balbirthomas, line 592: I do not see the significance of this comment. As I understand our tests should not be trying to validate the "schema" of the alert rule. The tests instead need to validate to alternative conditions
- If `juju_unit` is present in an alert label or expression it is not removed
- If `juju_unit` is not present in an alert label or expression it is not added.

This has nothing to do with "schema verification" i.e. we are not trying to validate the syntax of alert rules, their expressions or their labels.

tests/unit/test_endpoint_provider.py
```diff
@@ -544,3 +553,141 @@ def test_deeply_nested(self):
...
+    """
+
+    def setUp(self):
+        pass
+
+    @patch_network_get(private_address="192.0.8.2")
+    def test_alert_rules_with_topology(self):
```
balbirthomas, line 599: Why do we need a separate test case here. There is already a test case to check that each alert rule is topology labeled `test_each_alert_rule_is_topology_labeled()`. Could this existing test case be extended to add two additional checks
```
      self.assertNotIn("juju_unit", rule["labels"])
      self.assertNotIn("juju_unit=", rule["expr"])
```

tests/unit/test_endpoint_provider.py
```diff
@@ -544,3 +553,141 @@ def test_deeply_nested(self):
...
+
+                        # These two are to prevent false negatives in case something big changes
+                        self.assertIn("juju_application", rule["labels"])
+                        self.assertIn("juju_application=", rule["expr"])
+
+        self.harness.cleanup()
+
+    @patch_network_get(private_address="192.0.8.2")
+    def test_alert_rules_without_topology(self):
+        """Scenario: Free standing rules shouldn't have any topology whatsoever, including unit."""
```
balbirthomas, line 635: This is news to me : "Free standing rules shouldn't have any topology whatsoever" ?  Isn't Juju Topology injected into **all** alert rules expressions and labels ?

tests/unit/test_endpoint_provider.py
```diff
@@ -544,3 +553,141 @@ def test_deeply_nested(self):
...
+            with self.subTest(group=group["name"]):
+                for rule in group["rules"]:
+                    with self.subTest(rule=rule["alert"]):
+                        self.assertNotIn("juju_unit", rule["labels"])
+                        self.assertNotIn("juju_unit=", rule["expr"])
+
+        self.harness.cleanup()
+
+    @patch_network_get(private_address="192.0.8.2")
+    def test_alert_rules_with_hard_coded_unit_label(self):
```
balbirthomas, line 661: If the test name suggested what was being checked for it makes the code self documenting. For example 
` def test_endpoint_provider_does_not_overwrite_hard_coded_unit_label` . The name `test_alert_rules_with_hard_coded_unit_label` only says this test has "some thing" to do with "hard coded unit" labels but does not say what.

tests/unit/test_endpoint_provider.py
```diff
@@ -544,3 +553,141 @@ def test_deeply_nested(self):
...
+                self.sandbox.put_file(os.path.join(alert_rules_path, filepath), contents)
+
+            self.provider = MetricsEndpointProvider(
+                self, jobs=JOBS, alert_rules_path=alert_rules_path
+            )
+
+    return ProviderCharm
+
+
+class TestAlertRulesDoNotContainUnitNumber(unittest.TestCase):
```
balbirthomas, line 589: Why do we need a separate test class here ?

tests/unit/test_endpoint_provider.py
sed-i: Correct, this is intended for "generic" rules that are not associated with any charm in particular.

tests/unit/test_endpoint_provider.py
sed-i: Agreed.
The `endpoint_provider` prefix might be redundant given the filename.

How about:
```suggestion
    def test_unit_label_is_retained_if_hard_coded(self):
```

tests/unit/test_endpoint_provider.py
balbirthomas: Would you rather add two extra lines of code or add 2 dozen lines of code ? I see no issue with changing the name. Even the name as it is should be fine. The current name is just say that this test checks alert rules are labelled with topology. It does not say this test only checks alert rule labels are topology labelled.

tests/unit/test_endpoint_provider.py
balbirthomas: I do not see why the `MetricsEndpointProvider` is concerned with "free standing rules". Provider has the same behaviour regardless of where it is used isnt' it. In all cases it adds topology labels into alert rules (without the unit after this commit). Isn't this correct ? If so this comment not necessary. More importantly as mentioned above this test is redundant.

tests/unit/test_endpoint_provider.py
balbirthomas: Doesn't it make sense that unit tests of the Test Harness itself will need instantiate the test harness in the test ? I do not see how this extends to tests that use the Test Harness.

tests/unit/test_endpoint_provider.py
balbirthomas: Given the paucity of time, do you think it makes sense to make the smallest simplest change that will fix the bug, check it is fixed and deliver ? Would it make more sense to move the discussion to another refactoring ticket. ?

tests/unit/test_endpoint_provider.py
balbirthomas: As I understand we need two test cases for `MetricsEndpointProvider`

1. A test case to check that `juju_unit` is not added to alert rules and expressions by default. As mentioned above adding two lines to an existing test case meets this requirement.
2. A test case to check that if `juju_unit` is present in an alert rule label and expression then it is not overwritten. 

For 1 you should not need any new test class. For 2 you will need a separate test class, since you need a different set of alert rules who's labels and expressions do include Juju unit topology annotations.

Case 1 corresponds to the name `TestAlertRulesDoNotContainUnitNumber`.

tests/unit/test_endpoint_provider.py
sed-i: Feel free to create a refactoring ticket!

tests/unit/test_endpoint_provider.py
sed-i: I am confused. Feel free to push a commit!

tests/unit/test_endpoint_provider.py
sed-i: Gonna leave this for the OP, @dstathis.

tests/unit/test_endpoint_provider.py
sed-i: Gonna leave this for the OP, @dstathis.

balbirthomas:
For whatever it is worth. Here is my analysis of the impacts of the changes introduced by this commit. I am still going through to double check these so take it with a pinch of salt.

1. `JujuTopology.as_promql_label_dict()` removes (`pop("juju_unit"`) juju unit
    topology information. `as_promql_label_dict()`. The impacts of this change are

- `JujuTopology.identifer()` uses `as_promql_label_dict()` - for these impacts
  see point 2 below.

- `JujuTopology.promql_labels()` uses `as_promql_label_dict()` - for these impacts
  see point 3 below.

2. The `JujuTopology.identifer()` method previously used `as_dict()`. Now it
   uses it uses `as_promql_label_dict()`. `as_dict()` included `juju_unit` but
   `as_promql_label_dict()` does not. The impact of this change is as follows

- `ProviderTopology.scrape_identifier()` uses `JujuTopology.identifer()`.
  Previously `scrape_identifer()` included juju unit now it does not.
  `scrape_identifier()` is used by `MetricsEndpointConsumer` to set the
  prefix of the Prometheus scrape job name only for jobs forwarded by
  `MetricsEndpointProvider` (but not for jobs sent by `MetricsEndpointAggregator`)

- The `AlertRules._group_name()` uses `JujuTopology.identier()`
  to construct a name for an alert rule group. `AlertRules` can be instantiated
  with any subclass of `JujuTopology`. However `AlertRules` is only used by 
  `MetricsEndpointProvider` and `PrometheusRulesProvider` but not by
  `MetricsEndpointAggregator`.

  The alert rule group names constructed by `AlertRules` will never container
  juju unit regardless of which kind of topology (ProviderTopology or
  AggregatorTopology is used)
  
- `MetricsEndpointConsumer.alerts()` uses `JujuTopology.identifer()` to build
  and indexed list of alerts, which are indexed using the "identifier".  This index
  will no longer contain juju topology information, for alert rules from both
  MetricsEndpointProvider and MetricsEndpointAggregator.

- `MetricsEndpointConsumer._static_scrape_config()` uses `JujuTopology.identifier()` 
  for constructing the scrape job name on for jobs provided by `MetricsEndpointProvider`.
  After this commit these job names will not include juju unit topology information.

3. `JujuTopology.promql_labels()` previously previously used
   `JujuTopology.as_dict()` but now it uses `JujuTopology.as_promql_label_dict()`.
   The impact of this change is as follows.

- `JujuTopology.promql_labels()` is only used in `JujuTopology.render()` method
  which is used to inject topology information into an alert rule expression by
  the `AlertRules` object. Hence `AlertRules` object will never inject juju unit
  topology into any alert rule read by it.

4. `ProviderTopology.scrape_identifer()` now uses `JujuTopology.identier()` Since
   `ProviderTopology` previously constructed the identifer without juju unit information
   this change may not have any impacts.

lib/charms/prometheus_k8s/v0/prometheus_scrape.py
simskij: @dstathis, can you please incorporate the rationale Michele explains above into the comment? I think we'll want to be able to reference that later.

lib/charms/prometheus_k8s/v0/prometheus_scrape.py
simskij: @balbirthomas Please create a separate ticket for that work, unless you have already.

lib/charms/prometheus_k8s/v0/prometheus_scrape.py
balbirthomas: I am resolving this discussion so we can move conversation to https://github.com/canonical/prometheus-k8s-operator/issues/240

lib/charms/prometheus_k8s/v0/prometheus_scrape.py
balbirthomas: I am resolving this discussion so we can move conversation to https://github.com/canonical/prometheus-k8s-operator/issues/240

lib/charms/prometheus_k8s/v0/prometheus_scrape.py
simskij: well, they were truncated prior to the change, and with this change they no longer are. I expect us to keep the current behavior (ie. truncate at 8 chars). But also, as Michele wrote, this should break tests, or we are clearly missing some vital tests, no?

simskij (changes requested):
once tests pass and the conversations have been properly resolved, I'm happy to approve.

tests/unit/test_endpoint_provider.py
```diff
@@ -411,6 +414,14 @@ def test_alerts_in_both_formats_are_recursively_aggregated(self):
 
         self.assertEqual({}, DeepDiff(expected_rules_file, rules_file_dict, ignore_order=True))
 
+    def test_unit_not_in_alert_labels(self):
+        rules = AlertRules(topology=self.topology)
+        rules.add_path(os.path.join(self.sandbox.root, "rules", "prom"), recursive=True)
+        rules_file_dict = rules.as_dict()
+        for group in rules_file_dict["groups"]:
+            for rule in group["rules"]:
+                self.assertTrue("juju_unit" not in rule["labels"])
```
sed-i, line 423: Is this different from `test_each_alert_rule_is_topology_labeled`?
Should this also have the `expr` check?

tests/unit/test_endpoint_provider.py
```diff
@@ -544,3 +555,36 @@ def test_deeply_nested(self):
...
+        rel_id = self.harness.add_relation("metrics-endpoint", "provider")
+        self.harness.add_relation_unit(rel_id, "provider/0")
+
+        # check unit topology is present in labels but not in alert rule expression
+        relation = self.harness.charm.model.get_relation("metrics-endpoint")
+        alert_rules = json.loads(relation.data[self.harness.charm.app].get("alert_rules"))
+        for group in alert_rules["groups"]:
+            for rule in group["rules"]:
+                self.assertIn("juju_unit", rule["labels"])
+                self.assertIn("juju_unit=", rule["expr"])
```
sed-i, line 590: NotIn?

Abuelodelanada (approved):

tests/unit/test_endpoint_provider.py
```diff
@@ -201,6 +201,9 @@ def test_each_alert_rule_is_topology_labeled(self, _):
             self.assertIn("juju_application", labels)
             self.assertIn("juju_model_uuid", labels)
             self.assertIn("juju_charm", labels)
+            # alerts should not have unit information if not already present
+            self.assertNotIn("juju_unit", rule["labels"])
+            self.assertNotIn("juju_unit=", rule["expr"])
```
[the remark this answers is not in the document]
balbirthomas: Ack. Done. Should not make a big difference. Since one group should not be treated differently from another.

tests/unit/test_endpoint_provider.py
balbirthomas: This test is probably redundant and was added in @dstathis commit. @dstathis is this test still needed.

tests/unit/test_endpoint_provider.py
balbirthomas: Nope, the test is `test_unit_label_is_retained_if_hard_coded`.

sed-i: The docstring says `# check unit topology is present in labels but not in alert rule expression` and I do expect a NotIn for line 590.

balbirthomas: Thanks. Fixed now. Copy paste error from your prior tests. So see your code still remains :-).

dstathis: I don't understand how this test works. Where does it get the initial alert rule from?

tests/unit/test_endpoint_provider.py
dstathis: Yes we still need this. Your test checks the opposite case.

sed-i (approved):

tests/unit/test_endpoint_provider.py
dstathis: But I agree it should be `assertIn`

tests/unit/alert_rules_with_unit_topology/nothing_is_up.rule
```diff
@@ -0,0 +1,6 @@
+alert: AllUnitsDown
```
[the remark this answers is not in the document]
balbirthomas: How about NoActiveUnits ?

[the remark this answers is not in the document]
balbirthomas: Ignore that. Event better `Unit42Inactive` ?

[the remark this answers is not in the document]
rbarry82: Or `InactiveUnit`

tox.ini
```diff
@@ -88,7 +88,7 @@ commands =
     sh -c 'stat promql-transform > /dev/null 2>&1 || curl "https://github.com/canonical/promql-transform/releases/download/2.25.2-1rc1/promql-transform_2.25.2-1rc1_linux_amd64.tar.gz" -L -s | tar zxv promql-transform'
     coverage run \
       --source={[vars]src_path},{[vars]lib_path} \
-      -m pytest -v --tb native -s {posargs} {[vars]tst_path}/unit
+      -m pytest -v --tb native --log-cli-level=INFO -s {posargs} {[vars]tst_path}/unit
```
rbarry82, line 91: Intentional or debugging?

rbarry82 (dismissed):
Just quibbling about the alert name. Is the change to `tox.ini` intentional?

Abuelodelanada (dismissed):

tox.ini
dstathis: I believe that is intentional @sed-i ?

tests/unit/alert_rules_with_unit_topology/nothing_is_up.rule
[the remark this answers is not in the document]
balbirthomas: The rule is matching on a specific unit (42 - The answer to everything).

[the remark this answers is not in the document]
dstathis: Changed

mmanciop (approved):

tox.ini
sed-i: Yes that is intentional and in line with other charms, though obviously not necessary.

sed-i (approved):

PR #241, opened 2022-03-08
Add tox envs for deploying a local charm

sed-i:
This PR adds two parametrized tox envs for deploying charms with promql:

```shell
tox -e deploy-amd64
tox -e refresh-amd64 -- prom
```

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -10,6 +10,8 @@ src_path = {toxinidir}/src
 tst_path = {toxinidir}/tests
 lib_path = {toxinidir}/lib/charms/prometheus_k8s
 all_path = {[vars]src_path} {[vars]tst_path} {[vars]lib_path}
+promql_amd64_archive = https://github.com/canonical/promql-transform/releases/download/2.25.2-1rc1/promql-transform_2.25.2-1rc1_linux_amd64.tar.gz
+promql_arm64_archive = https://github.com/canonical/promql-transform/releases/download/2.25.2-1rc1/promql-transform_2.25.2-1rc1_linux_arm64.tar.gz
 
 [testenv]
 basepython = python3
@@ -26,6 +28,34 @@ passenv =
   HTTP_PROXY
   HTTPS_PROXY
   NO_PROXY
+
+[testenv:deploy-{amd64,arm64}]
+description = Deploy local charm with promql-transform
+allowlist_externals =
+    wget
+    tar
+    juju
+commands =
+    # Download arch-appropriate promql-transform
+    amd64: wget -O {temp_dir}/promql-transform.tar.gz {[vars]promql_amd64_archive}
+    arm64: wget -O {temp_dir}/promql-transform.tar.gz {[vars]promql_arm64_archive}
+    tar --overwrite -C {temp_dir} -xzvf {temp_dir}/promql-transform.tar.gz promql-transform
+    amd64: juju deploy --resource prometheus-image=ubuntu/prometheus:latest --resource promql-transform-amd64={temp_dir}/promql-transform ./prometheus-k8s_ubuntu-20.04-amd64.charm {posargs}
+    arm64: juju deploy --resource prometheus-image=ubuntu/prometheus:latest --resource promql-transform-arm64={temp_dir}/promql-transform ./prometheus-k8s_ubuntu-20.04-arm64.charm {posargs}
+
+[testenv:refresh-{amd64,arm64}]
+description = refresh charm from local path, with promql-transform
+allowlist_externals =
+    wget
+    tar
+    juju
+commands =
+    # Download arch-appropriate promql-transform
+    amd64: wget -O {temp_dir}/promql-transform.tar.gz {[vars]promql_amd64_archive}
+    arm64: wget -O {temp_dir}/promql-transform.tar.gz {[vars]promql_arm64_archive}
+    tar --overwrite -C {temp_dir} -xzvf promql-transform.tar.gz promql-transform
+    amd64: juju refresh --resource promql-transform-amd64={temp_dir}/promql-transform --path=./prometheus-k8s_ubuntu-20.04-amd64.charm {posargs}
+    arm64: juju refresh --resource promql-transform-arm64={temp_dir}/promql-transform --path=./prometheus-k8s_ubuntu-20.04-arm64.charm {posargs}
 
 [testenv:fmt]
 description = Apply coding style standards to code
--- a/tox.ini
+++ b/tox.ini
@@ -57,6 +57,15 @@ commands =
     amd64: juju refresh --resource promql-transform-amd64={temp_dir}/promql-transform --path=./prometheus-k8s_ubuntu-20.04-amd64.charm {posargs}
     arm64: juju refresh --resource promql-transform-arm64={temp_dir}/promql-transform --path=./prometheus-k8s_ubuntu-20.04-arm64.charm {posargs}
 
+[testenv:publish-{edge,beta}]
+description = release a previously uploaded charm to a charmhub edge/beta channel
+allowlist_externals =
+    charmcraft
+commands =
+    # Usage example: tox -e publish-edge -- --revision=19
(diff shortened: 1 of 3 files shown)


tox.ini
```diff
@@ -27,6 +29,43 @@ passenv =
...
+    amd64: juju refresh --resource promql-transform-amd64={temp_dir}/promql-transform --path=./prometheus-k8s_ubuntu-20.04-amd64.charm {posargs}
+    arm64: juju refresh --resource promql-transform-arm64={temp_dir}/promql-transform --path=./prometheus-k8s_ubuntu-20.04-arm64.charm {posargs}
+
+[testenv:publish-{edge,beta}]
+description = release a previously uploaded charm to a charmhub edge/beta channel
+allowlist_externals =
+    charmcraft
+commands =
+    # Usage example: tox -e publish-edge -- --revision=19
+    edge: charmcraft release prometheus-k8s --channel=edge --resource=prometheus-image:1 --resource=promql-transform-amd64:3 {posargs}
```
balbirthomas, line 66: Wouldn't both resource revisions need to be customizable ?

sed-i: Correct, but it's a tradeoff between:
1. resources don't get updated often so might as well hard code them until they need to change
2. make it customizable but require users to run `tox -e publish-edge -- --revision=19 --resource=prometheus-image:1 --resource=promql-transform-amd64:3`

(2) is almost the same length as than the original command though.

balbirthomas: Hmm. We could end up with the same problem then. We publish new resources and forget to updated the tox.ini so we end up making releases with old resources.

sed-i: Correct, but this is slightly better than omitting.
I am fine with dropping the publish tox env, just wanted to show how it would look like.
It has pluses and minuses obviously.

balbirthomas: True. But not my call to make.

rbarry82: The GH action to publish charm automatically fetches whatever is specified anyway.

We noticed initially because every merge to Grafana ends up putting Grafana 8 (latest) on `edge`, so this is probably less of a concern anyway. The action may be useful for manually deploying akin to a `make` target if there's an emergency fix which needs to go out while code review is still pending, though.

rbarry82:
This looks ok to me in general. As the comment, though, I'm not sure how useful it will be in the end considering the GH action, and then it will just be dead code.

sed-i:
> This looks ok to me in general. As the comment, though, I'm not sure how useful it will be in the end considering the GH action, and then it will just be dead code.

That is true but I would feel more comfortable if core lifecycle operations are explicit and out front (precise command is "anchored" in a tox env and copy-paste-able into terminal) rather than obscure and far away (requiring us to lookup docs or dig out what some gh action does).

Also, iirc we said that publishing to stable would remain a manual operation?

rbarry82:
> That is true but I would feel more comfortable if core lifecycle operations are explicit and out front (precise command is "anchored" in a tox env and copy-paste-able into terminal) rather than obscure and far away (requiring us to lookup docs or dig out what some gh action does).
> 
> Also, iirc we said that publishing to stable would remain a manual operation?

I agree with you, and I think we should keep these anyway. Pushing to beta will probably also stay stable, and "magic" publishing isn't very discoverable either.

Just pointing out that using this tox command to push to edge will probably involve your edge version being overwritten when whatever patch merges, so the discussion about outdated resource revisions may not be meaningful.

tox.ini
dstathis: I sort of think we should just not have this. All the options should be configurable at which point why not just run charmcraft?

sed-i: > why not just run charmcraft?

Most charms only have "image" resource, but some charms have also file resource, which are forgettable.

> All the options should be configurable

This tox env suggests the opposite: most of the time, most options are fixed.

sed-i:
> edge version being overwritten when whatever patch merges

True. If the GH action will have an excellent UI, then this tox env would only have documentational value.

tox.ini
simskij: > The GH action to publish charm automatically fetches whatever is specified anyway.
> 
> We noticed initially because every merge to Grafana ends up putting Grafana 8 (latest) on `edge`, so this is probably less of a concern anyway. The action may be useful for manually deploying akin to a `make` target if there's an emergency fix which needs to go out while code review is still pending, though.

Is what you're saying that it does not respect pinning done in upstream-source? iIf so, you should file a bug in the actions repo.

simskij (changes requested):
I think it makes sense, but instead of hard-coding, you should fetch the latest `resource-revisions` from Charmhub and use those.

sed-i:
> I think it makes sense, but instead of hard-coding, you should fetch the latest `resource-revisions` from Charmhub and use those.

@simskij tox commands are pretty limited (cannot pipe). Any suggestion on how to accomplish this without an external script?

We could, instead of tox envs, do this in a new script that would do whatever we want, including deploy, refresh, publish.

tox.ini
simskij: Ping @rbarry82

simskij:
> > I think it makes sense, but instead of hard-coding, you should fetch the latest `resource-revisions` from Charmhub and use those.
> 
> @simskij tox commands are pretty limited (cannot pipe). Any suggestion on how to accomplish this without an external script?
> 
> We could, instead of tox envs, do this in a new script that would do whatever we want, including deploy, refresh, publish.

No good ones unfortunately. 😅

Abuelodelanada (approved):

tox.ini
rbarry82: We've had some time away from this, and... considering the amount of "this adds risks/non-deterministic behavior to make the lives of our own team marginally easier" things we've come across, I don't think it's a great idea.

We're now hardcoding "fixed" options to specific versions. What will we do when/if `promql-transform` bumps and we may have a difference in stable/edge and/or some other stream? What about having a different ROCK in edge versus stable? What if someone forgets to use this because they are relying on the GH action and the image "goes backwards" do one which has a security vulnerability?

* I will not reinvent `make` badly with `tox`.
* I will not reinvent `make` badly with `tox`.
* I will not reinvent `make` badly with `tox`.
* I will not reinvent `make` badly with `tox`.
* I will not reinvent `make` badly with `tox`.

Considering that `tox` is not built for (and is terrible at) any kind of dynamic/tagged updating, a GH action which can be triggered (with or without [selectable inputs ala Jenkins](https://github.community/t/how-to-pass-input-parameters-to-github-actions/123941)), or which is able to set these in a more flexible way.

We should not recreate `make` with far less flexibility/functionality. Hyrum's Law applies here. Such a tox env (deploy/refresh) will pretty much never be used by anyone other than people writing patches for/testing this charm, and `juju deploy` is reasonably good at error logging anyway, telling authors "you didn't give me a value for `--resource foo`"). 

If we're concerned about publishing and forgetting, we should open an issue with `charmcraft` (and they can push it back to `charmrepo | charmstore | juju/charmhub` if it's better handled there). 

It is, simply, just not the right tool for the job. Any job, really. Publishing to edge should be done through CI, and stable/beta can _also_ be done through CI with parameterized workflows when we get there. Those workflows can pick out "appropriate" values without hardcoding them or `sh -c '...'` (from putting a square `tox` in a round "automation" hole), and `tox -e deploy-{arch} || tox -e refresh-{arch} ...` are better off just being shell aliases, as they're purely developer workflows.

Then they could even use `dotenv` or similar to discover this

Add them to a `zsh`

sed-i:
Publishing is now documented:
https://github.com/canonical/prometheus-k8s-operator/blob/8f93ba78447b4c15e36f3f4673cb60068a099983/RELEASE.md?plain=1#L29-L32

Expanding the existing docs for deploy/refresh can be done in a separate PR.
https://github.com/canonical/prometheus-k8s-operator/blob/8f93ba78447b4c15e36f3f4673cb60068a099983/CONTRIBUTING.md?plain=1#L80-L85

Closing.

Issue #184, opened 2021-12-08
A Growing list of Integration test cases


balbirthomas:
The purpose of this ticket is to create a growing list of Integration test cases for the Prometheus charm.

1. Test Prometheus can recover from a broken config upon refresh. Suggested steps are
     - In a Metrics Provider charm introduce an invalid alert rule
     - Relate the invalid alert rule charm to Prometheus. Prometheus charm will go into blocked state.
     - Fix alert rule and redeploy Metrics Provider . Now check blocked status is removed.
     - Test behaviour of Prometheus with and without ingress.

mmanciop:
> Test behaviour of Prometheus with and without ingress

1. Try to use the PromQL API to fire queries based on data and we put into Prometheus through remote-write or scrape and validate the result with and without ingress

rbarry82:
The alert rule cases are already very heavily unit testing, and integration testing doesn't buy us anything there (especially with the changes to charmcraft part caching which caused issues with `loki-tester` and the write/remove functions in my integration test PR when you try to dunamically modify them).

Ingress is [already covered](https://github.com/canonical/prometheus-k8s-operator/blob/main/tests/integration/test_ingress_with_traefik_k8s.py).

The linked PR cover almost every case we'd possibly want to get from this issue which is not already present.

PR #243, opened 2022-03-09
Adjust reload API calls to Prometheus when ingress is related

mmanciop:
Ensure that, if an ingress is related over the `ingress` relation, if we need to set the `--web.route-prefix` argument, we correctly take it into account for the requests that we issue to Prometheus, e.g., to reload the configurations.

### DIFF
--- a/src/charm.py
+++ b/src/charm.py
@@ -145,8 +145,12 @@ class PrometheusCharm(CharmBase):
             logger.info("Prometheus configuration reloaded")
         else:
             container.add_layer(self._name, new_layer, combine=True)
-            container.replan()
-            logger.info("Prometheus (re)started")
+            try:
+                container.replan()
+                logger.info("Prometheus (re)started")
+            except Exception as e:
+                logger.error(f"Failed to replan; pebble plan: {container.get_plan().to_dict()}")
+                raise e
 
         if (
             isinstance(self.unit.status, BlockedStatus)
@@ -208,7 +212,8 @@ class PrometheusCharm(CharmBase):
         external_url = self._external_url
         args.append(f"--web.external-url={external_url}")
 
-        if path := self._web_route_prefix:
+        path = self._web_route_prefix
+        if not external_url and path:
             # We need to ensure there is a '/' character at the end
             # of the path, or Prometheus will not correctly
             # concatenate redirect headers, e.g., sending back a
@@ -368,7 +373,7 @@ class PrometheusCharm(CharmBase):
 
     @property
     def _web_route_prefix(self) -> str:
-        if path := urlparse(self._external_url):
+        if path := urlparse(self._external_url).path:
             return str(path).strip()
 
         return None
--- a/src/charm.py
+++ b/src/charm.py
@@ -212,9 +212,8 @@ class PrometheusCharm(CharmBase):
             # We need to ensure there is a '/' character at the end
             # of the path, or Prometheus will not correctly
             # concatenate redirect headers, e.g., sending back a
-            # 'Location': '/test-prometheus-k8s-0-/reload' header
-            # (see the lack of '/' between '-/reload' and the previous
-            # part of the path)
+            # 'Location': '/test-prometheus-k8s-0-/reload' header instead of
+            # 'Location': '/test-prometheus-k8s-0/-/reload'
             if not path.endswith("/"):
                 path = f"{path}/"
 
--- a/src/charm.py
+++ b/src/charm.py
@@ -212,18 +212,6 @@ class PrometheusCharm(CharmBase):
         external_url = self._external_url
         args.append(f"--web.external-url={external_url}")
 
-        path = self._web_route_prefix
-        if not external_url and path:
-            # We need to ensure there is a '/' character at the end
-            # of the path, or Prometheus will not correctly
-            # concatenate redirect headers, e.g., sending back a
-            # 'Location': '/test-prometheus-k8s-0-/reload' header instead of
-            # 'Location': '/test-prometheus-k8s-0/-/reload'
-            if not path.endswith("/"):
-                path = f"{path}/"
-
-            args.append(f"--web.route-prefix={path}")
-
         if self.model.get_relation(DEFAULT_REMOTE_WRITE_RELATION_NAME):
             args.append("--enable-feature=remote-write-receiver")
 
@@ -370,13 +358,6 @@ class PrometheusCharm(CharmBase):
             bind_address = str(bind_address)
 
         return bind_address
-
-    @property
-    def _web_route_prefix(self) -> str:
-        if path := urlparse(self._external_url).path:
-            return str(path).strip()
-
-        return None
 
     @property
     def _external_url(self) -> str:
--- a/src/charm.py
+++ b/src/charm.py
@@ -208,7 +208,7 @@ class PrometheusCharm(CharmBase):
         external_url = self._external_url
         args.append(f"--web.external-url={external_url}")
 
-        if path := self._web_route_prefix.strip():
+        if path := self._web_route_prefix:
             # We need to ensure there is a '/' character at the end
             # of the path, or Prometheus will not correctly
             # concatenate redirect headers, e.g., sending back a
@@ -368,7 +368,10 @@ class PrometheusCharm(CharmBase):
 
     @property
     def _web_route_prefix(self) -> str:
-        return urlparse(self._external_url).path or ""
+        if path := urlparse(self._external_url):
+            return str(path).strip()
+
+        return None
 
     @property
     def _external_url(self) -> str:
--- a/src/charm.py
+++ b/src/charm.py
@@ -45,7 +45,6 @@ class PrometheusCharm(CharmBase):
 
         self._name = "prometheus"
         self._port = 9090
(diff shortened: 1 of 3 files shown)


mmanciop:
Converting to draft as this PR has no tests and I dearly hope someone can contribute them :D

src/charm.py
```diff
@@ -355,6 +367,10 @@ def _pod_ip(self) -> str:
 
         return bind_address
 
+    @property
+    def _web_route_prefix(self) -> str:
```
sed-i, line 371: Is "web route prefix" standard terminology? Should this be name "subpath" or "ingress_subpath" instead?

src/prometheus_server.py
```diff
@@ -14,15 +14,26 @@
...
             port: Optional; port on which Prometheus service is exposed.
+            web_route_prefix: Optional; the root path added to the Prometheus API path, e.g.,
+              when we relate to an ingress.
             api_timeout: Optional; timeout (in seconds) to observe when interacting with the API.
         """
-        self.base_url = f"http://{host}:{port}"
+        if web_route_prefix and not web_route_prefix.startswith("/"):
+            web_route_prefix = f"/{web_route_prefix}"
+
+        self.base_url = f"http://{host}:{port}{web_route_prefix or ''}"
```
sed-i, line 36: Maybe use `urljoin` here? It will be both more clear and less error prone

```
>>> from urllib.parse import urljoin
>>> urljoin("http://localhost:9090", "/sub1")
'http://localhost:9090/sub1'
>>> urljoin("http://localhost:9090", "sub2")
'http://localhost:9090/sub2'
>>> urljoin("http://localhost:9090/", "sub3")
'http://localhost:9090/sub3'
>>> urljoin("http://localhost:9090/", "/sub4")
'http://localhost:9090/sub4'
```

src/prometheus_server.py
```diff
@@ -35,7 +46,7 @@ def reload_configuration(self) -> bool:
         """
         url = f"{self.base_url}/-/reload"
         try:
-            response = post(url, timeout=self.api_timeout)
+            response = post(url, timeout=self.api_timeout, allow_redirects=False)
```
sed-i, line 49: Why not allow redirects? Might be worth a comment.

tests/unit/test_server.py
```diff
@@ -9,42 +9,84 @@
...
-    @responses.activate
-    def test_prometheus_server_reload_configuration_failure(self):
-        responses.add(
-            responses.POST,
-            "http://localhost:9090/-/reload",
-            status=500,
-        )
-
-        self.assertFalse(self.prometheus.reload_configuration())
+    class _TestServerWithoutRoutePrefix(unittest.TestCase):
```
sed-i, line 12: None of these tests ran: https://github.com/canonical/prometheus-k8s-operator/runs/5485103377?check_suite_focus=true#step:5:1

I don't think nested test classes are picked up automatically.
But why nest them?

src/charm.py
mmanciop: It's how prometheus calls it. I saw no reason to add to the confusion with a different term.

src/charm.py
```diff
@@ -205,7 +208,16 @@ def _command(self) -> str:
         external_url = self._external_url
         args.append(f"--web.external-url={external_url}")
 
-        if path := urlparse(external_url).path:
+        if path := self._web_route_prefix.strip():
```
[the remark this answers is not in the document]
mmanciop: I *guess* it could go in `self._web_route_prefix`.

src/prometheus_server.py
mmanciop: `urljoin` is now effectively my new favorite Python3 API.

src/prometheus_server.py
mmanciop: It was actually not needed anymore. The reason why at some point in making the PR I thought I needed it, is that under some circumstances (which we solved with `urljoin`), the URL would not be well-formed and Prometheus would give me a 301 back, which then resulted in a 405, as the redirect would use a `GET` method that is not supported by the reload API.

tests/unit/test_server.py
mmanciop: Crud you are right

src/prometheus_server.py
mmanciop: I take it back: `urljoin` does not work appending subpaths :-(

mmanciop: So, in the end it is about ensuring the base URL ends with `/` and the path does not start with `/`. That was a LOT of footguns.

src/prometheus_server.py
```diff
@@ -33,7 +59,7 @@ def reload_configuration(self) -> bool:
         Returns:
           True if reload succeeded (returned 200 OK); False otherwise.
         """
-        url = f"{self.base_url}/-/reload"
+        url = urljoin(self.base_url, "-/reload")
```
balbirthomas, line 62: Nitpick but worth thinking. Typically I would expect a programmer will write a URL path starting with a slash eg. `/home_dir` rather than `home`. While I see no harm in adding a terminating slash to `self.base_url`, I do not see why remove it from `/-/reload`. `urljoin` is perfectly fine with having both terminating slash in `self.base_url` and `/-/reload`.

simskij (dismissed):

simskij:
once the tests pass, I'm ok with this being merged.

src/charm.py
```diff
@@ -134,7 +133,11 @@ def _configure(self, _):
         # Restart prometheus only if command line arguments have changed,
         # otherwise just reload its configuration.
         if current_services == new_layer.services:
-            reloaded = self._prometheus_server.reload_configuration()
+            external_url = urlparse(self._external_url)
+
+            prometheus_server = Prometheus(web_route_prefix=external_url.path)
```
balbirthomas, line 138: Couple of things I am not clear about (possibly not directly related to this PR)

1. Prometheus [recommends](https://github.com/prometheus/prometheus/issues/4925) not to use both `--web.route-prefix` and `--web.external-url` if the latter **already includes a path**. I do not see any conditional check in our code that ensures this. In fact we are always constructing `_web_route_prefix()` from `_external_url()` and setting it in `_command()` (see `src/charm.py`) . Do we have a potential problem here ? 

2. Do we expect workloads from any related charm may try to make legitimate API requests directly to Prometheus API endpoint ? If so should Prometheus be restarted on an ingress changed event too, so that any new external URL and web route prefix take effect ?

tests/unit/test_server.py
```diff
@@ -40,11 +41,56 @@ def test_prometheus_server_reload_configuration_success(self):
...
         responses.add(
             responses.POST,
             "http://localhost:9090/-/reload",
             status=500,
         )
 
         self.assertFalse(self.prometheus.reload_configuration())
+
+    @responses.activate
+    def test_prometheus_server_with_route_prefix_returns_valid_data(self):
```
balbirthomas, line 56: Since most of these tests involve mocking I added a note to myself in [Prometheus issue 184](https://github.com/canonical/prometheus-k8s-operator/issues/184) as a reminder to create integration tests for ingress+Prometheus. If you have a deeper insight here on what the test case should look like kindly do refine it in the issue tracker.

balbirthomas (approved):

src/charm.py
mmanciop: @balbirthomas 
1. Where? I recall Prometheus not working correctly without both
2. Well, that's pretty much what `prometheus_remote_write` does, no? Today, if the cli args of prometheus are different, we do restart. It's how https://github.com/canonical/prometheus-k8s-operator/blob/f4fdb6747647af7df93bdea4f8dace581abd30c9/src/charm.py#L145 works AFAICT

src/prometheus_server.py
mmanciop: > urljoin is perfectly fine with having both terminating slash in self.base_url and /-/reload

I had tried it. `urljoin` seems to disagree with your assessment :D

mmanciop:
```
Python 3.9.5 (default, Nov 18 2021, 16:00:48) 
[GCC 10.3.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> from urllib.parse import urljoin
>>> urljoin("https://ubuntu.com/michele/is/right","/balbir/is/wrong")
'https://ubuntu.com/balbir/is/wrong'
```

Yes I am being petty here :D

balbirthomas:
```
Python 3.8.10 (default, Nov 26 2021, 20:14:08) 
[GCC 9.3.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> from urllib.parse import urljoin
>>> urljoin("https://ubuntu.com/michele/is/right","/balbir/is/wrong")
'https://ubuntu.com/balbir/is/wrong'
>>> urljoin("https://ubuntu.com/michele/is/right/","/balbir/is/wrong")
'https://ubuntu.com/balbir/is/wrong'
>>> 
```
This is what I am talking about.

balbirthomas: I see what you mean.

mmanciop:
Itests went green, running them again to be double sure.

balbirthomas (approved):

PR #245, opened 2022-03-11
Updated release and scrape job name documentation

balbirthomas:
This commit refines tow aspects of the charm documentation
1. The charmcraft command to publish the charm is now explicitly
 specified.
2. An advisory on how the select a scrape job name for use with
 the prometheus_scrape interface has now been added.

### DIFF
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -22,6 +22,16 @@ For more information about CharmHub channels, refer to the [Juju charm store](ht
 Refer to the [Publish your operator in Charmhub](https://discourse.charmhub.io/t/publish-your-operator-in-charmhub) documentation.
 After a `latest/stable` release, it is expected that the version of the charm is the same as the one in `latest/candidate`, and those two channels will diverge again when we are ramping up through `latest/candidate` releases for a new `latest/stable` release.
 
+It is **important** to note that when making new releases of this charm **two** resource need to be bound to it
+- The container OCI image : `prometheus-image`
+- The PromQL transform binary : `promql-transform-amd64`
+
+An example of the charmcraft command to publish a new release is shown below.
+```
+charmcraft release prometheus-k8s --channel=beta --revision=19 --resource=prometheus-image:1 --resource=promql-transform-amd64:3
+```
+Note: The charm revision and resource revisions will need to be set appropriately.
+
 ## A note on granularity of revisions
 
 We believe in shipping often and with confidence.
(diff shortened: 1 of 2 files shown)


lib/charms/prometheus_k8s/v0/prometheus_scrape.py
```diff
@@ -157,6 +157,12 @@ def __init__(self, *args):
 ]
 ```
 
+It is **important** to note that the `job_name` must be a static
+hardcoded string. For instance **do not** set the job name using your
+charm's `unit.name`. If the job name changes on upgrading your charm,
+scaling the number of units etc., this will break the continuity of
+the metrics time series gathered by Prometheus.
```
sed-i, line 164:
```suggestion
It is **important** to note that the `job_name` should be a fixed
string (e.g. hardcoded literal). For instance, if you include your charm's
`unit.name`, it would break the continuity of the metrics time series 
gathered by Prometheus when the leader unit changes (e.g. on upgrade
or rescale).
```

sed-i (approved):

RELEASE.md
```diff
@@ -22,6 +22,16 @@ For more information about CharmHub channels, refer to the [Juju charm store](ht
 Refer to the [Publish your operator in Charmhub](https://discourse.charmhub.io/t/publish-your-operator-in-charmhub) documentation.
 After a `latest/stable` release, it is expected that the version of the charm is the same as the one in `latest/candidate`, and those two channels will diverge again when we are ramping up through `latest/candidate` releases for a new `latest/stable` release.
 
+It is **important** to note that when making new releases of this charm **two** resource need to be bound to it
```
rbarry82, line 25:
```suggestion
It is **important** to note that when making new releases of this charm **two** resources need to be bound to it
```

lib/charms/prometheus_k8s/v0/prometheus_scrape.py
rbarry82: This suggestion is confusing. `fixed string (e.g. hardcoded literal)` is very verbose compared to the plain `string literal` which is used all over, including the official Python docs.

I'd go with something like:

```
It is **important** to note that `job_name` should be specified in a format which is constant throughout the application lifecycle. A good example would be a hardcoded string, or a string literal which takes formatting only from `charm|application`-level data (e.g. `"foo_{app.name}"`). When a juju application is refreshed, upgraded, or scaled, `unit`-level data may change. A partial list of things to avoid: `unit.name`, charm revision, resource revisions, or any data specific to your job which may change as units are added and removed, upgraded, or relations changed.

> If `job_name` changes for any reason, all existing time series gathered by Prometheus will be orphaned, which will break the continuity monitoring.
```

mmanciop (dismissed):

PR #247, opened 2022-03-11
[WIP] transaction from charm lib to python package

r-f-g:
This change represents a transaction from prometheus_scrape charm library
to prometheus-scrape as a python library. This allows the use of the
prometheus-scrape interface for LMA reactive charms instead of using the http
interface and to achieve consistency between the Prometheus charm on
Kubernetes and on machine.


r-f-g:
This is a draft as it is a significant change that needs to be properly tested.

mmanciop:
I am not sure I understand the scope and goal of this patch. I understand that the goal is to replace the charm library we have today with a Python package, but that is not the standard way of doing libraries for sidecar charms.

Moreover, and changing the relation interface from `prometheus_scrape` to `prometheus-scrape` would break every single charm out there that has integrated with COS, and that is not a breaking change we can accept.

As a note: For reactive charms, we have backwards compatibility with LMA relations via https://charmhub.io/cos-proxy .

Happy to discuss the idea further, I have the feeling I am missing something.

PR #249, opened 2022-03-15
Use hostname instead of IP address

dstathis:
Use hostname instead of IP address to avoid issues with pod churn. Also, make sure we update remote write to use this on charm upgrade.

QA: no specific steps here other than just that Prometheus generally works.

### DIFF
--- a/src/charm.py
+++ b/src/charm.py
@@ -7,6 +7,7 @@
 import logging
 import os
 import re
+import socket
 from typing import Dict
 from urllib.parse import urlparse
 
@@ -157,6 +158,12 @@ class PrometheusCharm(CharmBase):
             and self.unit.status.message != CORRUPT_PROMETHEUS_CONFIG_MESSAGE
         ):
             return
+
+        # On upgrade, make sure the remote write endpoint is changed. Note that
+        # this is only needed for upgrade from previous versions that
+        # use bind_address rather than the fqdn. If at some point we break the
+        # upgrade path, we should look into removing this.
+        self.remote_write_provider.update_endpoint()
 
         self.unit.status = ActiveStatus()
 
@@ -352,14 +359,6 @@ class PrometheusCharm(CharmBase):
         return Layer(layer_config)
 
     @property
-    def _pod_ip(self) -> str:
-        """Returns the pod ip of this unit."""
-        if bind_address := self.model.get_binding("prometheus-peers").network.bind_address:
-            bind_address = str(bind_address)
-
-        return bind_address
-
-    @property
     def _external_url(self) -> str:
         """Return the external hostname to be passed to ingress via the relation."""
         if "web_external_url" in self.model.config:
@@ -375,7 +374,7 @@ class PrometheusCharm(CharmBase):
         # are routable virtually exclusively inside the cluster (as they rely)
         # on the cluster's DNS service, while the ip address is _sometimes_
         # routable from the outside, e.g., when deploying on MicroK8s on Linux.
-        return f"http://{self._pod_ip}:{self._port}"
+        return f"http://{socket.getfqdn()}:{self._port}"
 
 
 if __name__ == "__main__":
--- a/src/charm.py
+++ b/src/charm.py
@@ -82,26 +82,12 @@ class PrometheusCharm(CharmBase):
         self.framework.observe(self.on.prometheus_pebble_ready, self._configure)
         self.framework.observe(self.on.config_changed, self._configure)
         self.framework.observe(self.on.upgrade_charm, self._configure)
-        self.framework.observe(self.ingress.on.ingress_changed, self._on_ingress_changed)
+        self.framework.observe(self.ingress.on.ingress_changed, self._configure)
         self.framework.observe(self.on.receive_remote_write_relation_created, self._configure)
         self.framework.observe(self.on.receive_remote_write_relation_changed, self._configure)
         self.framework.observe(self.on.receive_remote_write_relation_broken, self._configure)
         self.framework.observe(self.metrics_consumer.on.targets_changed, self._configure)
         self.framework.observe(self.alertmanager_consumer.on.cluster_changed, self._configure)
-
-    def _on_upgrade_charm(self, event):
-        """Handler for the upgrade_charm event during which will update the K8s service."""
-        self._configure(event)
-
-    def _on_ingress_changed(self, event):
-        self._configure(event)
-        # The remote_write_provider object has the update external URL,
-        # as the data from the ingress relation is looked up in the Charm
-        # constructor, and it is already updated. However, we do need to
-        # trigger the update of the endpoint to the remote_write clients,
-        # as remote_write_provider does not observe any ingress-specific
-        # events.
-        self.remote_write_provider.update_endpoint()
 
     def _configure(self, _):
         """Reconfigure and either reload or restart Prometheus.
@@ -159,10 +145,7 @@ class PrometheusCharm(CharmBase):
         ):
             return
 
-        # On upgrade, make sure the remote write endpoint is changed. Note that
-        # this is only needed for upgrade from previous versions that
-        # use bind_address rather than the fqdn. If at some point we break the
-        # upgrade path, we should look into removing this.
+        # Make sure that if the remote_write endpoint changes, it is reflected in relation data.
         self.remote_write_provider.update_endpoint()
 
         self.unit.status = ActiveStatus()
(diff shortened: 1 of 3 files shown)


src/charm.py
```diff
@@ -158,6 +159,12 @@ def _configure(self, _):
         ):
             return
 
+        # On upgrade, make sure the remote write endpoint is changed. Note that
+        # this is only needed for upgrade from previous versions that
+        # use bind_address rather than the fqdn. If at some point we break the
+        # upgrade path, we should look into removing this.
+        self.remote_write_provider.update_endpoint()
```
mmanciop, line 166: Yeah but then shouldn’t we rather do it in an upgrade_charm hook?

sed-i:
Can you try to remove `IPAddressWorkaround` now that it uses fqdn?
https://github.com/canonical/prometheus-k8s-operator/blob/525792b24becd5f8e07afbef70179d6ea6f2308f/tests/integration/helpers.py#L231

balbirthomas:
@dylanstathis could you please elaborate on what is "the issue with pod churn" mentioned in the PR message text. This understanding will help me replicate the issue so that I can validate this fix does indeed solve the problem and is the right fix for the problem.

rbarry82:
> @dstathis could you please elaborate on what is "the issue with pod churn" mentioned in the PR message text. This understanding will help me replicate the issue so that I can validate this fix does indeed solve the problem and is the right fix for the problem.

Dylan will have to say for sure, but as a generalization, a rescheduled pod for unit-0 will get the same hostname/DNS name, but the same IP address is not guaranteed (or even likely)

dstathis:
> > @dstathis could you please elaborate on what is "the issue with pod churn" mentioned in the PR message text. This understanding will help me replicate the issue so that I can validate this fix does indeed solve the problem and is the right fix for the problem.
> 
> Dylan will have to say for sure, but as a generalization, a rescheduled pod for unit-0 will get the same hostname/DNS name, but the same IP address is not guaranteed (or even likely)

Yes exactly this

balbirthomas:
> > Dylan will have to say for sure, but as a generalization, a rescheduled pod for unit-0 will get the same hostname/DNS name, but the same IP address is not guaranteed (or even likely)

> Yes exactly this

So steps to reproduce the problem would be

```
juju deploy prometheus-k8s
kubectl delete pod prometheus-k8s/0
```

An look for what to happen ??

dstathis:
> > > Dylan will have to say for sure, but as a generalization, a rescheduled pod for unit-0 will get the same hostname/DNS name, but the same IP address is not guaranteed (or even likely)
> 
> > Yes exactly this
> 
> So steps to reproduce the problem would be
> 
> ```
> juju deploy prometheus-k8s
> kubectl delete pod prometheus-k8s/0
> ```
> 
> An look for what to happen ??

Okay I think the point is kind of missed here. The bug that brought this to my attention is that a juju refresh wouldn't change the endpoint given over a remote write relation. However, the QA effort should be more focused on making sure this doesn't break *anything else*. The `update_endpoint` call is what is actually fixing a bug while the change to using `socket.getfqdn()` is a measure to make future bugs relating to pod churn less likely. If you are looking to test this PR it would be more helpful to check that it doesn't break other Prometheus workflows as the `_external_url` method is used in other places.

src/charm.py
dstathis: This is the upgrade charm hook

mmanciop: This is almost every hook ;-)

Wouldn’t it make sense to add an intermediate `_on_charm_upgrade(self, event)` method that does this invocation and then delegates to `self._configure(event)`?

dstathis: Yes but I didn't not want to so aggressively violate the design decisions of the original author. I will bring it up in the next meeting.

dstathis: Upon more inspection, I realized that this is needed in `upgrade_charm`, `pebble_ready` (for pod churn), and `ingress_changed`. So I just fixed the comment and kept it in `_configure`.

dstathis:
> Can you try to remove `IPAddressWorkaround` now that it uses fqdn?
> 
> https://github.com/canonical/prometheus-k8s-operator/blob/525792b24becd5f8e07afbef70179d6ea6f2308f/tests/integration/helpers.py#L231

I think we should create a code upkeep issue for this and do it later. For the time being, shortening update-status-interval does not hurt anything.

mmanciop (approved):

rbarry82 (approved):
A #FIXME or issue to get rid of `IPAddressWorkaround` would be great, too.

src/charm.py
balbirthomas: I am wondering -- "if it makes sense for the Ingress object listen to pod churn events (`RelationJoined`, `RelationDeparted` etc) and raise `IngressChanged` events".

balbirthomas (approved):

mmanciop:
> I am wondering -- "if it makes sense for the Ingress object to listen to pod churn events (`RelationJoined`, `RelationDeparted` etc) and raise `IngressChanged` events".

@balbirthomas the Ingress object _does listen_ to those events, but those events are issued to a Unit X only when _other units_ join or depart. So, since we want to update the IP of one unit that keeps existing, I don't think it would help.

rbarry82:
> > I am wondering -- "if it makes sense for the Ingress object to listen to pod churn events (`RelationJoined`, `RelationDeparted` etc) and raise `IngressChanged` events".
> 
> @balbirthomas the Ingress object _does listen_ to those events, but those events are issued to a Unit X only when _other units_ join or depart. So, since we want to update the IP of one unit that keeps existing, I don't think it would help.

The (good) alternative would be to have `traefik-k8s` do lookups of the service by DNS instead of IP, but it's more complex than it appears since the ROCKS images/pods do not have any default search domains, so it would need to be a fully qualified lookup. 

If we're using the FQDN, though, it should not be necessary to change it on churn. DNS will still work

mmanciop:
> > > I am wondering -- "if it makes sense for the Ingress object to listen to pod churn events (`RelationJoined`, `RelationDeparted` etc) and raise `IngressChanged` events".
> > 
> > 
> > @balbirthomas the Ingress object _does listen_ to those events, but those events are issued to a Unit X only when _other units_ join or depart. So, since we want to update the IP of one unit that keeps existing, I don't think it would help.
> 
> The (good) alternative would be to have `traefik-k8s` do lookups of the service by DNS instead of IP, but it's more complex than it appears since the ROCKS images/pods do not have any default search domains, so it would need to be a fully qualified lookup.
> 
> If we're using the FQDN, though, it should not be necessary to change it on churn. DNS will still work

There is precisely zero guarantee at this stage that `traefik-k8s` and whatever it is proxying are in the same model (hence, same K8s cluster) or not. Relying on the an FQDN that exists only in the cluster local DNS service would be super unsafe.

mmanciop:
> > > I am wondering -- "if it makes sense for the Ingress object to listen to pod churn events (`RelationJoined`, `RelationDeparted` etc) and raise `IngressChanged` events".
> > 
> > 
> > @balbirthomas the Ingress object _does listen_ to those events, but those events are issued to a Unit X only when _other units_ join or depart. So, since we want to update the IP of one unit that keeps existing, I don't think it would help.
> 
> The (good) alternative would be to have `traefik-k8s` do lookups of the service by DNS instead of IP, but it's more complex than it appears since the ROCKS images/pods do not have any default search domains, so it would need to be a fully qualified lookup.
> 
> If we're using the FQDN, though, it should not be necessary to change it on churn. DNS will still work

There is precisely zero guarantee at this stage that `traefik-k8s` and whatever it is proxying are in the same model (hence, same K8s cluster) or not. Relying on the an FQDN that exists only in the cluster local DNS service would be super unsafe.

rbarry82:
> > If we're using the FQDN, though, it should not be necessary to change it on churn. DNS will still work
> 
> There is precisely zero guarantee at this stage that `traefik-k8s` and whatever it is proxying are in the same model (hence, same K8s cluster) or not. Relying on the an FQDN that exists only in the cluster local DNS service would be super unsafe.

Well, nothing is saying that it _has_ to be CoreDNS, or that CoreDNS isn't configured to use some resolver on the system, but what DNS infrastructure will look like in the end isn't my call.

rbarry82 (approved):

tests/unit/test_remote_write.py
```diff
@@ -163,6 +163,7 @@ def setUp(self, *unused):
 
     @patch.object(KubernetesServicePatch, "_service_object", new=lambda *args: None)
     @patch.object(Prometheus, "reload_configuration", new=lambda _: True)
+    @patch("socket.getfqdn", new=lambda *args: "fqdn")
```
sed-i, line 166: Above you use `fqdn = socket.getfqdn()`. Any reason why it is mocked here?

sed-i (approved):

PR #253, opened 2022-03-26
Make sure that ingress works flexibly with `GrafanaSourceProvider`

rbarry82:
## Issue
Previously, `GrafanaSourceProvider` assumed that a bare `{host}:{port}` would be sufficient as a condition, but Prometheus, when `web.external-url` is set, also changes all web entrypoints to subpath routing. The majority of the work here happened in Grafana, but some changes were necessary to actually get the values from Prometheus in.

## Solution
Instantiate `GrafanaSourceProvider` with `_external_url`.

## Testing Instructions
Easy way to check: use the explore tab of Grafana to query Prometheus data. If it does not work, you’ll be told by Grafana that there are “Bad gateway” issues.

Make sure Traefik is in Active status (https://github.com/canonical/traefik-k8s-operator#setup), otherwise it does not pass URLs to Prometheus and you are not testing much.

Also try to change a few times, back and forth, the routing_mode setting for Traefik (https://github.com/canonical/traefik-k8s-operator#configurations), as that affects changes that need to cascade through the Grafana Source relation.

You can check which URLs are provided by Traefik to Prometheus and others using this action: https://github.com/canonical/traefik-k8s-operator/blob/afb2e1f2aca5b1d2e8ee479aa21df9d5e4417b3b/actions.yaml#L1

## Release Notes
Make sure that ingress works flexibly with `GrafanaSourceProvider`

### DIFF
--- a/lib/charms/grafana_k8s/v0/grafana_source.py
+++ b/lib/charms/grafana_k8s/v0/grafana_source.py
@@ -25,17 +25,16 @@ provider in a charm which `provides: grafana-source`, creation of a
 The default arguments are:
 
     `charm`: `self` from the charm instantiating this library
+    `source_type`: None
+    `source_port`: None
+    `source_url`: None
     `relation_name`: grafana-source
     `refresh_event`: A `PebbleReady` event from `charm`, used to refresh
         the IP address sent to Grafana on a charm lifecycle event or
         pod restart
-    `source_type`: ""
-    `source_port`: ""
-    `source_uri`: ""
 
-`source_uri` should be a fully-resolvable URI for a valid Grafana source.
-As only HTTP sources are supported via libraries in the current charm, this
-is likely to be `http://example.com/api` or similar.
+The value of `source_url` should be a fully-resolvable URL for a valid Grafana
+source, e.g., `http://example.com/api` or similar.
 
 If your configuration requires any changes from these defaults, they
 may be set from the class constructor. It may be instantiated as
@@ -47,7 +46,9 @@ follows:
         def __init__(self, *args):
             super().__init__(*args, **kwargs)
             ...
-            self.grafana_source_provider = GrafanaSourceProvider(self)
+            self.grafana_source_provider = GrafanaSourceProvider(
+                self, source_type="prometheus", source_port="9090"
+            )
             ...
 
 The first argument (`self`) should be a reference to the parent (datasource)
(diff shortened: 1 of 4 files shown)


mmanciop:
> Make sure it works.

Easy way to check: use the explore tab of Grafana to query Prometheus data. If it does not work, you’ll be told by Grafana that there are “Bad gateway” issues.

Make sure Traefik is in Active status (https://github.com/canonical/traefik-k8s-operator#setup), otherwise it does not pass URLs to Prometheus and you are not testing much.

Also try to change a few times, back and forth, the `routing_mode` setting for Traefik (https://github.com/canonical/traefik-k8s-operator#configurations), as that affects changes that need to cascade through the Grafana Source relation.

tests/unit/test_charm.py
```diff
@@ -34,10 +34,27 @@ def setUp(self, *unused):
...
         grafana_host = self.harness.get_relation_data(rel_id, self.harness.model.unit.name)[
             "grafana_source_host"
         ]
-        self.assertEqual(grafana_host, "{}:{}".format("1.1.1.1", "9090"))
+        self.assertEqual(grafana_host, "http://{}:{}".format(fqdn, "9090"))
+
+    @patch_network_get(private_address="1.1.1.1")
+    def test_web_external_url_is_passed_to_grafana(self, *unused):
+        self.harness.set_leader(True)
+        self.harness.update_config({"web_external_url": "http://test:80/foo/bar"})
```
mmanciop, line 46: The `web_external_url` option overrides whatever may come from Traefik.

We must add a test with the URL coming from Traefik.

mmanciop (changes requested):
Code is fine, needs a unit test with a URL coming from traefik and them IMO good to go.

tests/unit/test_charm.py
mmanciop: I tried two hours to debug SDI to make one such test. I give up :D

mmanciop (approved):

mmanciop (changes requested):
~~We have now this issue to deal with: https://github.com/canonical/grafana-k8s-operator/pull/85#issuecomment-1079661302~~

~~I have never seen it between prometheus and traefik before, so it might be related with this PR.~~

No it's an SDI bug https://github.com/canonical/traefik-k8s-operator/issues/32

mmanciop:
@jnsgruk it will please you to know what it seems to be another SDI bug

mmanciop (approved):

lib/charms/grafana_k8s/v0/grafana_source.py
```diff
@@ -305,15 +315,14 @@ class GrafanaSourceEvents(ObjectEvents):
...
     def __init__(
         self,
         charm: CharmBase,
+        source_type: str,
+        source_port: Optional[str] = "",
+        source_url: Optional[str] = "",
         refresh_event: Optional[BoundEvent] = None,
         relation_name: str = DEFAULT_RELATION_NAME,
-        source_type: Optional[str] = "prometheus",
-        source_port: Optional[str] = "9090",
```
dstathis, line 316: Is there a reason we are breaking backwards compatibility here by changing the defaults?

mmanciop: Yes: they are bad defaults. They are not useful to anybody except Prometheus, and actively making it harder for people providing more datasources to understand how to use the API.

PR #254, opened 2022-03-29
remote_write_2

dstathis:
Update the remote write library with various improvements.

## Issue
There are two significant issues that this addresses:
* Alert rules do not get propagated from the workload charm, through grafana-agent, to prometheus
* Alert rules get `juju_unit` added to the expression and the labels

## Solution
* Make sure to only modify an alert rule if it has not already been modified
* Use the same updated JujuTopology class that prometheus_scrape uses

## Testing Instructions
1. Deploy prometheus, grafana-agent (remote_write_2 branch), and a workload (cassandra, spring-music, etc.)
2. Relate prometheus->grafana-agent->workload
3. Ensure alert rules from workload are present and labeled correctly (no unit)
4. Ensure metrics are present and labeled correctly (optionally relate prometheus to grafana)
5. Michele style tests

## Release Notes
* Overhaul of prometheus_remote_write bringing it up to working order

### DIFF
--- a/lib/charms/prometheus_k8s/v0/prometheus_remote_write.py
+++ b/lib/charms/prometheus_k8s/v0/prometheus_remote_write.py
@@ -451,9 +451,8 @@ class AlertRules:
 
     def _is_already_modified(self, name: str) -> bool:
         """Detect whether a group name has already been modified with juju topology."""
-        uuid_regex = "[a-fA-F0-9]{8}-[a-fA-F0-9]{4}-[a-fA-F0-9]{4}-[a-fA-F0-9]{4}-[a-fA-F0-9]{12}"
-        final_regex = ".*_{}_.*_alerts".format(uuid_regex)
-        if re.match(final_regex, name) is not None:
+        modified_matcher = re.compile(r"^.*?_[\da-f]{8}-([\da-f]{4}-){3}[\da-f]{12}_.*?alerts$")
+        if modified_matcher.match(name) is not None:
             return True
         return False
 
--- a/lib/charms/prometheus_k8s/v0/prometheus_remote_write.py
+++ b/lib/charms/prometheus_k8s/v0/prometheus_remote_write.py
@@ -14,6 +14,7 @@ import json
 import logging
 import os
 import platform
+import re
 import subprocess
 from collections import OrderedDict
 from pathlib import Path
@@ -32,7 +33,7 @@ LIBAPI = 0
 
 # Increment this PATCH version before using `charmcraft publish-lib` or reset
 # to 0 if you are raising the major API version
-LIBPATCH = 2
+LIBPATCH = 3
 
 
 logger = logging.getLogger(__name__)
@@ -131,6 +132,10 @@ class JujuTopology:
             application: an application name as a string
             unit: a unit name as a string
             charm_name: name of charm as a string
+
+        Note:
+            `JujuTopology` should not be constructed directly by charm code. Please
+            use `ProviderTopology` or `AggregatorTopology`.
         """
         self.model = model
         self.model_uuid = model_uuid
@@ -167,6 +172,7 @@ class JujuTopology:
                 - "application"
                 - "unit"
                 - "charm_name"
+
                 `unit` and `charm_name` may be empty, but will result in more limited
                 labels. However, this allows us to support payload-only charms.
 
@@ -185,16 +191,15 @@ class JujuTopology:
     def identifier(self) -> str:
         """Format the topology information into a terse string."""
         # This is odd, but may have `None` as a model key
-        return "_".join([str(val) for val in self.as_dict().values()]).replace("/", "_")
+        return "_".join([str(val) for val in self.as_promql_label_dict().values()]).replace(
+            "/", "_"
+        )
 
     @property
     def promql_labels(self) -> str:
         """Format the topology information into a verbose string."""
         return ", ".join(
-            [
-                'juju_{}="{}"'.format(key, value)
-                for key, value in self.as_dict(rename_keys={"charm_name": "charm"}).items()
-            ]
+            ['{}="{}"'.format(key, value) for key, value in self.as_promql_label_dict().items()]
         )
 
     def as_dict(self, rename_keys: Optional[Dict[str, str]] = None) -> OrderedDict:
@@ -234,6 +239,10 @@ class JujuTopology:
             "juju_{}".format(key): val
             for key, val in self.as_dict(rename_keys={"charm_name": "charm"}).items()
         }
+        # The leader is the only unit that sets alert rules, if "juju_unit" is present,
+        # then the rules will only be evaluated for that unit
+        if "juju_unit" in vals:
+            vals.pop("juju_unit")
 
         return vals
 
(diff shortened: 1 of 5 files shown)


pyproject.toml
```diff
@@ -20,7 +20,6 @@ profile = "black"
 [tool.flake8]
 max-line-length = 99
 max-doc-length = 99
-max-complexity = 10
```
sed-i, line 23: Just curious which one is the offending function

tests/unit/prometheus_alert_rules/with_template_string_and_unit.rule
```diff
@@ -0,0 +1,9 @@
+alert: PrometheusTargetMissing
```
sed-i, line 1: Re filename: I think we call the %% literal "topology stub", not "template string".

tests/unit/test_remote_write.py
```diff
@@ -153,6 +153,63 @@ def test_no_remote_write_endpoint_provided(self):
         self.harness.update_relation_data(rel_id, "provider/0", {})
         assert list(self.harness.charm.remote_write_consumer.endpoints) == []
 
+    def test_alert_rule_is_correct(self):
```
sed-i, line 156: Function name is a bit vague.

tests/unit/test_remote_write.py
```diff
@@ -153,6 +153,63 @@ def test_no_remote_write_endpoint_provided(self):
...
+                        "juju_model",
+                        "juju_model_uuid",
+                        "severity",
+                    },
+                )
+                break
+        else:
+            assert False  # Could not find the correct alert rule to check
+
+    def test_alert_rule_is_correct_with_unit(self):
```
sed-i, line 184: Function name is a bit vague.

tests/unit/test_remote_write.py
```diff
@@ -153,6 +153,63 @@ def test_no_remote_write_endpoint_provided(self):
...
+                        "juju_charm",
+                        "juju_model",
+                        "juju_model_uuid",
+                        "severity",
+                        "juju_unit",
+                    },
+                )
+                break
+        else:
+            assert False  # Could not find the correct alert rule to check
```
sed-i, line 211: Maybe instead of for-else it would be clearer if you first `filter(str.endswith, ...)`?

lib/charms/prometheus_k8s/v0/prometheus_remote_write.py
```diff
@@ -392,20 +401,24 @@ def _from_file(self, root_path: Path, file_path: Path) -> List[dict]:
 
             # update rules with additional metadata
             for alert_group in alert_groups:
-                # update group name with topology and sub-path
-                alert_group["name"] = self._group_name(
-                    str(root_path),
-                    str(file_path),
-                    alert_group["name"],
-                )
+                if not self._is_already_modified(alert_group["name"]):
```
sed-i, line 404: In what scenario would it already be modified?

lib/charms/prometheus_k8s/v0/prometheus_remote_write.py
```diff
@@ -436,6 +449,14 @@ def _group_name(self, root_path: str, file_path: str, group_name: str) -> str:
         # filter to remove empty strings
         return "_".join(filter(None, group_name_parts))
 
+    def _is_already_modified(self, name: str) -> bool:
+        """Detect whether a group name has already been modified with juju topology."""
+        uuid_regex = "[a-fA-F0-9]{8}-[a-fA-F0-9]{4}-[a-fA-F0-9]{4}-[a-fA-F0-9]{4}-[a-fA-F0-9]{12}"
+        final_regex = ".*_{}_.*_alerts".format(uuid_regex)
```
sed-i, line 455: A sick mind could create a false positive here.
I.e. this check doesn't guarantee 100% what it claims to do.
Any chance we could drop this function?

lib/charms/prometheus_k8s/v0/prometheus_remote_write.py
```diff
@@ -1027,32 +1052,24 @@ def alerts(self) -> dict:
...
-            try:
-                scrape_metadata = json.loads(relation.data[relation.app]["scrape_metadata"])
-                identifier = ProviderTopology.from_relation_data(scrape_metadata).identifier
-                alerts[identifier] = self._transformer.apply_label_matchers(alert_rules)
-            except KeyError as e:
-                logger.warning(
-                    "Relation %s has no 'scrape_metadata': %s",
-                    relation.id,
-                    e,
-                )
```
sed-i, line 1039: Could you please briefly outline why this section is dropped?

tests/unit/prometheus_alert_rules/with_template_string_and_unit.rule
sed-i: That doesn't make it correct :)
What is "template string"? Is it the topology stub or something else?

lib/charms/prometheus_k8s/v0/prometheus_remote_write.py
dstathis: No we can't drop it. I think this will create a false positive seldom enough that it should be okay.

lib/charms/prometheus_k8s/v0/prometheus_remote_write.py
dstathis: The way that we previously built the identifier was to look at `scrape_metadata`. This did not work for rules that were being forwarded from other charms as the scrape metadata was for the agent. Given that our libraries fill in the topology labels automatically, It is okay to expect such labels to be there and simply use those to create the identifier.

tests/unit/prometheus_alert_rules/with_template_string_and_unit.rule
dstathis: Yes it is the topology string. I'm not going to change Balbir's naming scheme here. If you think it is important, open a new PR

lib/charms/prometheus_k8s/v0/prometheus_remote_write.py
rbarry82: We all know I love regexps, but there are probably nicer ways to do this. Either just treat hex like hex and use multiple groups for the middle:

```python
r'[\da-f]{8}-([\da-f]{4}-){3}[\da-f]{12}'
```

If we can pull it out, just:
```python
uuid.UUID(str(val)) or ...
``` 

Really, though, I worry about the greedy matchers without a stop. `.*_` is not cool without `?`. Also please use raw strings with regexps.

```python
modified_matcher = re.compile(r'^.*?_[\da-f]{8}-([\da-f]{4}-){3}[\da-f]{12}_.*?alerts$')
if modified_matcher.match(name): ...
```

it's more complex looking from the outset, but less troublesome.

That said, since we know what position in the formatted string this _should_ be in, why don't we split it and check against `uuid.UUID()`? This is ok, and I'm giving it :+1:, but it's a little bit "write-only"

rbarry82 (approved):

tests/unit/test_remote_write.py
dstathis: I find this way to be more clear. If I'm outvoted, I'll change it

tests/unit/test_remote_write.py
dstathis: Fixed.

tests/unit/test_remote_write.py
dstathis: Fixed.

rbarry82 (approved):

Abuelodelanada (approved):

tests/unit/test_remote_write.py
dstathis: The `else` is paired with the `for` not the `if`

lib/charms/prometheus_k8s/v0/prometheus_remote_write.py
dstathis: Used Ryan's

PR #255, opened 2022-03-29
Explain rationale for not using some prometheus scrape settings in charms

mmanciop:
## Issue

clarify what some of the allowed settings are supported by the `prometheus_scrape` library, but should not be used directly by charmers.

## Solution

More docu.

## Context

We have seen PRs for charms like Kubeflow exposing config options for scrape-interval that should not be exposed directly by their charms.

## Testing Instructions

N/A

## Release Notes

Provide guidance in terms of which Job settings should be used by charmers integrating the library, and which not.

### DIFF
--- a/lib/charms/prometheus_k8s/v0/prometheus_scrape.py
+++ b/lib/charms/prometheus_k8s/v0/prometheus_scrape.py
@@ -309,7 +309,7 @@ over unit relation data using the `prometheus_scrape_unit_name` and
 `scrape_jobs` and `alert_rules` keys in application relation data
 of Metrics provider charms hold eponymous information.
 
-"""
+"""  # noqa: W505
 
 import json
 import logging
--- a/lib/charms/prometheus_k8s/v0/prometheus_scrape.py
+++ b/lib/charms/prometheus_k8s/v0/prometheus_scrape.py
@@ -162,11 +162,9 @@ For instance, if you include variable elements, like your `unit.name`, it may br
 the continuity of the metrics time series gathered by Prometheus when the leader unit
 changes (e.g. on upgrade or rescale).
 
-It is also possible to configure other scrape related parameters using
-these job specifications as described by the Prometheus
-[documentation](https://prometheus.io/docs/prometheus/latest/configuration/configuration/#scrape_config).
-The permissible subset of job specific scrape configuration parameters
-supported in a `MetricsEndpointProvider` job specification are:
+It is also technically possible, but **strongly discouraged**, to configure the following
+scrape-related parameters as described by the
+[Prometheus documentation](https://prometheus.io/docs/prometheus/latest/configuration/configuration/#scrape_config).
 
 - `job_name`
 - `metrics_path`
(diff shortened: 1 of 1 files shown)


lib/charms/prometheus_k8s/v0/prometheus_scrape.py
```diff
@@ -181,6 +179,16 @@ def __init__(self, *args):
 - `label_name_length_limit`
 - `label_value_length_limit`
 
+These options are technically allows to be
+specified over the relation interface to enable
+facilities like the [Prometheus Scrape Config](https://charmhub.io/prometheus-scrape-config-k8s)
+charm. These settings **should not** be exposed
```
simskij, line 185: This sentence is very difficult to digest.

simskij (changes requested):

mmanciop:
@sed-i how do I get to tell `pflake8` not do bother me about the length of ONE line that includes ONE URL? :D

I cannot add `# noqa:W505` above the offending line because in CharmHub, that will declare a top-level section LOL

lib/charms/prometheus_k8s/v0/prometheus_scrape.py
```diff
@@ -162,14 +162,10 @@ def __init__(self, *args):
...
 changes (e.g. on upgrade or rescale).
 
-It is also possible to configure other scrape related parameters using
-these job specifications as described by the Prometheus
-[documentation](https://prometheus.io/docs/prometheus/latest/configuration/configuration/#scrape_config).
-The permissible subset of job specific scrape configuration parameters
-supported in a `MetricsEndpointProvider` job specification are:
+Additionally, it is also technically possible, but **strongly discouraged**, to
+configure the following scrape-related settings, which behave as described by the
+[Prometheus documentation](https://prometheus.io/docs/prometheus/latest/configuration/configuration/#scrape_config):
```
rbarry82, line 167: This is possibly a reasonable case for shortcut reference links in Markdown.

Alternatively, I suspect that the following will actually look "normal":
```python
configure the following scrape-related settings, which behave as described by the
 [Prometheus documentation](
 https://prometheus.io/docs/prometheus/latest/configuration/configuration/#scrape_config):
```

Same for below.

rbarry82 (approved):

lib/charms/prometheus_k8s/v0/prometheus_scrape.py
```diff
@@ -308,7 +309,7 @@ def _on_scrape_targets_changed(self, event):
 `scrape_jobs` and `alert_rules` keys in application relation data
 of Metrics provider charms hold eponymous information.
 
-"""
+"""  # noqa: W505
```
dstathis, line 312: I'm approving but you should fix the doc string length or we should remove the requirement.

dstathis (approved):

lib/charms/prometheus_k8s/v0/prometheus_scrape.py
mmanciop: Do I feel like I should take the chance to see how borked markdown references are in the CharmHub / Discourse rendering? Make a guess :-)

PR #259, opened 2022-04-12
Add rerelate test (adapted from Loki #139)

sed-i:
## Issue
1. Check if the difficulties in [Loki #139](https://github.com/canonical/loki-k8s-operator/pull/139) are limited to loki or not.
1. Improve test coverage.

## Solution
Adapt the rerelate test for prometheus.

## Testing Instructions
The test passes locally with Juju 2.9.29, on a 2cpu-4g VM.

1. `multipass launch --cpus 2 --mem 4G --name charm-dev-2cpu-4g charm-dev`
1. `sudo snap refresh juju --channel=2.9/edge`
1. `tox -e integration -- --keep-models -k test_rerelate`

## Release Notes
Add integration test for re-relating charms.


balbirthomas:
So this test was passing when the re-relate was between alertmanager and prometheus, but when you added grafana-agent and grafana into the re-relate test the test started failing. Is that right ? If so is this difference in behaviour repeatable ?

sed-i:
> So this test was passing when the re-relate was between alertmanager and prometheus, but when you added grafana-agent and grafana into the re-relate test the test started failing. Is that right ? If so is this difference in behaviour repeatable ?

I think all those bad file descriptor / websocket tracebacks are because juju is stuck and test times out on
```
ERROR juju.worker.caasapplicationprovisioner.runner exited "grafana": failed to watch for changes to application "grafana": application "grafana" not found
```
Re-running, let's see.

balbirthomas:
> I think all those bad file descriptor / websocket tracebacks are because juju is stuck and test times out on

You certainly may be right but if we hope to overcome the problems these flaky tests are causing us, it behoves us try and pin down the prices cause and either request escalation on specific Python Libjuju, Juju or OF tickets or fix it in our tests.

simskij:
@sed-i I'll retrigger the integration tests on this one to see whether it passes now with all the fixes done in Juju and python-libjuju. Is there anything left to do on this or can we remove the draft status after that?

rbarry82:
Needed a merge with main to fix up all of the unit test failures and pin to an older juju (without the scaling bugs) in the integration tests. Let's see how it goes

dstathis (approved):

tests/integration/test_rerelate.py
```diff
@@ -0,0 +1,166 @@
...
+        ),
+    )
+
+    await asyncio.gather(
+        ops_test.model.add_relation(app_name, tester_app_name),
+        ops_test.model.add_relation(app_name, "alertmanager"),
+        ops_test.model.add_relation(app_name, "grafana:grafana-source"),
+        ops_test.model.add_relation(app_name, "grafana-agent:send-remote-write"),
+    )
+    await ops_test.model.wait_for_idle(apps=[app_name], status="active", timeout=600)
```
Abuelodelanada, line 166: Shouldn't we wait for idle in all apps?

```python
apps_names = [app_name, tester_app_name, "alertmanager", "grafana", "grafana-agent"]


...


    await ops_test.model.wait_for_idle(apps=apps_names, status="active", timeout=600)
```

Abuelodelanada (approved):

tests/integration/test_rerelate.py
```diff
@@ -0,0 +1,166 @@
...
+
+@pytest.mark.abort_on_fail
+async def test_rerelate(ops_test: OpsTest):
+    await asyncio.gather(
+        ops_test.model.add_relation(app_name, tester_app_name),
+        ops_test.model.add_relation(app_name, "alertmanager"),
+        ops_test.model.add_relation(app_name, "grafana:grafana-source"),
+        ops_test.model.add_relation(app_name, "grafana-agent:send-remote-write"),
+    )
+    await ops_test.model.wait_for_idle(status="active", timeout=600)
```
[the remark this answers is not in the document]
rbarry82: I'm not sure why there are different tests for removing the relation, re-relating, then removing. We'd get a traceback no matter where, and an integration test with only two calls seems over-normalized.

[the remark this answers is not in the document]
rbarry82: As an aside, using the "actual" app names rather than "implicit all" maybe somewhat more reliable.

If it's not specified, it's implicitly `self.applications -> property(self._live_entity_map("applications"))`. The property will be hit every time, and the "live entity map" can change in the meantime, because it may be [updated by a watcher asynchronously](https://github.com/juju/python-libjuju/blob/bd9490d1c0abebf0fe324ecdfcb43e3b30ac993f/juju/model.py#L1044), and properties [assigned to variables](https://github.com/juju/python-libjuju/blob/bd9490d1c0abebf0fe324ecdfcb43e3b30ac993f/juju/model.py#L2446) are called every time. Like [in a `while True:` loop](https://github.com/juju/python-libjuju/blob/bd9490d1c0abebf0fe324ecdfcb43e3b30ac993f/juju/model.py#L2472)

Relying on the implicit here without any synchronization primitives or libjuju actually snapshotting the state of the value is almost asking for `apps` to be different every time it wakes up from sleep. We should be explicit.

tests/integration/test_rerelate.py
```diff
@@ -0,0 +1,165 @@
...
+    # Block until it is really gone. Added after an itest failed when tried to redeploy:
+    # juju.errors.JujuError: ['cannot add application "...": application already exists']
+    try:
+        await ops_test.model.block_until(
+            lambda: tester_app_name not in ops_test.model.applications,
+            lambda: "alertmanager" not in ops_test.model.applications,
+            lambda: "grafana" not in ops_test.model.applications,
+            lambda: "grafana-agent" not in ops_test.model.applications,
+            timeout=300,
+        )
```
rbarry82, line 126: `remove` is an alias to `destroy()`, so we may as well use that, though libjuju uses them inconsistently.

Doing this from a cross-model relation may also avoid this (or at least put the tester/agent there), since you can just destroy that model and create a new one.

That said: 

```python
await asyncio.gather(
    ops_test.model.remove_application(tester_app_name, block_until_done=True),
    ops_test.model.remove_application("alertmanager", block_until_done=True),
    ...
)
```

Cleans up the code here a little bit, with the cost of not being able to specify the timeout. 

I imagine that we're actually going to check the data inside Prometheus at some point in these tests, right? Otherwise this is just exercising the negative case from the Juju bug which took us a month. That's ok, but we don't need an explicit test for it here, because we'd see it elsewhere.

But this specific case is actually one of the Juju bugs which only seems to come up in GH runners. Is this actually verifying anything which breaking/re-creating the relation doesn't already do?

sed-i: - Yeah should probably add reldata checks at some point in the future.
- The loki itest is also [waiting](https://github.com/canonical/loki-k8s-operator/blob/d29139883a65149b910cc2f558a3cc4a623ff150/tests/integration/test_rerelate.py#L130) for all apps to be removed, and is also not using `-new` in the app name.
- itest here now fail so I'll add the forced removal like in loki.

sed-i: > Is this actually verifying anything which breaking/re-creating the relation doesn't already do?
The intention here is to capture any possible differences in event sequencing or reldata stucture between remove relation and remove app (without manually remove relation first). It's a bit more "holistic" I guess and as a side effect trips on the juju bug.

rbarry82 (approved):

Issue #258, opened 2022-04-08
Update status should do a health check


balbirthomas:
At present, the Prometheus charm update status hook does not really validate the healthy status of the Prometheus workload. This may be useful functionality to have.

balbirthomas:
Assigned to @dstathis since this is alread work in progress in [PR265](https://github.com/canonical/prometheus-k8s-operator/pull/265). 

@dstathis I hope you have seen @jnsgruk comment on your PR.

simskij:
We've decided to not proceed with this at this point. Once pebble health checks and liveliness checks allow for emitting events, we'll likely revisit this.

Issue #262, opened 2022-04-23
[pod-spec] prometheus_scrape_unit_address may be set to None


phvalguima:
### Bug Description

For pod-spec charms, the MetricsEndpointProvider may return None to prometheus_scrape_unit_address on:
```
    def _set_unit_ip(self, _):
...
        for relation in self._charm.model.relations[self._relation_name]:
            relation.data[self._charm.unit]["prometheus_scrape_unit_address"] = str(
                self._charm.model.get_binding(relation).network.bind_address
            )
```

Juju will only return an IP when the application pod defined on the pod-spec starts. One possible fix is to check if the bind_address is set to None and raise an exception:
```
    def _set_unit_ip(self, _):
...
        if not self._charm.model.get_binding(relation).network.bind_address:
            raise PrometheusProviderBindAddressNotAvailable()
        for relation in self._charm.model.relations[self._relation_name]:
            relation.data[self._charm.unit]["prometheus_scrape_unit_address"] = str(
                self._charm.model.get_binding(relation).network.bind_address
            )
```

### To Reproduce

1. juju deploy ch:argo-controller
2. juju deploy ch:minio
3. juju deploy ch:prometheus-k8s --channel=latest/edge
4. juju add-relation argo-controller prometheus-k8s
Argo-controller will not yet be ready to spin up the application container as minio relation is not set yet, prometheus will receive prometheus_scrape_unit_address=None

### Environment

microk8s v1.21
juju v2.9.28

### Relevant log output

```shell
$ juju show-unit prometheus-k8s/0
prometheus-k8s/0:
  opened-ports: []
  charm: local:focal/prometheus-k8s-0
  leader: true
  relation-info:
  - endpoint: prometheus-peers
    related-endpoint: prometheus-peers
    application-data: {}
    local-unit:
      in-scope: true
      data:
        egress-subnets: 10.152.183.74/32
        ingress-address: 10.152.183.74
        private-address: 10.152.183.74
  - endpoint: metrics-endpoint
    related-endpoint: metrics-endpoint
    application-data:
      alert_rules: '{"groups": [{"name": "argo_62a27dd7-6426-4fc4-85ba-4470c63f53e0_argo-controller_argo-controller_unit_unavailable_alerts",
        "rules": [{"alert": "ArgoUnitIsUnavailable", "expr": "up < 1", "for": "0m",
        "labels": {"severity": "critical", "juju_model": "argo", "juju_model_uuid":
[...]

PR #265, opened 2022-04-27
Prometheus health checks

dstathis:
## Issue
#258 

## Solution
Run the built in health checks and set status accordingly.

## Testing Instructions
Attempt to make Prometheus unhealthy and ensure that the status changes correctly.

## Release Notes
Run health checks during the update_status Juju hook.

### DIFF
--- a/src/charm.py
+++ b/src/charm.py
@@ -25,7 +25,7 @@ from charms.prometheus_k8s.v0.prometheus_scrape import MetricsEndpointConsumer
 from charms.traefik_k8s.v0.ingress_per_unit import IngressPerUnitRequirer
 from ops.charm import CharmBase
 from ops.main import main
-from ops.model import ActiveStatus, BlockedStatus, WaitingStatus
+from ops.model import ActiveStatus, BlockedStatus, UnknownStatus, WaitingStatus
 from ops.pebble import ChangeError, Layer
 
 from prometheus_server import Prometheus
@@ -89,6 +89,7 @@ class PrometheusCharm(CharmBase):
         self.framework.observe(self.on.receive_remote_write_relation_broken, self._configure)
         self.framework.observe(self.metrics_consumer.on.targets_changed, self._configure)
         self.framework.observe(self.alertmanager_consumer.on.cluster_changed, self._configure)
+        self.framework.observe(self.on.update_status, self._update_status)
 
     def _configure(self, _):
         """Reconfigure and either reload or restart Prometheus.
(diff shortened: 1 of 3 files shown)


rbarry82:
Once static checks pass, :+1: from me

src/charm.py
```diff
@@ -160,6 +161,26 @@ def _configure(self, _):
 
         self.unit.status = ActiveStatus()
 
+    def _update_status(self, _):
+        """Update the status of the charm."""
+        # Run configure to reset the status
+        self._configure(_)
```
balbirthomas, line 167: I am surprised by the need to run configure in update status. That the purpose is to "reset the status". But isn't that the whole purpose of `_update_status()`. Making the Prometheus charm reload its configuration every time an update status event triggers seems unnecessary.

tests/unit/test_server.py
```diff
@@ -94,3 +94,51 @@ def test_prometheus_server_with_route_prefix_reload_configuration_failure(self):
         )
 
         self.assertFalse(self.prometheus.reload_configuration())
+
+    @responses.activate
```
balbirthomas, line 98: Wouldn't it be more valuable to test that in response to a update status event the Prometheus charm sets blocked or unknown status when Prometheus server is not healthy or not ready respectively. 

This and the following unit tests only check rather trivial behaviors of a single if statement in two methods. Note I am not asking for removal of these unit tests. I am trying to say the more vital unit tests is missing.

balbirthomas (changes requested):

jnsgruk:
Just a passing note: can we solve this problem better with Pebble health checks now they're here? https://juju.is/docs/sdk/pebble?

src/charm.py
```diff
@@ -160,6 +161,26 @@ def _configure(self, _):
...
+        external_url = urlparse(self._external_url)
+        prometheus_server = Prometheus(web_route_prefix=external_url.path)
+
+        if not prometheus_server.is_healthy():
+            self.unit.status = BlockedStatus("Prometheus failed health check")
+            return
+
+        if not prometheus_server.is_ready():
+            self.unit.status = UnknownStatus("Prometheus is not ready")
+            return
```
sed-i, line 182: Just curious why Unknown here vs. Blocked above.
For example, Unknown is used by juju when the charm doesn't set any status.
Here the charm is communicating and prom is responding. Maybe Blocked is more consistent?

src/charm.py
sed-i: `_configure` might be needed here for https://github.com/canonical/prometheus-k8s-operator/issues/262

src/charm.py
```diff
@@ -160,6 +161,26 @@ def _configure(self, _):
...
+        if not isinstance(self.unit.status, ActiveStatus):
+            # Status is already not active so do nothing.
+            return
+
+        external_url = urlparse(self._external_url)
+        prometheus_server = Prometheus(web_route_prefix=external_url.path)
+
+        if not prometheus_server.is_healthy():
+            self.unit.status = BlockedStatus("Prometheus failed health check")
+            return
```
sed-i, line 178: Should we have some more serious side effects here?
Unless we rely on self-monitoring?

dstathis:
> Just a passing note: can we solve this problem better with Pebble health checks now they're here? https://juju.is/docs/sdk/pebble?

Unfortunately the only things that pebble health checks let us do is shut down or restart which I think would not be particularly helpful.

src/charm.py
dstathis: If a health check fails then becomes healthy again, we want to change back to active state. Without _configure we have no way to do that

balbirthomas: Prometheus status may be set to blocked status if it obtains an invalid configuration from the command line or through relation data. 

Is it not possible for the update status event handler to check that Prometheus is not currently in blocked status due to these reasons and only then set active status if Prometheus is healthy and ready ?

Invoking configure every time in update status implies every time an update status event fires Prometheus will reload its configuration. Could this be a problem ?

rbarry82: As @balbirthomas said: Why don't we actually... check whether #262 has happened or what the state is/should be before blindly calling `_configure`?

dstathis:
After a fair amount of discussion I think it is fairly clear that this will require a lot more work and thought than should be spent at the moment. I will close this PR but leave the issue open to be revisited later.

PR #268, opened 2022-05-04
MetricsProvider sets unit address on updated status

balbirthomas:
## Issue
The `MetricsEndpointProvider` only set unit IP addresses in response to pebble ready events. Pod spec charms did not have pebble ready events and hence `MetricsEndpointConsumer` could receive unit address such as `None`.

## Solution

This PR ensures that the `MetricsEndpointProvider` sets
unit address on update status and config changed events also.
This change is required because podspec charms do not have
a pebble ready event and uptil now the `MetricsEndpointProvider`
was setting IP address only in response to pebble ready event
This change also ensures that the `MetricsEndpointProvider`
does not set invalid IP addresses.

## Context
Each `MetricsEndpointProvider` unit is a Prometheus scrape target for `MetricsEndpointConsumer` and the former needs to provide the an IP address from which Prometheus may scrape metrics.

## Testing Instructions
To be tested with original reporters issue.

## Release Notes
Prometheus operator charm now supports scraping metrics from podspec charms also.


tests/unit/test_endpoint_provider.py
```diff
@@ -152,6 +152,22 @@ def test_provider_unit_sets_bind_address_on_pebble_ready(self, *unused):
...
+    @patch_network_get(private_address="192.0.8.2")
+    def test_provider_unit_sets_bind_address_on_update_status(self, *unused):
+        rel_id = self.harness.add_relation(RELATION_NAME, "provider")
+        self.harness.charm.on.update_status.emit()
+        data = self.harness.get_relation_data(rel_id, self.harness.charm.unit.name)
+        self.assertIn("prometheus_scrape_unit_address", data)
+        self.assertEqual(data["prometheus_scrape_unit_address"], "192.0.8.2")
+
+    @patch_network_get(private_address=None)
+    def test_provider_does_not_set_invalid_unit_address(self, *unused):
```
sed-i, line 164: Is the function name up to date? It says "not invalid".

lib/charms/prometheus_k8s/v0/prometheus_scrape.py
```diff
@@ -1611,13 +1616,33 @@ def _set_unit_ip(self, _):
...
+            if not self._is_valid_unit_address(unit_ip):
+                # relation data will be updated later when a valid address becomes available
+                continue
+
+            relation.data[self._charm.unit]["prometheus_scrape_unit_address"] = unit_ip
             relation.data[self._charm.unit]["prometheus_scrape_unit_name"] = str(
                 self._charm.model.unit.name
             )
 
+    def _is_valid_unit_address(self, address: str) -> bool:
```
sed-i, line 1630: I think we can trust ops to deliver on their promise for `bind_address`: either None or an address.
I.e. I would drop this function and use `if bind_address` inline.

tests/unit/test_endpoint_provider.py
balbirthomas: It says provider *does not set* invalid unit address. Am I missing something ?

lib/charms/prometheus_k8s/v0/prometheus_scrape.py
balbirthomas: I see no harm in actually validating the IP address.

lib/charms/prometheus_k8s/v0/prometheus_scrape.py
```diff
@@ -1611,13 +1616,33 @@ def _set_unit_ip(self, _):
...
         """
         for relation in self._charm.model.relations[self._relation_name]:
-            relation.data[self._charm.unit]["prometheus_scrape_unit_address"] = str(
-                self._charm.model.get_binding(relation).network.bind_address
-            )
+            unit_ip = str(self._charm.model.get_binding(relation).network.bind_address)
+
+            if not self._is_valid_unit_address(unit_ip):
+                # relation data will be updated later when a valid address becomes available
+                continue
```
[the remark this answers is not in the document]
balbirthomas: I am averse to setting any unit data at all, including empty string. I have added a clause to delete both unit address and name if present.

lib/charms/prometheus_k8s/v0/prometheus_scrape.py
```diff
@@ -1611,13 +1616,36 @@ def _set_unit_ip(self, _):
...
-            relation.data[self._charm.unit]["prometheus_scrape_unit_address"] = str(
-                self._charm.model.get_binding(relation).network.bind_address
-            )
+            unit_ip = str(self._charm.model.get_binding(relation).network.bind_address)
+
+            if not self._is_valid_unit_address(unit_ip):
+                # relation data will be updated later when a valid address becomes available
+                if "prometheus_scrape_unit_address" in relation.data[self._charm.unit]:
+                    del relation.data[self._charm.unit]["prometheus_scrape_unit_address"]
+                    del relation.data[self._charm.unit]["prometheus_scrape_unit_name"]
```
sed-i, line 1625: Technically this could throw a key error.
Maybe
```suggestion
                with contextlib.suppress(KeyError):
                    del relation.data[self._charm.unit]["prometheus_scrape_unit_address"]
                    del relation.data[self._charm.unit]["prometheus_scrape_unit_name"]
```

tests/unit/test_endpoint_provider.py
sed-i: I'm just confused:
- There is `private_address=None` and `assertNotIn("prometheus_scrape_unit_address" ...`, so it is invalid
- If it is not invalid, then why not call it "provider sets valid address"?

balbirthomas: The test is to ensure that `MetricsEndpointProvider` does not set `prometheus_scrape_unit_address` if it gets an invalid (eg. `None`) unit address. So `MetricsEndpointProvider` is giving a guarantee that if it sets a unit IP address, it will be valid.

lib/charms/prometheus_k8s/v0/prometheus_scrape.py
balbirthomas: FWIW please note changes suggested like this risk the introduction of a bug because they do not import `contextlib` into this module at the same time. I will make this change locally and make another push.

Abuelodelanada (approved):

PR #273, opened 2022-05-12
Don't emit a bunch of warning logs during normal operation

### LINKED ISSUE #266: Prometheus should emit debug rather than warning when alert rules directory is missing
It should also only happen in install and upgrade events.

rbarry82:
## Issue
Various alert rules checks currently emit terrifying-looking warning messages on every invocation of charms, even when "must provide alert rules" is not a requirement.

Closes #266 

## Solution
Use logger.debug instead

## Testing Instructions
Deploy. Make sure there aren't scary warning logs if alert rules aren't present.

## Release Notes
Don't emit  warning logs during normal operation

### DIFF
--- a/tests/unit/test_endpoint_consumer.py
+++ b/tests/unit/test_endpoint_consumer.py
@@ -436,7 +436,7 @@ class TestEndpointConsumer(unittest.TestCase):
         with self.assertLogs(level="WARNING") as logger:
             _ = self.harness.charm.prometheus_consumer.alerts()
             messages = logger.output
-            self.assertEqual(len(messages), 2)
+            self.assertEqual(len(messages), 1)
             self.assertIn(
                 "Alert rules were found but no usable group or identifier was present", messages[1]
             )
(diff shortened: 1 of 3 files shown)


sed-i (dismissed):
Looks like we got a few fragile unit tests.

Abuelodelanada (approved):

tests/unit/test_endpoint_consumer.py
```diff
@@ -436,9 +436,9 @@ def test_consumer_logs_an_error_on_missing_alerting_data(self):
         with self.assertLogs(level="WARNING") as logger:
             _ = self.harness.charm.prometheus_consumer.alerts()
             messages = logger.output
-            self.assertEqual(len(messages), 2)
+            self.assertEqual(len(messages), 1)
             self.assertIn(
-                "Alert rules were found but no usable group or identifier was present", messages[1]
+                "Alert rules were found but no usable group or identifier was present", messages[0]
```
sed-i, line 441: Are those assertions needed at all?

sed-i (approved):

tests/unit/test_endpoint_consumer.py
rbarry82: I mean, it's `test_consumer_logs_an_error_on_missing_alerting_data`, so checking the logs is kinda crucial.

dstathis (approved):
:+1: Very much needed!

PR #274, opened 2022-05-13
Add tests covering multi-unit prometheus deployments

### LINKED ISSUE #173: Get rid of the implicit assumption that this charm only has one unit
> Seems like this check could be handy in other charms.
That being said, iirc, this charm implicitly assumes in several places there is only one unit.

_Originally posted by @sed-i in https://github.com/canonical/prometheus-operator/pull/147#discussion_r753562148_

### LINKED ISSUE #184: A Growing list of Integration test cases

sed-i:
## Issue
Currently there are no tests covering multi-unit prometheus deployments.
Closes #173.
Closes #184 

## Solution
Add tests.

## Context
NTA.

## Testing Instructions
- `tox -e integration -- -k test_prometheus_scrape_multiunit`

## Release Notes
Add tests covering multi-unit prometheus deployments.


sed-i:
@rbarry82 mind taking a look at what I have so far?

tests/integration/test_prometheus_scrape_multiunit.py
```diff
@@ -0,0 +1,188 @@
...
+    )
+
+    # WHEN prometheus is not related to anything
+    # THEN all prometheus units should have only one scrape target (self-scraping)
+    for unit_num in range(num_units):
+        targets = await get_prometheus_active_targets(ops_test, prometheus_app_name, unit_num)
+        assert len(targets) == 1
+        self_scrape = next(iter(targets))
+        assert self_scrape["labels"]["job"] == "prometheus"
+        assert self_scrape["labels"]["instance"] == "localhost:9090"
```
rbarry82, line 81: This can/should be tidied up into a single in-file helper and/or `assert all(...)`, but not really important for now.

tests/integration/test_prometheus_scrape_multiunit.py
```diff
@@ -0,0 +1,188 @@
...
+    # WHEN prometheus is scaled up
+    num_additional_units = 1
+    await ops_test.model.applications[prometheus_app_name].scale(scale_change=num_additional_units)
+    new_num_units = num_units + num_additional_units
+
+    # THEN nothing breaks
+    await ops_test.model.wait_for_idle(
+        apps=[prometheus_app_name],
+        status="active",
+        timeout=600,
```
rbarry82, line 133: This timeout is probably really excessive. Even 300 is a lot better, but if we can't scale in 60 seconds or less, even on GH, something is dramatically wrong.

tests/integration/test_prometheus_scrape_multiunit.py
```diff
@@ -0,0 +1,188 @@
...
+    """Upgrade prometheus and confirm all is still green (see also test_upgrade_charm.py)."""
+    # WHEN prometheus is upgraded
+    await ops_test.model.applications[prometheus_app_name].refresh(
+        path=prometheus_charm, resources=prometheus_resources
+    )
+
+    # THEN nothing breaks
+    await ops_test.model.wait_for_idle(status="active", idle_period=60)
+    await asyncio.gather(
+        *[check_prometheus_is_ready(ops_test, prometheus_app_name, u) for u in range(num_units)]
```
rbarry82, line 118: It's a larger scope than just here but assertions in a helper method are impossible. Without actually looking at it, a reviewer could wonder whether an assertion was missed in this, the actual test.

`check_prometheus_is_ready` -> `check_prometheus_readiness(...) -> bool`, and get the assertions in the actual tests.

rbarry82:
The changes here are broadly good, though I personally think "is the Prometheus endpoint available?" isn't a really sufficient test.

I guess the "ideal" would probably be something like:

* Deploy a single Prometheus and tester. Verify we have metrics.
* Scale Prometheus up to 3 units. Verify they all have metrics (this could just be a further condition on the previous test).
* Get Prometheus to fire an alert. Verify all 3 are firing. (could also be part of the previous test)
* Upgrade Prometheus. Without waiting for idle, also scale up the tester so a `relation-joined` fires while it's in the middle of doing things, or start an upgrade of the tester first with different alert rules to make sure that Prometheus behaves "as expected" when there's a collision between relation events and upgrade events.
* Make sure metrics keep continuity and alerts which were firing are still firing (could be additional assertions in the scaling test)
* Scale the tester up and make sure the new tester's metrics show up

Scaling down is much harder to test since we don't necessarily know "which" tester is going to go away, but this way we exercise the _functionality_ of Prometheus rather than just "does it exist"?

tests/integration/test_prometheus_scrape_multiunit.py
sed-i: Yes but then the assertion failures would be too vague.

tests/integration/test_prometheus_scrape_multiunit.py
sed-i: Agreed. Waiting for your PR to merge to bring it in :)

tests/integration/test_prometheus_scrape_multiunit.py
sed-i: Done in next commit.

sed-i:
> "is the Prometheus endpoint available?" isn't a really sufficient test.

Agreed. What I currently have under `# THEN nothing breaks` is mostly a placeholder and we should add more.


> Get Prometheus to fire an alert. Verify all 3 are firing.

It seems wrong to me to include this here, because we want to stay away from all-in-one tests:
1. Prometheus firing an alert is a "prometheus thing" that is tested in a different itest file.
2. As long as all the configs are the same for all units (which IS tested here), then I think we can safely assume they all behave the same.

> Upgrade Prometheus. Without waiting for idle, also scale up the tester so a `relation-joined` fires while it's in the middle of doing things, or start an upgrade of the tester first with different alert rules to make sure that Prometheus behaves "as expected" when there's a collision between relation events and upgrade events.
>
> Make sure metrics keep continuity and alerts which were firing are still firing (could be additional assertions in the scaling test)
>
> Scale the tester up and make sure the new tester's metrics show up

Sounds like a good tests but I'm not sure if it's better to have them here or in `test_upgrade_charm`.

rbarry82:
> It seems wrong to me to include this here, because we want to stay away from all-in-one tests:
> 
> 1. Prometheus firing an alert is a "prometheus thing" that is already tested in a different itest file.
> 2. As long as all the configs are the same for all units (which IS tested here), then I think we can safely assume they all behave the same.

I think that was sort of the discussion earlier about an e2e test just for each test, without adding in things from the bundle.

We test each piece in relative isolation, but it's like testing the brakes/steering linkage/drivetrain of a car without testing to make sure it actually drives. We don't need `test_...` for each test, but for each "major" step, because we have already been bitten by "safely assume" elsewhere.
 
> > Upgrade Prometheus. Without waiting for idle, also scale up the tester so a `relation-joined` fires while it's in the middle of doing things, or start an upgrade of the tester first with different alert rules to make sure that Prometheus behaves "as expected" when there's a collision between relation events and upgrade events.
> > Make sure metrics keep continuity and alerts which were firing are still firing (could be additional assertions in the scaling test)
> > Scale the tester up and make sure the new tester's metrics show up
> 
> Sounds like a good tests but I'm not sure if it's better to have them here or in `test_upgrade_charm`.

Since `test_upgrade_charm` doesn't deal with scaling or possibly async/staggered relation events if not all of the upgrades complete at the same time (and they probably won't). All other extant tests aren't trying to scale _and_ upgrade _and_ make sure that we don't have a break in upgrades, and, for better or worse, this is a very viable "real world" case if script refreshes a model while once it's verified in dev/preprod/whatever.

sed-i:
To verify metrics continuity, I'm thinking:
`numSeries` and `numLabelPairs` remains the exact same across a prom upgrade.

```shell
curl -s 10.1.79.206:9090/api/v1/status/tsdb | jq ".data.headStats"
{
  "numSeries": 677,
  "numLabelPairs": 378,
  "chunkCount": 5950,
  "minTime": 1652717360704,
  "maxTime": 1652723525883
}
```

WDYT @rbarry82?

sed-i:
> To verify metrics continuity, I'm thinking: `numSeries` and `numLabelPairs` remain the exact same across a prom upgrade.
> 
> ```shell
> curl -s 10.1.79.206:9090/api/v1/status/tsdb | jq ".data.headStats"
> {
>   "numSeries": 677,
>   "numLabelPairs": 378,
>   "chunkCount": 5950,
>   "minTime": 1652717360704,
>   "maxTime": 1652723525883
> }
> ```
> 
> WDYT @rbarry82?

Surprisingly these are not as stable as I hoped: sometimes two prom units have different values, and sometimes the values before and after the upgrade differ.

```
  File "/home/ubuntu/code2/prometheus-k8s-operator/tests/integration/test_prometheus_scrape_multiunit.py",
  line 157, in test_upgrade_prometheus
    assert head_stats_before == head_stats_after
AssertionError: assert {Stats(numSeries=499, numLabelPairs=352)} == 
                       {Stats(numSeries=499, numLabelPairs=352), Stats(numSeries=527, numLabelPairs=358)}
  Extra items in the right set:
  Stats(numSeries=527, numLabelPairs=358)

```

```
  File "/home/ubuntu/code/prometheus-k8s-operator/tests/integration/test_prometheus_scrape_multiunit.py",
  line 157, in test_upgrade_prometheus
    assert head_stats_before == head_stats_after
AssertionError: assert {Stats(numSeries=507, numLabelPairs=356)} == {Stats(numSeries=535, numLabelPairs=362)}
```

Ideas?

tests/integration/test_prometheus_scrape_multiunit.py
```diff
@@ -0,0 +1,342 @@
...
+num_units = 2  # Using the same number of units for both prometheus and the tester
+
+# The period of time required to be idle before `wait_for_idle` returns is set to 90 sec because
+# unit upgrades were observed to take place 40-70 seconds apart.
+idle_period = 90
+
+
+async def test_setup_env(ops_test: OpsTest):
+    await ops_test.model.set_config(
+        {"logging-config": "<root>=WARNING; unit=DEBUG", "update-status-hook-interval": "60m"}
```
rbarry82, line 49: Why 60m?

tests/integration/test_prometheus_scrape_multiunit.py
rbarry82: Regardless of how vague the assertions are, a failure would require manual investigation or fixing the problematic patch anyway.

Especially considering that we don't actually indicate which unit failed, and that Prometheus self-scrape should _always_ be there. If we're really worried about it, make it `assert all(...), "No self-scrape jobs present!"`

rbarry82:
> > To verify metrics continuity, I'm thinking: `numSeries` and `numLabelPairs` remain the exact same across a prom upgrade.
> > ```shell
> > curl -s 10.1.79.206:9090/api/v1/status/tsdb | jq ".data.headStats"
> > {
> >   "numSeries": 677,
> >   "numLabelPairs": 378,
> >   "chunkCount": 5950,
> >   "minTime": 1652717360704,
> >   "maxTime": 1652723525883
> > }
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > WDYT @rbarry82?
> 
> Surprisingly these are not as stable as I hoped: sometimes two prom units have different values, and sometimes the values before and after the upgrade differ.
> 
> ```
>   File "/home/ubuntu/code2/prometheus-k8s-operator/tests/integration/test_prometheus_scrape_multiunit.py",
>   line 157, in test_upgrade_prometheus
>     assert head_stats_before == head_stats_after
> AssertionError: assert {Stats(numSeries=499, numLabelPairs=352)} == 
>                        {Stats(numSeries=499, numLabelPairs=352), Stats(numSeries=527, numLabelPairs=358)}
>   Extra items in the right set:
>   Stats(numSeries=527, numLabelPairs=358)
> ```
> 
>  ```
>   File "/home/ubuntu/code/prometheus-k8s-operator/tests/integration/test_prometheus_scrape_multiunit.py",
>   line 157, in test_upgrade_prometheus
>     assert head_stats_before == head_stats_after
> AssertionError: assert {Stats(numSeries=507, numLabelPairs=356)} == {Stats(numSeries=535, numLabelPairs=362)}
> ```
> 
> Ideas?

I would pretty much always expect the stats after to be greater, so that's a fair assertion.

It's more concerning that they differ. Guess would be that `head_stats_before` and `head_stats_after` should be lists, not set (or be dict with each unit as the key), and "extra items in the right set" is because the units diverged when one upgraded/came up faster, so the sets didn't ahve the same values.

tests/integration/test_prometheus_scrape_multiunit.py
sed-i: So apparently update-status was crucial for robustness
```
  tester/0 [idle] blocked: Waiting for Pebble ready
  tester/1 [idle] blocked: Waiting for Pebble ready
```

tests/integration/test_prometheus_scrape_multiunit.py
sed-i: A fresh example: comparison of configuration failed with
```
    assert all(config_by_unit[0] == config_by_unit[u] for u in range(1, num_units))
AssertionError: assert False
 +  where False = all(<generator object test_prometheus_scrape_relation_with_prometheus_tester.<locals>.<genexpr> at 0x7f35547184a0>)
```
and the more explicit form would allow skipping a step in future failures, as the diff itself is printed.

rbarry82 (dismissed):

rbarry82 (approved):

tests/integration/test_prometheus_scrape_multiunit.py
```diff
@@ -0,0 +1,385 @@
...
+    "prometheus-tester-image": oci_image(
+        "./tests/integration/prometheus-tester/metadata.yaml",
+        "prometheus-tester-image",
+    )
+}
+num_units = 2  # Using the same number of units for both prometheus and the tester
+
+# The period of time required to be idle before `wait_for_idle` returns is set to 90 sec because
+# unit upgrades were observed to take place 40-70 seconds apart.
+idle_period = 90
```
dstathis, line 48: I think if you saw upgrades taking 70 seconds, I'd go for something more like 200s. 90s seems to be cutting it kind of close.

dstathis (approved):

.github/workflows/pull-request.yaml
```diff
@@ -99,6 +99,7 @@ jobs:
       - name: Setup operator environment
         uses: charmed-kubernetes/actions-operator@main
         with:
+          channel: 1.23/stable
```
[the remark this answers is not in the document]
sed-i: Removed in next commit.

.github/workflows/release-edge.yaml
```diff
@@ -99,6 +99,7 @@ jobs:
       - name: Setup operator environment
         uses: charmed-kubernetes/actions-operator@main
         with:
+          channel: 1.23/stable
```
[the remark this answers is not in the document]
sed-i: Removed in next commit.

tests/integration/helpers.py
```diff
@@ -253,3 +288,59 @@ async def __aenter__(self):
...
+        app_data_only: limit output to application relation data.
+    See https://github.com/juju/python-libjuju/issues/642.
+    """
+    endpoint_arg = f"--endpoint {endpoint}" if endpoint else ""
+    related_unit_arg = f"--related-unit {related_unit}" if related_unit else ""
+    app_data_arg = "--app" if app_data_only else ""
+    cmd = filter(
+        None,
+        f"juju show-unit {unit_name} {endpoint_arg} {related_unit_arg} {app_data_arg}".split(" "),
+    )
```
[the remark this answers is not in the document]
sed-i: Good point, but this function is not used anymore so I'm removing it entirely in the next commit.

tests/integration/test_prometheus_scrape_multiunit.py
sed-i: Increasing `idle_period` would increase the time it takes itests to run.
I think that as long as we don't know how this is set by juju, might as well keep it lower than higher and increase when the need arises.

dstathis (approved):

rbarry82 (approved):

PR #276, opened 2022-05-14
Resolve integration test flakiness

### LINKED ISSUE #197: Refactor time based flaky integration tests
## Integration tests may fail if Github workflow is slow

A recent commit introduced tests that use an extended `timeout` value. These timeout values were extended because, when Github workflows are slow the test fails sporadically. However the choice of a timeout value is arbitrary. Hence the tests may still fail if github gets even slower. The purpose of this ticket is to refactor these tests so that they are not timeout based. The tests in question are

1. https://github.com/canonical/prometheus-k8s-operator/blob/main/tests/integration/test_kubectl_delete.py
2. https://github.com/canonical/prometheus-k8s-operator/blob/main/tests/integration/test_upgrade_charm.py

### LINKED ISSUE #234: Update Metrics Consumer integration test
### Bug Description

At present the metrics consumer integration test `test_deploy_from_edge_and_upgrade_from_local_path()` only checks if the `MetricsEndpointConsumer` becomes `active` after and `juju refresh`. This test can be enhanced by also checking if after refresh the `MetricsEndpointConsumer` retains the same alert rules (and perhaps scrape jobs) as it had prior to the refersh. Also this test may be renamed to emphasise it is a upgrade test for the `MetricsEndpointConsumer`.

### To Reproduce

NA

### Environment

NA

### Relevant log output

```shell
NA
```

### Additional context

NA

rbarry82:
## Issue
Integration tests were full of workarounds for bugs which were no longer meaningful, and potentially overlapping time windows which resulted in `tenacity` being added to resolve inconsistencies.

Closes #197 
Closes #234 

## Solution
Remove `IPAddressWorkaround` and arbitrary very long timeouts.

Align `idle_timeout` with operations (upgrade, relation joins) which may need to succeed before data queries work.

## Testing Instructions
Run integration tests

## Release Notes
Integration test cleanup and removal of old FIXMEs.


tests/integration/test_charm.py
```diff
@@ -78,6 +73,45 @@ async def test_prometheus_scrape_relation_with_prometheus_tester(
...
+
+    await ops_test.model.applications[tester_app_name].refresh(
+        path=prometheus_tester_charm,
+        resources={
+            "prometheus-tester-image": oci_image(
+                "./tests/integration/prometheus-tester/metadata.yaml",
+                "prometheus-tester-image",
+            )
+        },
+    )
```
sed-i, line 98: This does not test alert rules changing across an upgrade because `set_config` -> `_configure` -> `_set_alerts` -> `push`, so the new alerts would be in effect before the upgrade.

Unfortuantely, `.refresh()` does not take a config dict. This was unexpected.

The Juju help CLI gives some hope:
```
Charm settings may be added or updated at upgrade time by specifying the
--config option, pointing to a YAML-encoded application config file.

  juju refresh foo --config config.yaml
```

Without this option in python-libjuju, I think this test is not giving us what we need.

sed-i (commented):
Looks good other than the currently absent `refresh(config=...)`.

tests/integration/test_charm.py
rbarry82: It does, actually. `_configure` is not involved in the hot path at all here. What we actually strike is `_set_scrape_job_spec` in `MetricsEndpointProvider` as part of `upgrade_charm`, and this invocation (after the upgrade) will have the alert rules from the config set, and the assertion after this verifies it.

It is _not_ testing whether upgrading Prometheus picks up alert rules from the config change (it won't, since the relation data will not be re-set unless the tester/provider is upgraded), but whether they can be changed at all now that the entire `write/remove/rebuild` pattern has been removed.

tests/integration/prometheus-tester/config.yaml
```diff
@@ -6,3 +6,7 @@ options:
     default: 1s
     description: "Prometheus job configuration for scrape interval"
     type: string
+  alert-rules-path:
+    default: src/prometheus_alert_rules
+    description: "Path for alert rules passed to the Provider"
+    type: string
```
dstathis, line 12: I'm not sure why we are making this configurable. Isn't the correct path already set on charm compile time? Or is this so we can point at different directories all of which are packed in to the charm?

dstathis (approved):

sed-i (approved):

tests/integration/prometheus-tester/config.yaml
rbarry82: It's so we can point at different directories packed in the charm instead of rebuilding the whole thing every time (which causes a mess with `charmcraft` caching)

Issue #104, opened 2021-08-20
Alert rules not resilient against leader failure


mmanciop:
If the leader unit of the charm with built-in alert rules experiences failure, alert rules are not reliably sent to Prometheus. We should consider updating alert rules in the relation data when the leader unit of the charm with built-in alert rules is elected.

balbirthomas:
You mean every time a new leader is elected for the `MetricsEndpointProvider` charm, if required alert rules should be updated.

mmanciop:
I am hazy on whether a `juju refresh` will always cause a new leader to be elected, especially when the Juju application has just one unit. So we might need both the `on.upgrade_charm` and `on.leader_elected` hooks to be registered and acted upon only by the new leader.

balbirthomas:
I tried to reproduce this issue as follows

1. Deploy prometheus-k8s and prometheus-tester adding a relation between the two
2. The used kubectl to delete the leader pod of prometheus (since `juju remove-unit` is not supported on k8s)

However Juju created a new pod without reassigning the leader to the other (non deleted) pod. There is a juju hook tool called `leader-set` but AFAIK this does not change the leader itself only created "leader setting" key value pairs. Hence I am not sure if there is controlled way to reproduce this issue.

It is of course trivial to add a leader elected event handler binding it to the config changed handler or upgrade handler. However since all relation data is stored as application relation data the newly elected leader (of `MetricsEndpointProvider`) does not have to do anything to assert leadership, since the application data should still persist.

balbirthomas:
I tried the following. 

1. Deployed 10 units of Prometheus (`juju deploy prometheus-k8s -n 10 --channel=edge`). Unit 3 was the leader.
2. Deployed [prometheus-tester](https://github.com/balbirthomas/prometheus-tester) and added a relation with Prometheus.
3. Checked that all Prometheus were scrapping the prometheus-tester target and had associated alert rules.
4. Scaled down Prometheus application to 1 (`juju scale-application prometheus-k8s 1`). Unit 0 became the leader.
5. Checked that Unit 0 was still being scrapped by Prometheus and still had the associated alert rules.

mmanciop:
And I assume your test worked @balbirthomas ? :-)

The failure scenario I had in mind actually is the following:

1. Deploy a new charm, which fails to elect a leader
2. Scrape jobs and alert rules are never sent to Prometheus because there is no leader that can write the app data bag, and Prometheus does not know anything about the missing jobs and alerts

[...]

PR #277, opened 2022-05-15
set relation data on leader elected so alert rules are properly set

dstathis:
## Issue
#104

## Solution
Set relation data as part of `leader_elected`

## Testing Instructions
General broken or not.

## Release Notes
Fixed a bug where alert rules are sometimes not set

### DIFF
--- a/tests/unit/test_endpoint_provider.py
+++ b/tests/unit/test_endpoint_provider.py
@@ -615,3 +615,25 @@ class TestAlertRulesContainingUnitTopology(unittest.TestCase):
             for rule in group["rules"]:
                 self.assertIn("juju_unit", rule["labels"])
                 self.assertIn("juju_unit=", rule["expr"])
+
+
+class TestNoLeader(unittest.TestCase):
+    """Tests the case where leader is not set immediately."""
+
+    def setUp(self):
+        self.harness = Harness(EndpointProviderCharm, meta=PROVIDER_META)
+        self.addCleanup(self.harness.cleanup)
+        self.harness.set_leader(False)
+        self.harness.begin_with_initial_hooks()
+
+    @patch("ops.testing._TestingModelBackend.network_get")
+    def test_alert_rules(self, _):
+        """Verify alert rules are added when leader is elected after the relation is created."""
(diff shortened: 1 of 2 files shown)


tests/unit/test_endpoint_provider.py
```diff
@@ -615,3 +615,26 @@ def test_unit_label_is_retained_if_hard_coded(self):
...
+        self.addCleanup(self.harness.cleanup)
+        self.harness.set_leader(False)
+        self.harness.begin_with_initial_hooks()
+
+    @patch("ops.testing._TestingModelBackend.network_get")
+    def test_alert_rules(self, _):
+        """Verify alert rules are added when leader is elected after the relation is created."""
+        rel_id = self.harness.add_relation(RELATION_NAME, "provider")
+        self.harness.add_relation_unit(rel_id, "provider/0")
+        self.harness.set_leader(True)
```
sed-i, line 634:
```suggestion
        # First, make sure there are no rules present
        data = self.harness.get_relation_data(rel_id, self.harness.model.app.name).get(
            "alert_rules"
        )
        self.assertIsNone(data)
        
        # Now, trigger leader elected and make sure the rules are in place
        self.harness.set_leader(True)
```

dstathis: At first I thought this was a good idea but when I thought about it more I think we shouldn't check for this. It would be checking an implementation detail instead of checking the intended behavior. In other words, It wouldn't be a problem if alert rules showed up before `leader_elected`.

sed-i: Yep, makes sense.

.github/workflows/pull-request.yaml
```diff
@@ -101,5 +101,6 @@ jobs:
         with:
           juju-channel: latest/stable
           provider: microk8s
+          channel: 1.23/stable
```
sed-i, line 104: This [might be fixed](https://github.com/charmed-kubernetes/actions-operator/pull/38) now. Wanna give 1.24 another chance?

sed-i (dismissed):

rbarry82 (approved):

Issue #263, opened 2022-04-26
Passing multiple scrape jobs to the MetricsEndpointProvider leads to a broken Prometheus config


simskij:
### Bug Description

Passing multiple scrape jobs to the MetricsEndpointProvider leads to a broken Prometheus config.

### To Reproduce

1. Alter a test charm to set up multiple scrape jobs, like

```python
        self._scraping = MetricsEndpointProvider(
            self,
            jobs=[
                {"static_configs": [{"targets": ["*:9100"]}]},
                {"static_configs": [{"targets": ["*:8080","*:8081"]}]}
            ]
        )
```

2. Build and deploy
3. Relate to Prometheus

### Environment

-

### Relevant log output

```shell
unit-prometheus-0: 13:43:28 INFO unit.prometheus/0.juju-log metrics-endpoint:18: Pushed new configuration
unit-prometheus-0: 13:43:28 ERROR unit.prometheus/0.juju-log metrics-endpoint:18: Prometheus failed to reload the configuration
unit-prometheus-0: 13:43:28 INFO juju.worker.uniter.operation ran "metrics-endpoint-relation-changed" hook (via hook dispatching script: dispatch)
unit-prometheus-0: 13:43:29 INFO unit.prometheus/0.juju-log metrics-endpoint:18: Pushed new configuration
unit-prometheus-0: 13:43:29 ERROR unit.prometheus/0.juju-log metrics-endpoint:18: Prometheus failed to reload the configuration
unit-prometheus-0: 13:43:29 INFO juju.worker.uniter.operation ran "metrics-endpoint-relation-changed" hook (via hook dispatching script: dispatch)
```

and the unit databag for prometheus:
```yaml
  - endpoint: metrics-endpoint
    related-endpoint: metrics-endpoint
    application-data:
      scrape_jobs: '[{"metrics_path": "/metrics", "static_configs": [{"targets": ["*:9100"]}]},
        {"metrics_path": "/metrics", "static_configs": [{"targets": ["*:8080", "*:8081"]}]}]'
      scrape_metadata: '{"model": "cos", "model_uuid": "eb29584f-7b1d-4b5b-8a69-d99b5a3829bf",
        "application": "kube-metrics", "unit": "kube-metrics/0", "charm_name": "kube-metrics-k8s"}'
```

### Additional context

Specifying multiple ports to scrape in the same scrape job works as intended.

mmanciop:
[...]

PR #292, opened 2022-05-25
Fix scrape jobs so that there are no duplicate job names.

dstathis:
This is because duplicate job names cause errors in the prometheus config

## Issue
https://github.com/canonical/prometheus-k8s-operator/issues/263

## Solution
Append a hash to then end of each duplicate job name

## Testing Instructions
Ensure that if duplicate job name or no job names are used, Prometheus does not crash.

## Release Notes
Fixed an issue where duplicate job names were causing Prometheus to crash.

### DIFF
--- a/lib/charms/prometheus_k8s/v0/prometheus_scrape.py
+++ b/lib/charms/prometheus_k8s/v0/prometheus_scrape.py
@@ -311,6 +311,8 @@ of Metrics provider charms hold eponymous information.
 
 """  # noqa: W505
 
+import copy
+import hashlib
 import ipaddress
 import json
 import logging
@@ -881,6 +883,8 @@ class MetricsEndpointConsumer(Object):
             if static_scrape_jobs:
                 scrape_jobs.extend(static_scrape_jobs)
 
+        scrape_jobs = _dedupe_job_names(scrape_jobs)
+
         return scrape_jobs
 
     def alerts(self) -> dict:
@@ -1233,6 +1237,38 @@ class MetricsEndpointConsumer(Object):
             static_config["targets"] = [host_address]  # type: ignore
 
         return static_config
+
+
+def _dedupe_job_names(jobs):
+    # Convert to a dict with job names as keys
+    # I think this line is O(n^2) but it should be okay given the list sizes
+    jobs_copy = copy.deepcopy(jobs)
+    jobs_dict = {job["job_name"]: list(filter(lambda x: x["job_name"] == job["job_name"], jobs_copy)) for job in jobs_copy}
+
+    # If multiple jobs have the same name, convert the name to "name_<hash-of-job>"
+    for key in jobs_dict:
+        if len(jobs_dict[key]) > 1:
+            for job in jobs_dict[key]:
+                job_json = json.dumps(job)
+                hashed = hashlib.sha256(job_json.encode()).hexdigest()
+                job["job_name"] = '{}_{}'.format(job["job_name"], hashed)
+    new_jobs = []
+    for key in jobs_dict:
+        new_jobs.extend([i for i in jobs_dict[key]])
+
+    # Deduplicate jobs which are deeply equal
+    # Again this in O(n^2) but it should be okay
+    deduped_jobs = []
+    seen = []
+    for job in new_jobs:
+        job_json = json.dumps(job)
+        hashed = hashlib.sha256(job_json.encode()).hexdigest()
+        if hashed in seen:
+            continue
+        seen.append(hashed)
+        deduped_jobs.append(job)
+
+    return deduped_jobs
 
 
 def _resolve_dir_against_charm_path(charm: CharmBase, *path_elements: str) -> str:
--- a/lib/charms/prometheus_k8s/v0/prometheus_scrape.py
+++ b/lib/charms/prometheus_k8s/v0/prometheus_scrape.py
@@ -1239,7 +1239,7 @@ class MetricsEndpointConsumer(Object):
         return static_config
 
 
-def _dedupe_job_names(jobs):
+def _dedupe_job_names(jobs: List[dict]):
     """Deduplicate a list of dicts by appending a hash to the value of the 'job_name' key.
 
     Additionally fully dedeuplicate any identical jobs.
--- a/lib/charms/prometheus_k8s/v0/prometheus_scrape.py
+++ b/lib/charms/prometheus_k8s/v0/prometheus_scrape.py
@@ -1240,9 +1240,17 @@ class MetricsEndpointConsumer(Object):
 
 
 def _dedupe_job_names(jobs):
+    """Deduplicate a list of dicts by appending a hash to the value of the 'job_name' key.
+
+    Additionally fully dedeuplicate any identical jobs.
+
+    Args:
+        jobs: A list of prometheus scrape jobs
+    """
+    jobs_copy = copy.deepcopy(jobs)
+
     # Convert to a dict with job names as keys
     # I think this line is O(n^2) but it should be okay given the list sizes
-    jobs_copy = copy.deepcopy(jobs)
     jobs_dict = {
         job["job_name"]: list(filter(lambda x: x["job_name"] == job["job_name"], jobs_copy))
         for job in jobs_copy
@@ -1259,7 +1267,7 @@ def _dedupe_job_names(jobs):
     for key in jobs_dict:
         new_jobs.extend([i for i in jobs_dict[key]])
 
-    # Deduplicate jobs which are deeply equal
+    # Deduplicate jobs which are equal
     # Again this in O(n^2) but it should be okay
     deduped_jobs = []
     seen = []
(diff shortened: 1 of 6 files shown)


rbarry82 (approved):

lib/charms/prometheus_k8s/v0/prometheus_scrape.py
```diff
@@ -1390,6 +1394,41 @@ def _labeled_unit_config(
         return static_config
 
 
+def _dedupe_job_names(jobs):
```
sed-i, line 1397:
```suggestion
def _dedupe_job_names(jobs: List[dict]):
```

tests/unit/test_functions.py
```diff
@@ -0,0 +1,61 @@
+# Copyright 2022 Canonical Ltd.
+# See LICENSE file for licensing details.
+
+import unittest
+
+import deepdiff
+from charms.prometheus_k8s.v0.prometheus_scrape import _dedupe_job_names
+
+
```
sed-i, line 9: Do we need a test that deduping is not done in-place? I.e. that the passed list is unaffected.

lib/charms/prometheus_k8s/v0/prometheus_scrape.py
```diff
@@ -1390,6 +1394,41 @@ def _labeled_unit_config(
...
+    # Convert to a dict with job names as keys
+    # I think this line is O(n^2) but it should be okay given the list sizes
+    jobs_copy = copy.deepcopy(jobs)
+    jobs_dict = {
```
sed-i, line 1401: Add comment describing the purpose of this block?
"Create a mapping between job_name values and a list of a all the jobs matching that name"?

lib/charms/prometheus_k8s/v0/prometheus_scrape.py
```diff
@@ -1390,6 +1394,41 @@ def _labeled_unit_config(
...
+        if len(jobs_dict[key]) > 1:
+            for job in jobs_dict[key]:
+                job_json = json.dumps(job)
+                hashed = hashlib.sha256(job_json.encode()).hexdigest()
+                job["job_name"] = "{}_{}".format(job["job_name"], hashed)
+    new_jobs = []
+    for key in jobs_dict:
+        new_jobs.extend([i for i in jobs_dict[key]])
+
+    # Deduplicate jobs which are deeply equal
```
sed-i, line 1417: Could you clarify what "deeply equal" means?

lib/charms/prometheus_k8s/v0/prometheus_scrape.py
```diff
@@ -1390,6 +1394,41 @@ def _labeled_unit_config(
...
+    # Again this in O(n^2) but it should be okay
```
sed-i, line 1418: Looks O(n) to me. What am I missing?

lib/charms/prometheus_k8s/v0/prometheus_scrape.py
sed-i: How about this approach:
```python
from collections import Counter

# Create a mapping between a (job_name, as_json) tuple to its multiplicity
jobs_ = Counter([(job.get("job_name", ""), json.dumps(job)) for job in jobs])

deduped = []
for (job_name, as_json), multiplicity in jobs_.itmes():
    if multiplicity == 1:
        deduped.append(...)
    else:  # multiplicity > 1
        adjusted_job_name = job_name + "_" + sha256(as_json)
        if adjusted_job_name not in ...:
           deduped.append(...)
```

tests/integration/test_deduplication.py
```diff
@@ -0,0 +1,58 @@
+#!/usr/bin/env python3
```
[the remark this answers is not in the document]
dstathis: I intentionally wanted to name it this to include similar tests when we dedupe alert rules.

tests/unit/test_functions.py
```diff
@@ -0,0 +1,61 @@
+# Copyright 2022 Canonical Ltd.
```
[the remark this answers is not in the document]
dstathis: I don't really understand why. We are testing standalone functions.

tests/unit/test_functions.py
```diff
@@ -0,0 +1,61 @@
...
+            },
+            {
+                "job_name": "job0",
+                "static_configs": [{"targets": ["localhost:9091"]}],
+                "scrape_interval": "5s",
+            },
+        ]
+        expected = [
+            {
+                "job_name": "job0_6f9f1c305506707b952aef3885fa099fe36158f6359b8a06634068270645aefd",
```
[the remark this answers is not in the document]
dstathis: It's the same right?

tests/integration/test_deduplication.py
```diff
@@ -0,0 +1,58 @@
...
+            prometheus_tester_charm,
+            resources=tester_resources,
+            application_name=tester_app_name,
+            config={"scrape_jobs": json.dumps(jobs)},
+        ),
+    )
+    await ops_test.model.add_relation(prometheus_app_name, tester_app_name)
+    await ops_test.model.wait_for_idle(status="active")
+
+    targets = await get_prometheus_active_targets(ops_test, prometheus_app_name)
```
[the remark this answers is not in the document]
dstathis: utests test a specific method. itests test that the charm behaves correctly.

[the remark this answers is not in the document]
sed-i: Granted, but what is the added value here?
Specifically, given the utest, this seems to only be testing prometheus itself.
(BTW active targets is already tested as part of another itest.)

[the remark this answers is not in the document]
rbarry82: The unit test is not actual Prometheus and can't validate the behavior of the "real" code.

The existing active targets integration test doesn't verify dupes.

lib/charms/prometheus_k8s/v0/prometheus_scrape.py
dstathis: How long does `if hashed in seen` take? :smile:

lib/charms/prometheus_k8s/v0/prometheus_scrape.py
dstathis: I think I had that but it was in the wrong place. Fixing.

lib/charms/prometheus_k8s/v0/prometheus_scrape.py
```diff
@@ -1390,6 +1394,41 @@ def _labeled_unit_config(
         return static_config
 
 
+def _dedupe_job_names(jobs):
+    # Convert to a dict with job names as keys
+    # I think this line is O(n^2) but it should be okay given the list sizes
```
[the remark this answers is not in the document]
dstathis: Added similar docstring.

lib/charms/prometheus_k8s/v0/prometheus_scrape.py
dstathis: Just means equal. Comment changed.

lib/charms/prometheus_k8s/v0/prometheus_scrape.py
dstathis: I don't think this works. We want to detect jobs which are different but have the same name.

tests/integration/prometheus-tester/config.yaml
```diff
@@ -10,3 +10,7 @@ options:
     default: src/prometheus_alert_rules
     description: "Path for alert rules passed to the Provider"
     type: string
+  scrape_jobs:
+    default: "[]"
+    description: "Scrape jobs to pass to the MetricsEndpointProvider constructor"
```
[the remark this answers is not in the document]
dstathis: Pretty sure it's `Provider`

tests/unit/test_functions.py
```diff
@@ -0,0 +1,61 @@
...
+
+import unittest
+
+import deepdiff
+from charms.prometheus_k8s.v0.prometheus_scrape import _dedupe_job_names
+
+
+class TestFunctions(unittest.TestCase):
+    def test_dedupe_job_names(self):
+        jobs = [
```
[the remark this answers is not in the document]
dstathis: This is probably overkill.

tests/unit/test_functions.py
dstathis: Done.

tests/integration/test_deduplication.py
```diff
@@ -0,0 +1,58 @@
...
+# See LICENSE file for licensing details.
+
+import asyncio
+import json
+
+from helpers import get_prometheus_active_targets, oci_image
+from pytest_operator.plugin import OpsTest
+
+prometheus_app_name = "prometheus"
+prometheus_resources = {"prometheus-image": "prom/prometheus:v2.35.0"}
```
sed-i, line 12: Should probably use `upstream-source` here.

tests/unit/test_functions.py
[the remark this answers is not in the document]
sed-i: Changing or duplicating a test case is more involved when this is hard-coded.

sed-i (approved):

tests/integration/test_deduplication.py
dstathis: Done.

PR #295, opened 2022-05-28
Add maximum retention size config option (in percent)

sed-i:
## Issue
We currently only set maximum retention time for our timeseries data, meaning we’re likely to run out of disk at some point. COS Lite, being treated as a lightweight appliance with resource constraints, should enforce a reasonable default maximum retention storage size as well. It should default to 80% of the PVC size.

Crossref: [OPENG-231](https://warthogs.atlassian.net/browse/OPENG-231)

## Solution
Add maximum retention size config option.

## Context
NTA.

## Testing Instructions
Run all tests.

## Release Notes
Add maximum retention size config option.

### DIFF
--- a/src/charm.py
+++ b/src/charm.py
@@ -59,7 +59,7 @@ class PrometheusCharm(CharmBase):
         self.metrics_consumer = MetricsEndpointConsumer(self)
 
         # Manages ingress for this charm
-        self.ingress = IngressPerUnitRequirer(self, endpoint="ingress", port=self._port)
+        self.ingress = IngressPerUnitRequirer(self, relation_name="ingress", port=self._port)
 
         external_url = urlparse(self._external_url)
 
--- a/src/charm.py
+++ b/src/charm.py
@@ -303,18 +303,17 @@ class PrometheusCharm(CharmBase):
                 pvc_name = volume.persistentVolumeClaim.claimName
                 break
 
-        if pvc_name:
-            capacity = cast(
-                PersistentVolumeClaim,
-                client.get(PersistentVolumeClaim, name=pvc_name, namespace=self.model.name),
-            ).status.capacity[
-                "storage"  # Storage name must match metadata.yaml
-            ]
-
-            return capacity
-
-        else:
+        if not pvc_name:
             raise ValueError("No PVC found for pod " + pod_name)
+
+        capacity = cast(
+            PersistentVolumeClaim,
+            client.get(PersistentVolumeClaim, name=pvc_name, namespace=self.model.name),
+        ).status.capacity[
+            "storage"  # Storage name must match metadata.yaml
+        ]
+
+        return capacity
 
     def _is_valid_timespec(self, timeval: str) -> bool:
         """Is a time interval unit and value valid.
--- a/src/charm.py
+++ b/src/charm.py
@@ -290,6 +290,7 @@ class PrometheusCharm(CharmBase):
         This may need to be handled differently once Juju supports multiple storage instances
         for k8s (https://bugs.launchpad.net/juju/+bug/1977775).
         """
+        # Assuming the storage name is "databases" (must match metadata.yaml).
         # This assertion would be picked up by every integration test so no concern this would
         # reach production.
         assert (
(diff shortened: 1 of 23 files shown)


src/charm.py
```diff
@@ -236,16 +240,83 @@ def _command(self) -> str:
...
+        if not capacity.endswith("B"):
+            capacity += "B"
+
+        # For simplicity, always communicate to prometheus in GiB
+        storage_value = multiplier * bitmath.parse_string(capacity).to_GiB().value
+        return f"{storage_value}GB"
+
+    def _get_pvc_capacity(self) -> str:
+        # Get PVC capacity from kubernetes
+        client = Client()
```
Abuelodelanada, line 289: Maybe it's a matter of taste, but I would inject `Client` dependency instead of asking `_get_pvc_capacity` to instantiate it. It will be also easier to test.

src/charm.py
```diff
@@ -236,16 +240,83 @@ def _command(self) -> str:
...
+                PersistentVolumeClaim,
+                client.get(PersistentVolumeClaim, name=pvc_name, namespace=self.model.name),
+            ).status.capacity[
+                "storage"  # Storage name must match metadata.yaml
+            ]
+
+            return capacity
+
+        else:
+            raise ValueError("No PVC found for pod " + pod_name)
```
Abuelodelanada, line 317: I would handle special cases first:

```python
        if not pvc_name:
            raise ValueError("No PVC found for pod " + pod_name)

        capacity = cast(
            PersistentVolumeClaim,
            client.get(PersistentVolumeClaim, name=pvc_name, namespace=self.model.name),
        ).status.capacity[
            "storage"  # Storage name must match metadata.yaml
        ]

        return capacity            
```

Abuelodelanada (approved):

src/charm.py
sed-i: (Discussed and agreed that leaking implementation is also not ideal.)

src/charm.py
```diff
@@ -236,16 +240,82 @@ def _command(self) -> str:
...
+        #              'persistentVolumeClaim': {'claimName': 'am-data-d7f6a623-am-0'}}, ...]
+        pvc_name = ""
+        for volume in cast(
+            Pod, client.get(Pod, name=pod_name, namespace=self.model.name)
+        ).spec.volumes:
+            # TODO: How to handle this when the store is not a singleton (i.e. when metadata
+            #  specifies "multiple: range: 1-10")?
+            if volume.persistentVolumeClaim.claimName.startswith(f"{self.app.name}-database-"):
+                pvc_name = volume.persistentVolumeClaim.claimName
+                break
```
sed-i, line 304: @rbarry82 what would be the best way here to obtain the PVC?

Is multiple stores possible? I added the following to prometheus-k8s metadata:
```diff
 storage:
   database:
     type: filesystem
+    multiple:
+      range: 3-5
```
And when I deploy I see

`ERROR juju.worker.caasapplicationprovisioner.runner exited "prometheus-k8s": ensuring application "prometheus-k8s": duplicated storage name "database" for "prometheus-k8s" not valid`

src/charm.py
```diff
@@ -236,16 +240,93 @@ def _command(self) -> str:
...
+        # Take the first volume whose name starts with "<app-name>-database-".
+        # The volumes array looks as follows for app "am" and storage "data":
+        # 'volumes': [{'name': 'am-data-d7f6a623',
+        #              'persistentVolumeClaim': {'claimName': 'am-data-d7f6a623-am-0'}}, ...]
+        pvc_name = ""
+        for volume in cast(
+            Pod, client.get(Pod, name=pod_name, namespace=self.model.name)
+        ).spec.volumes:
+            # claimName looks like this: 'prom-database-325a0ee8-prom-0'
+            matcher = re.compile(f"^{self.app.name}-database-.*?-{pod_name}$")
```
rbarry82, line 312: This should be `fr'...` if we're using f-strings. 

Raw strings are best practice with regexps to avoid unwanted escapes.

src/charm.py
```diff
@@ -236,16 +240,93 @@ def _command(self) -> str:
...
+                break
+
+        if not pvc_name:
+            raise ValueError("No PVC found for pod " + pod_name)
+
+        capacity = cast(
+            PersistentVolumeClaim,
+            client.get(PersistentVolumeClaim, name=pvc_name, namespace=self.model.name),
+        ).status.capacity[
+            "storage"  # Storage name must match metadata.yaml
```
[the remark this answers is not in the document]
sed-i: This comment is incorrectly placed. Fixed in next commit.

Cannot look it up dynamically, unless assuming there is only one:
```yaml
storage:
  database:
    type: filesystem
```

simskij:
Would it be possible to support both fixed sizes and percentages? If so, I think that would be killer.

src/charm.py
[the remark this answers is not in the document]
rbarry82: If it's actually the parent (`storage`), then the key is fixed anyway.

Otherwise, if it "must match" and it's an actual key such as `database:`, it's in `charm.meta.storages`

`storage` seems concrete enough. At first glance, it looked like a placeholder

rbarry82 (approved):

PR #297, opened 2022-06-08
Use `cos-tool` for transformation and validation, validate alert rules

### LINKED ISSUE #170: Validate alert rules
Before alert rules are created in the prometheus workload container it would probably be handy if invalid rules were flagged.
Validating alert rules has several aspects:
- schema (easy)
- promql syntax (more involved)

> Promtool already provides validation of alert rules. I am not sure if we should reimplement this functionality using pydantic. Besides schema validation for relation data is a planned feature in OF.

_Originally posted by @balbirthomas in https://github.com/canonical/prometheus-operator/pull/161#r754135835_

> Is there a Github ticket or product spec to discuss the requirements around validation of alert rules ? Please note Prometheus provides [a tool](https://www.prometheus.io/docs/prometheus/latest/configuration/unit_testing_rules/) to test alert rules.

_Originally posted by @balbirthomas in https://github.com/canonical/prometheus-operator/pull/161#r754228001_

### LINKED ISSUE #203: `scrape_identifier` and `promql_labels` use `charm` and `charm_name` inconsistently
`JujuTopology` sometimes uses `charm_name` (`as_promql_label_dict()`, `identifier()`, etc) and other places uses `charm`.

This leads to far too much complexity in the code to maintain a consistent, non-breaking API, as each method must determine whether or not to substitute `charm_name -> charm`.

We should use the same identifier everywhere, so these can be coalesced, and the `rename_keys` logic can be stripped out entirely

rbarry82:
## Issue
Kind of a lot.

A mechanism was needed to validate alert rules _before_ Prometheus died if they were loaded, and `JujuTopology` is now in `observability-libs`, so removal of the custom code was slated.

Closes #170 
Closes #203

## Solution
Supplant `promql-transform` with `cos-tool`, which does what `promql-transform` does plus also handling logql transformation and both types of alert rules. Add tests for all of that, and an event fired back if an alert rule is invalid, which is then excluded from being loaded into Prometheus

Replace the "custom" `JujuTopology` with the one from `observability-libs`, which works well, but required a really annoying amount of updates to tests, especially unit tests to avoid `InvalidUUID` exceptions blowing them up.

## Testing Instructions
Test, run, be joyful.

## Release Notes
Use `JujuTopology` from observability-libs, validate alert rules


lib/charms/prometheus_k8s/v0/prometheus_remote_write.py
```diff
@@ -1108,18 +924,51 @@ def apply_label_matchers(self, rules):
...
+
+            # Smash "our" rules format into what upstream actually uses, which is more like:
+            #
+            # groups:
+            #   - name: foo
+            #     rules:
+            #       - alert: BlahBlah
+            #         expr: up
+            #       - alert: Haha
+            #         expr: up
```
sed-i, line 947:
```suggestion
            #       - alert: Alert1
            #         expr: up
            #       - alert: Alert2
            #         expr: up
```

lib/charms/prometheus_k8s/v0/prometheus_remote_write.py
```diff
@@ -1108,18 +924,51 @@ def apply_label_matchers(self, rules):
...
+            transformed_rules = {"groups": []}  # type: ignore
+            for rule in rules["groups"]:
+                transformed = {"name": str(uuid.uuid4()), "rules": [rule]}
+                transformed_rules["groups"].append(transformed)
+
+            rule_path.write_text(yaml.dump(rules))
+
+            args = [str(self.path), "v", str(rule_path)]
```
sed-i, line 955: Comment what "v" means?

lib/charms/prometheus_k8s/v0/prometheus_remote_write.py
```diff
@@ -1108,18 +924,51 @@ def apply_label_matchers(self, rules):
...
+            #       - alert: BlahBlah
+            #         expr: up
+            #       - alert: Haha
+            #         expr: up
+            transformed_rules = {"groups": []}  # type: ignore
+            for rule in rules["groups"]:
+                transformed = {"name": str(uuid.uuid4()), "rules": [rule]}
+                transformed_rules["groups"].append(transformed)
+
+            rule_path.write_text(yaml.dump(rules))
```
sed-i, line 953: Is `tempfile.TemporaryDirectory()` guaranteed to succeed in the k8s context?
Is there an advantage in communicating this over stdin?

lib/charms/prometheus_k8s/v0/prometheus_remote_write.py
```diff
@@ -1108,18 +924,51 @@ def apply_label_matchers(self, rules):
...
+
+            args = [str(self.path), "v", str(rule_path)]
+            # noinspection PyBroadException
+            try:
```
sed-i, line 957: Should the try-except be absorbed by `self._exec`?
```python
success, stdout = self._exec(args)
return success, stdout
```

lib/charms/prometheus_k8s/v0/prometheus_remote_write.py
```diff
@@ -1108,18 +924,51 @@ def apply_label_matchers(self, rules):
...
+
+    def inject_label_matchers(self, expression, topology):
+        """Add label matchers to an expression."""
         if not topology:
             return expression
         if not self.path:
-            logger.debug(
-                "`promql-transform` unavailable. leaving expression unchanged: %s", expression
-            )
+            logger.debug("`cos-tool` unavailable. leaving expression unchanged: %s", expression)
```
sed-i, line 969:
```suggestion
            logger.debug("`cos-tool` unavailable. Leaving expression unchanged: %s", expression)
```

lib/charms/prometheus_k8s/v0/prometheus_remote_write.py
```diff
@@ -1108,18 +924,51 @@ def apply_label_matchers(self, rules):
...
             return expression
-        args = [str(self.path)]
+        args = [str(self.path), "t"]
```
sed-i, line 971: Comment what does "t" do?
Or perhaps use long-form args so this is less cryptic?

.gitignore
```diff
@@ -11,3 +11,4 @@ venv
 tests/integration/prometheus-tester/lib/charms/prometheus_k8s/v0/prometheus_scrape.py
 promql-transform
 promql-transform*
+cos-tool*
```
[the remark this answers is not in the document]
rbarry82: We had this ignore ever since `promql-transform` was introduced. 

Moving the download would require a bunch of hacking around `class CosTool` and the path lookup, which is a bad tradeoff.

lib/charms/prometheus_k8s/v0/prometheus_remote_write.py
```diff
@@ -416,12 +219,17 @@ def _from_file(self, root_path: Path, file_path: Path) -> List[dict]:
...
                         # only insert labels that do not already exist
-                        for label, val in self.topology.as_promql_label_dict().items():
+                        for label, val in self.topology.label_matcher_dict.items():
                             if label not in alert_rule["labels"]:
                                 alert_rule["labels"][label] = val
                         # insert juju topology filters into a prometheus alert rule
-                        alert_rule["expr"] = self.topology.render(alert_rule["expr"])
+                        # insert juju topology filters into a prometheus alert rule
+                        alert_rule["expr"] = self.tool.inject_label_matchers(
+                            re.sub(r"%%juju_topology%%,?", "", alert_rule["expr"]),
```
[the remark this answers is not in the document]
rbarry82: I'm reasonably sure we already documented that when `promql-transform` came in. The `STUB` isn't even in `observability_libs.juju_topology`. The "missing piece" was in the alert rules, and nobody is using it other than us.

Once we strip it out of all of our test data, this regexp can go away.

lib/charms/prometheus_k8s/v0/prometheus_remote_write.py
rbarry82: Yes, pods can still make tempdirs.

The advantage is that the rule parsing code in Go, from upstream, expects to find a file. We _could_ pass stdin and read in `[]byte` in Go, but that means that it would not be possible to validate multiple files in a single command if we ever need to, and this is a use case supported by `cortex-tool` and `mimir-tool`, so matching behavior there makes sense.

lib/charms/prometheus_k8s/v0/prometheus_remote_write.py
rbarry82: `stdout` and `stderr` (in `e.output`) are not equivalent and shouldn't be treated that way.

lib/charms/prometheus_k8s/v0/prometheus_remote_write.py
```diff
@@ -1063,32 +872,39 @@ def alerts(self) -> dict:
...
-class PromqlTransformer:
-    """Uses promql-transform to inject label matchers into alert rule expressions."""
+class CosTool:
+    """Uses cos-tool to inject label matchers into alert rule expressions and validate rules."""
 
     _path = None
     _disabled = False
 
+    def __init__(self, charm):
+        self._charm = charm
```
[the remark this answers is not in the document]
rbarry82: Honestly, `self._charm` isn't used anywhere, it was left from `PromqlTransform`, and `__init__` got moved to the top of the class definition.

`__init__` probably doesn't need any args at all. The more-or-less `getattr` lookup in `CosTool.path` serves the same purpose as the check you're suggesting.

tests/unit/test_endpoint_aggregator.py
```diff
@@ -77,10 +78,11 @@ def __init__(self, *args, **kwargs):
         )
 
 
+@patch("charms.observability_libs.v0.juju_topology.JujuTopology.is_valid_uuid", lambda *args: True)
 class TestEndpointAggregator(unittest.TestCase):
     def setUp(self):
         self.harness = Harness(EndpointAggregatorCharm, meta=AGGREGATOR_META)
-        self.harness.set_model_info(name="testmodel", uuid="1234567890")
+        self.harness.set_model_info(name="testmodel", uuid="1234567")
```
sed-i, line 85: Does this mean the class-level patch does not work?

tests/unit/test_endpoint_provider.py
```diff
@@ -87,13 +87,21 @@ def __init__(self, *args, **kwargs):
         )
 
 
+@patch("charms.observability_libs.v0.juju_topology.JujuTopology.is_valid_uuid", lambda *args: True)
```
sed-i, line 90: Is the class level patch needed in addition to the method level patches?

tests/unit/test_transform.py
```diff
@@ -141,3 +143,58 @@ def test_handles_multiple_labels(self):
...
+class TestValidateAlerts(unittest.TestCase):
+    """Test that the cos-tool validation works."""
+
+    def setUp(self):
+        self.harness = Harness(ToolProviderCharm)
+        self.harness.set_model_name("validate")
+        self.addCleanup(self.harness.cleanup)
+        self.harness.begin()
+
+    @unittest.mock.patch("platform.processor", lambda: "x86_64")
```
sed-i, line 158: I'm curious why this needs patching?

tests/unit/test_endpoint_provider.py
rbarry82: Yes, in all cases. Patches to `setUp` do not apply to methods if it's instantiated later, and class-level patches do not apply to `setUp`

tests/unit/test_transform.py
rbarry82: In case they ever run on a different architecture, probably. This already existed also.

lib/charms/prometheus_k8s/v0/prometheus_scrape.py
```diff
@@ -1594,6 +1450,22 @@ def __init__(
         # If there is no leader during relation_joined we will still need to set alert rules.
         self.framework.observe(self._charm.on.leader_elected, self._set_scrape_job_spec)
 
+    def _on_relation_changed(self, event):
+        """Check for alert rule messages in the relation data before moving on."""
+        if self._charm.unit.is_leader():
+            ev = json.loads(event.relation.data[event.app].get("event", "{}"))
```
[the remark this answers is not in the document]
rbarry82: It would require adding `cos-tool` to the providers.

lib/charms/prometheus_k8s/v0/prometheus_scrape.py
```diff
@@ -415,6 +418,36 @@ def __init__(
...
+        }
+
+    def restore(self, snapshot):
+        """Restore alert rule information."""
+        self.valid = snapshot["valid"]
+        self.errors = snapshot["errors"]
+
+
+class MetricsEndpointProviderEvents(ObjectEvents):
+    """Events raised by :class:`InvalidAlertRuleEvent`s."""
```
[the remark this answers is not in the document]
rbarry82: This is not accurate. They are raised by `InvalidAlertRuleEvent`.

lib/charms/prometheus_k8s/v0/prometheus_remote_write.py
rbarry82: That's exactly the same ;)

Success here is implicitly "didn't throw `CalledProcessError`". 

If the return code is 0, there's no output at all. If the return code is 1, an exception must be caught and stderr (from `e`) must be read.

It's not the same semantics as the return values from `transform`, so `_exec` can't do both without coupling it. Catching the exception is fine.

lib/charms/prometheus_k8s/v0/prometheus_remote_write.py
```diff
@@ -943,7 +751,7 @@ def __init__(
 
         super().__init__(charm, relation_name)
         self._charm = charm
-        self._transformer = PromqlTransformer(self._charm)
+        self.tool = CosTool(self._charm)
```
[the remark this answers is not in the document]
rbarry82: Yes. There's not actual need to make it a private attribute, and making it public allows it to be reached from the object in the charm if we ever need/want to for other operations.

lib/charms/prometheus_k8s/v0/prometheus_remote_write.py
```diff
@@ -1063,32 +872,39 @@ def alerts(self) -> dict:
                 except KeyError:
                     logger.error("Alert rules were found but no usable labels were present")
 
+            _, errmsg = self.tool.validate_alert_rules(alert_rules)
+            if errmsg:
+                if alerts[identifier]:
+                    del alerts[identifier]
+                relation.data[self._charm.app]["event"] = json.dumps({"errors": errmsg})
```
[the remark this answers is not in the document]
rbarry82: * No
* No
* No

This is a "battle-tested" pattern, and `juju show-application` will also show `valid: true` instead of `valid: false`, so there's no confusion about it.

lib/charms/prometheus_k8s/v0/prometheus_remote_write.py
```diff
@@ -1063,32 +872,39 @@ def alerts(self) -> dict:
...
+                continue
+
         return alerts
 
 
 # Copy/pasted from prometheus_scrape.py
-class PromqlTransformer:
-    """Uses promql-transform to inject label matchers into alert rule expressions."""
+class CosTool:
+    """Uses cos-tool to inject label matchers into alert rule expressions and validate rules."""
```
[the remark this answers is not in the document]
rbarry82: No. It's utility wrappers around the same binary.

`_charm` doesn't seem to be used, but since `PromqlTransform` used it, I left it until we find out whether there was a reason for htis.

lib/charms/prometheus_k8s/v0/prometheus_remote_write.py
```diff
@@ -1108,18 +924,51 @@ def apply_label_matchers(self, rules):
                     if label in rule["labels"]:
                         topology[label] = rule["labels"][label]
 
-                rule["expr"] = self._apply_label_matcher(rule["expr"], topology)
+                rule["expr"] = self.inject_label_matchers(rule["expr"], topology)
         return rules
 
-    def _apply_label_matcher(self, expression, topology):
+    def validate_alert_rules(self, rules: dict) -> Tuple[bool, str]:
```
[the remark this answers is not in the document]
rbarry82: See above about the possibility of using this externally. "This string is empty" doesn't necessarily mean "the returncode was 0".

lib/charms/prometheus_k8s/v0/prometheus_remote_write.py
```diff
@@ -1108,18 +924,51 @@ def apply_label_matchers(self, rules):
...
+        """Will validate correctness alert rules, returning a boolean and any errors."""
+        if not self.path:
+            logger.debug("`cos-tool` unavailable. Not validating alert correctness.")
```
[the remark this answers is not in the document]
rbarry82: It was `debug` in `promql-transform` also. We don't need `WARNING` lines every time `update-status` fires if the omission was intentional.

lib/charms/prometheus_k8s/v0/prometheus_remote_write.py
```diff
@@ -1108,18 +924,51 @@ def apply_label_matchers(self, rules):
...
+            #   - name: foo
+            #     rules:
+            #       - alert: BlahBlah
+            #         expr: up
+            #       - alert: Haha
+            #         expr: up
+            transformed_rules = {"groups": []}  # type: ignore
+            for rule in rules["groups"]:
+                transformed = {"name": str(uuid.uuid4()), "rules": [rule]}
+                transformed_rules["groups"].append(transformed)
```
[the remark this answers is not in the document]
rbarry82: `name` doesn't exist at all in the existing object. It's not replacing, it's _adding_ so it matches the expected format and can be validated.  Manipulating a `dict` inside a `list` inside a `dict` in one line is enough golfing to make it difficult to read.

lib/charms/prometheus_k8s/v0/prometheus_scrape.py
```diff
@@ -2139,11 +2011,14 @@ def _label_alert_rules(self, unit_rules, appname) -> list:
         labeled_rules = []
         for unit_name, rules in unit_rules.items():
             for rule in rules:
-                rule["labels"].update(
-                    AggregatorTopology.create(
-                        self.model.name, self.model.uuid, appname, unit_name
-                    ).as_promql_label_dict()
-                )
+                # the new JujuTopology removed this, so build it up by hand
```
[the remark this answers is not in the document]
sed-i: The original comment will become stale as soon as this merges.

tests/unit/test_endpoint_aggregator.py
sed-i: Mocking is still needed if we generate a proper uuid?

tests/unit/test_transform.py
sed-i: I'm assuming utests run on the same machine as itests, so still unclear to me why this needs patching.

sed-i (approved):

lib/charms/prometheus_k8s/v0/prometheus_scrape.py
```diff
@@ -415,6 +418,36 @@ def __init__(
         super().__init__(self.message)
 
 
+class InvalidAlertRuleEvent(EventBase):
```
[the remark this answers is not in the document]
rbarry82: Done

Abuelodelanada (approved):

PR #299, opened 2022-06-08
Add itests to check for data persistance

Abuelodelanada:
## Issue

This PR add itests to:

* Check if prometheus data persists across unit restarts (pod churns / `kubectl delete pod`)
* Check if prometheus data persists across charm upgrade (`juju refresh`)

## Testing Instructions

Run itests

### DIFF
--- a/tests/integration/test_remote_write_grafana_agent.py
+++ b/tests/integration/test_remote_write_grafana_agent.py
@@ -9,6 +9,7 @@ import pytest
 from helpers import check_prometheus_is_ready, oci_image, run_promql
 
 logger = logging.getLogger(__name__)
+prometheus_resources = {"prometheus-image": oci_image("./metadata.yaml", "prometheus-image")}
 
 
 @pytest.mark.abort_on_fail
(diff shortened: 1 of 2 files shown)


tests/integration/test_remote_write_grafana_agent.py
```diff
@@ -47,6 +48,60 @@ async def test_remote_write_with_grafana_agent(ops_test, prometheus_charm):
     )
 
 
+@pytest.mark.abort_on_fail
+async def test_check_data_persist_on_kubectl_delete_pod(ops_test, prometheus_charm):
```
sed-i, line 52: Is this in addition to or instead of `test_kubectl_delete.py`?

tests/integration/test_remote_write_grafana_agent.py
```diff
@@ -47,6 +48,60 @@ async def test_remote_write_with_grafana_agent(ops_test, prometheus_charm):
...
+    prometheus_app_name = "prometheus"
+    pod_name = f"{prometheus_app_name}-0"
+    query = "prometheus_tsdb_head_chunks_created_total{}"
+    total0 = await run_promql(ops_test, query, prometheus_app_name)
+    sum0 = int(total0[0]["value"][1])
+
+    cmd = [
```
sed-i, line 59: This is also used in `test_kubectl_delete.py`. Shall we add a helper?

tests/integration/test_remote_write_grafana_agent.py
```diff
@@ -47,6 +48,60 @@ async def test_remote_write_with_grafana_agent(ops_test, prometheus_charm):
     )
 
 
+@pytest.mark.abort_on_fail
+async def test_check_data_persist_on_kubectl_delete_pod(ops_test, prometheus_charm):
+    prometheus_app_name = "prometheus"
+    pod_name = f"{prometheus_app_name}-0"
+    query = "prometheus_tsdb_head_chunks_created_total{}"
+    total0 = await run_promql(ops_test, query, prometheus_app_name)
+    sum0 = int(total0[0]["value"][1])
```
sed-i, line 57: Add comment about why [0] and [1]?

```suggestion
    num_head_chunks_before = int(total0[0]["value"][1])
```

tests/integration/test_remote_write_grafana_agent.py
```diff
@@ -47,6 +48,60 @@ async def test_remote_write_with_grafana_agent(ops_test, prometheus_charm):
...
+    logger.debug(stdout)
+    await ops_test.model.block_until(
+        lambda: len(ops_test.model.applications[prometheus_app_name].units) > 0
+    )
+    await ops_test.model.wait_for_idle(apps=[prometheus_app_name], status="active", timeout=60)
+    assert check_prometheus_is_ready(ops_test, prometheus_app_name, 0)
+
+    total1 = await run_promql(ops_test, query, prometheus_app_name)
+    sum1 = int(total1[0]["value"][1])
+    assert sum0 <= sum1
```
sed-i, line 81: - Any chance we could change `<=` to `<` or alternatively at the top `assert sum0 > 0`?
- Should we also compare filesystem contents e.g. `kubectl exec du`?

tests/integration/test_remote_write_grafana_agent.py
```diff
@@ -47,6 +48,60 @@ async def test_remote_write_with_grafana_agent(ops_test, prometheus_charm):
...
+
+
+@pytest.mark.abort_on_fail
+async def test_check_data_persist_on_charm_upgrade(ops_test, prometheus_charm):
```
sed-i, line 85: Should this be moved into `test_upgrade_charm.py`?

sed-i (commented):
Fetching `ingress_per_unit` would require some work. I think we can skip fetching this one until fqdn replaces bind_address in traefik.

I did a bunch of itest improvements for #295 , you may want to branch off of there.

tests/integration/test_remote_write_grafana_agent.py
Abuelodelanada: This test may completely replace `test_kubectl_delete.py`, since do the same and also check for data persistance

tests/integration/test_remote_write_grafana_agent.py
Abuelodelanada: As far as I understand there is no warranty that the value after (sum1) deleting the pod is going to be greater than the value before (sum0) deleting the pod.... in fact since this operation is quick, the value most of the times I run the test with `pdb.set_trace()` is the same, which confirms that there is data persistence.

tests/integration/test_remote_write_grafana_agent.py
Abuelodelanada: That make sense!

tests/integration/test_remote_write_grafana_agent.py
sed-i: Is there any chance both values would be zero, giving a false negative?

Abuelodelanada: I ran a lot of manual tests and never got a zero.... seems unlikely

rbarry82 (approved):

PR #310, opened 2022-06-10
Add remote write tests with zinc through grafana_agent

rbarry82:
## Issue
Remote write with grafana agent already has a test, but it also spends a lot of its time deleting/adding pods to check for data persistence. Cross-model relations also have a test, but it doesn't really exercise client applications.

## Solution
Add two new tests (and remove the existing cross-model test, since this is a superset) which test `prometheus <-> grafana-agent <-> zinc` both in the same model and as a cross-model relation.

## Context
`zinc-k8s` is published on charmhub, but the published version is only in `stable` and doesn't actually contain the needed relation endpoints.

It was brought more or less up-to-date for operator day and never published. Fetch it as part of the tests in `tox` and build it when the tests are run.

## Testing Instructions
Run the integration tests

## Release Notes
Add remote write tests with zinc through grafana_agent


jnsgruk:
I can probably publish a later version of Zinc - just need to check it's state with @simskij

tests/integration/test_remote_write_with_zinc.py
```diff
@@ -13,10 +13,56 @@
...
+    await ops_test.model.wait_for_idle(apps=apps, status="active", idle_period=90)
+
+    assert await has_metric(
+        ops_test,
+        f'up{{juju_model="{ops_test.model_name}",juju_application="{agent_name}"}}',
+        prometheus_name,
+    )
+
+    await ops_test.model.reset()
+    await ops_test.model.wait_for_idle(idle_period=90)
```
sed-i, line 61: - Is reset much faster than a separate test file?
- Can we consolidate this function and the other `test_remote_write_*`?

sed-i (approved):

sed-i (approved):

jnsgruk:
@rbarry82 I've updated the `zinc-k8s` charm on the `edge` channel so it now supports the relations, so you should be able to simplify these tests a little :)

Abuelodelanada (approved):

tox.ini
```diff
@@ -103,8 +106,10 @@ deps =
     prometheus-api-client
     tenacity
 commands =
+    git clone --single-branch --depth=1 https://github.com/jnsgruk/zinc-k8s-operator.git {[testenv:integration]zinc_dir}
```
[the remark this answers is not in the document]
sed-i: Can this be removed now?

tox.ini
```diff
@@ -103,8 +106,10 @@ deps =
...
     pytest -vv --tb native --log-cli-level=INFO --color=yes -s {posargs} {toxinidir}/tests/integration
 allowlist_externals =
+    git
```
sed-i, line 112: Can this be removed now?

PR #313, opened 2022-06-15
Add an action to check the prometheus config

### LINKED ISSUE #180: Juju action for debugging generated Prometheus configuration
## Use Case

A system administrator would like to understand why the Prometheus charm has entered Blocked status due to a bad Prometheus configuration file. For this purpose he would like a Juju action that runs `promtool check config /etc/prometheus/prometheus.yml` and shows him the output.

rbarry82:
## Issue
Add an action to run `promtool check config` to tell users what may be wrong with their Prometheus configuration.

Closes #180 

## Solution
Add an action to call `promtool`. This also requires checking `pebble.ExecError` if the returncode is not 0 (in case of failure) to populate the action results.

## Testing Instructions
Run the tests. Or create a bad config and test the action

## Release Notes
Add an action to check the prometheus config

### DIFF
--- a/src/charm.py
+++ b/src/charm.py
@@ -27,10 +27,10 @@ from charms.traefik_k8s.v0.ingress_per_unit import IngressPerUnitRequirer
 from lightkube import Client
 from lightkube.core.exceptions import ApiError as LightkubeApiError
 from lightkube.resources.core_v1 import PersistentVolumeClaim, Pod
-from ops.charm import CharmBase
+from ops.charm import ActionEvent, CharmBase
 from ops.main import main
 from ops.model import ActiveStatus, BlockedStatus, WaitingStatus
-from ops.pebble import ChangeError, Layer
+from ops.pebble import ChangeError, ExecError, Layer
 
 from prometheus_server import Prometheus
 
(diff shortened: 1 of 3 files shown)


dstathis (approved):

simskij (approved):

src/charm.py
```diff
@@ -274,6 +277,27 @@ def _command(self) -> str:
 
         return " ".join(command)
 
+    def _on_validate_configuration(self, event: ActionEvent) -> None:
```
sed-i, line 280: Should we be using this during layer construction?

rbarry82: It's not actually possible at least during initial layer construction (the container will not be running). When a new config is generated, we _could_, but the question then becomes what we would do with it.

Prometheus already goes into `BlockedStatus` if the service cannot start, which gives admins a chance to run this action.

If anything, validation in this sense should extract the relevant portions from `promtool` into `cos-tool` and validate rules/configs per-relation in a manner similar to alerts. I opened an issue to track this [here](https://github.com/canonical/prometheus-k8s-operator/issues/314).

PR #315, opened 2022-06-16
fix juju topology template substitution

simskij:
The current code incorrectly replaces the template with a bare comma instead of actually injecting the topology

### DIFF
--- a/lib/charms/prometheus_k8s/v0/prometheus_scrape.py
+++ b/lib/charms/prometheus_k8s/v0/prometheus_scrape.py
@@ -666,7 +666,7 @@ class AlertRules:
                         alert_rule["labels"].update(self.topology.label_matcher_dict)
                         # insert juju topology filters into a prometheus alert rule
                         alert_rule["expr"] = self.tool.inject_label_matchers(
-                            re.sub(r"%%juju_topology%%(,?)", r"\1", alert_rule["expr"]),
+                            re.sub(r"%%juju_topology%%,?", "", alert_rule["expr"]),
                             self.topology.label_matcher_dict,
                         )
 


rbarry82:
Do you have an example? 

This puts a comma back only if one was already present. The previous `template.replace("%%juju_topology%%")` also would have left it in place, and we were matching the behavior there.

simskij:
> Do you have an example?
> 
> This puts a comma back only if one was already present. The previous `template.replace("%%juju_topology%%")` also would have left it in place, and we were matching the behavior there.

And that was most likely wrong as well. 😅 A typical example would be something along the lines of `up{%%juju_topology%%, juju_unit="some-unit"}`.

This is however doubly wrong, cause the current regex matches `%%juju_topology%%,`, with `,` as a capture group, and then only reinserts the `,`.

If you check the remote write lib, you'll see that the regular expressions have diverged, and the one in remote write works as expected while the one in scrape, does not. (my patch is copied from there)

rbarry82:
> And that was most likely wrong as well. 😅 A typical example would be something along the lines of `up{%%juju_topology%%, juju_unit="some-unit"}`.
> 
> This is however doubly wrong, cause the current regex matches `%%juju_topology%%,`, with `,` as a capture group, and then only reinserts the `,`.

> If you check the remote write lib, you'll see that the regular expressions have diverged, and the one in remote write works as expected while the one in scrape, does not. (my patch is copied from there)

They actually diverged because I forgot to update remote_write. The one in scrape matches the regexp which is currently present in the Loki PR. I suppose the difference is that Loki's substitution requires adding `job=".+"` (or another "wildcard" matcher) or the LogQL expression will fail to parse at all, and keeping the comma in that case is important, while the PromQL modules from upstream don't care.

rbarry82 (approved):

dstathis (approved):

PR #316, opened 2022-06-16
replace processor call with machine call

simskij:
the current call to `platform.processor` is not reliable and quite frequently returns `"unknown"`, for instance in the CI, or on arch linux. this patch replaces all occurrences with calls to `platform.machine` which should be way more reliable.

### DIFF
--- a/lib/charms/prometheus_k8s/v0/prometheus_remote_write.py
+++ b/lib/charms/prometheus_k8s/v0/prometheus_remote_write.py
@@ -1022,7 +1022,7 @@ class CosTool:
             return expression
 
     def _get_tool_path(self) -> Optional[Path]:
-        arch = platform.processor()
+        arch = platform.machine()
         arch = "amd64" if arch == "x86_64" else arch
         res = "cos-tool-{}".format(arch)
         try:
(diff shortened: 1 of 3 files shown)


rbarry82:
I wonder if this is also the cause for the unit test failure here (where the value is not mocked).

Should we fail back to `x86_64`? It's only `arm64` and `x86_64`, and an equivalent to:

```go
switch(runtime.GOARCH); {
case "arm64":
  binary = "cos-tool-arm64"
default:
  binary = "cos-tool-amd64"
}
```

Is probably a sane enough default

simskij:
> I wonder if this is also the cause for the unit test failure here (where the value is not mocked).
> 
> Should we fail back to `x86_64`? It's only `arm64` and `x86_64`, and an equivalent to:
> 
> ```go
> switch(runtime.GOARCH); {
> case "arm64":
>   binary = "cos-tool-arm64"
> default:
>   binary = "cos-tool-amd64"
> }
> ```
> 
> Is probably a sane enough default

I think it's probably good that it breaks if the user were to use some other fringe architecture, allowing them to raise an issue about having the arch added instead of troubleshooting for ages.

rbarry82:
In that case, it's probably worth using the [Run-On-Arch action](https://github.com/uraimo/run-on-arch-action) to ensure that these are actually set.

Whatever scenario in the GH runners (Google hasn't turned up much) which leads to flaky architecture detection should be squashed unless we intend to globally patch it at the top of every unit test. Or use Run-On-Arch

PR #318, opened 2022-06-21
Use compute resource patch to limit resources

sed-i:
## Issue
Need to be able to limit resource usage of a charm.

Crossref: [OPENG-273](https://warthogs.atlassian.net/browse/OPENG-273)

## Solution
- Use lib to patch k8s compute resource `limits` and `requests`.
- Automatically deduce "requests" from "limit"

## Context
[observability-libs/19](https://github.com/canonical/observability-libs/pull/19).

## Testing Instructions
- Deploy charm and look at `kubectl describe statefulset -n welcome prom`.
- itests.

## Release Notes
Use compute resource patch to limit resources.


Abuelodelanada (approved):

simskij:
We should do [this](https://github.com/canonical/observability-libs/pull/19#issuecomment-1164044084) prior to merging this PR.

src/charm.py
```diff
@@ -47,6 +52,67 @@
 logger = logging.getLogger(__name__)
 
 
+def monotonize(value, default_value, multiplier):
+    """A continuous function from the given value to a multiplier-adjusted value.
```
balbirthomas, line 56: Docstrings for parameters and return value will be nice.

src/charm.py
```diff
@@ -47,6 +52,67 @@
 logger = logging.getLogger(__name__)
 
 
+def monotonize(value, default_value, multiplier):
```
balbirthomas, line 55: 1. Is this method only used in the context of the Kubernetes Resource patch library ? If so can it be a utility method provided by the resource patch library ?
2. Are the function arguments expected to be non-negative ?
3. If arguments are non-negative, given what the code is doing would it make sense for the function signature to be ```
```
monotonize(default_value, multiplier, value = 0):
    assert default_vale >= 0
    assert multiplier >= 0
    assert multiplier < 1
    assert value >=0
```
5. The name of this function, its docstring and body do not seem to concur. The body suggests that this function is trying to do find something along the lines of a maximum (given above suggested signature) of "value", "default_value" and "multiplier*default_value". Is this correct ? Or alternatively could something like a maximum (or `find_maximum()`) be used instead of `monotonize()` ? Would this be more suggested of intended purpose ? The word "monotonize" suggests that this method takes in an unordered series and returns an ordered Monotonic series.
7. Is this really a [continuous function](https://en.wikipedia.org/wiki/Continuous_function) ? Or rather does the continuity of this function have any relevance to its purpose ? If not could the docstring  avoid any references to continuity ?

src/charm.py
```diff
@@ -47,6 +52,67 @@
...
+    if adjusted_value > default_value:
+        return adjusted_value
+
+    if value > default_value:
+        return default_value
+
+    return value
+
+
+def convert_k8s_quantity_to_legacy_binary_gigabytes(
```
balbirthomas, line 82: Could this be a utility methods in the Kubernetes resource patch library ? Seems like it is useful primarily in that context.

src/charm.py
```diff
@@ -47,6 +52,67 @@
...
+    capacity: str, multiplier: Union[Decimal, float, str] = 1.0
+) -> str:
+    """Convert a K8s quantity string to legacy binary notation in GB, which prometheus expects.
```
balbirthomas, line 85: Docstrings for Parameters and return value would be nice.

src/charm.py
```diff
@@ -110,6 +188,47 @@ def __init__(self, *args):
         self.framework.observe(
             self.on.validate_configuration_action, self._on_validate_configuration
         )
+        self.framework.observe(self.resources_patch.on.patch_failed, self._on_k8s_patch_failed)
+
+    @classmethod
```
balbirthomas, line 193: Does any client of this method have the need to use this _private method_ without an instance of the charm available, as in `PrometheusCharm._limits_to_requests()` ? If not, is there any reason for this to be a `classmethod` ? Could it just be a normal private instance method ?

src/charm.py
```diff
@@ -110,6 +188,47 @@ def __init__(self, *args):
...
+    def _limits_to_requests(cls, resource_limits: ResourceSpecDict):
+        """Calculate "requests" from a "limits" dict.
+
+        By default, we set the "requests" portion of the resource limits to a sensible value, while
```
balbirthomas, line 197: Is there any thing specific to Prometheus in the logic of this function ? If not could this be a method in the Kubernetes resource patch library ?

src/charm.py
```diff
@@ -110,6 +188,47 @@ def __init__(self, *args):
...
+        keeping the "limits" portion unspecified.
+        If the user specifies a "limits" (via config options), then the "requests" portion is
+        calculated to maintain a monotone function, as follows:
```
balbirthomas, line 200: The use of phrase "monotone function" is strictly not correct here (please see my comments above). Would it make more sense to call this methods something like `_constrained_requests(limits: dict)` ? Thereby suggesting that this method constrains the kubernetes requests to honor the user specified limits.

src/charm.py
```diff
@@ -110,6 +188,47 @@ def __init__(self, *args):
...
+        - Calculate the adjusted limit by multiplying the user value by 0.8.
+        - If the adjusted value is larger than the default "requests", use the adjusted value for
```
balbirthomas, line 202: I appreciate the effort this detailed docstring. However would this be clearer if said into code ? And instead would it make more sense to make clear in the docstring what is the intended purpose of this logic -- for example to ensure that the "requests" are not more than the "limits"  and that this constraining uses a value of limits that is scaled down by a multiplier.

Also isn't this functionality more rightly placed in the context of the kubernetes resource patch library along with the other methods above. ?

Abuelodelanada (approved):

src/charm.py
sed-i: Fixed in next commit, lemme know what you think!

src/charm.py
sed-i: Fixed in next commit, lemme know what you think!

src/charm.py
balbirthomas: Thanks. Though I am not aware myself, it may be worth considering if there are other applications which may want to use the Kubernetes resource patch library that do similar things like Prometheus in this regard

balbirthomas (approved):

src/charm.py
balbirthomas: 1. If you think this is also going to be used in other places in COS then doesn't it make sense to provide it as a utility function in the Resource patch library.
2. The code below does not translate to the function definition of $y(x; y_0, 0 \lt k \lt 1)$ described above. For example the value `None` is not a mathematically meaningful value for $x$ or $y_0$ and $k$ which translates to `value` and `default_value` and `multiplier` in code where either can get a value `None`. Hence it is a bit misleading to think of body of `monotonize()` in terms of $y()$ since $x \leq \mathrm{None}$ or $y_0 \lt \mathrm{None} \leq y_0/k$ does not make mathematical sense. As shown in the code below `value` can be `None`. For these reasons I think the math here is a digression and we would be better of just looking at the code.
3. As I understand all of these "resource" and "limits" should never be negative because they represent real quantities that can not be negative.
4.  "monotonize" is bound to suggest to any cognizant person that this function transforms its arguments by putting them in a [monotonic ordering](https://en.wikipedia.org/wiki/Monotonic_function). But this function accepts multiple values and returns a single value (not a reordering of the same multiple input values). If you are indeed seeking a better name for this function then we could consider the purpose of this function in the context of the code were it is intended to be used rather than the loose mathematical analogy discussed above. As I understand this method is used in `limits_to_requests()` where it is used to constrain the Kubernetes resource "request" after taking into consideration the Kubernetes resource "limits" and a "scaling factor" (termed "multiplier"). If this understanding is correct then would a person reading this code find it easier to understand the intended purpose of this function if it were termed something like `constrain_k8s_request(limits, requests, scaling)` or something similar ?
5. This is probably a digression and not relevant to the intended purpose of this function and its code.

I do not mean to belabor these points, but I did struggle to understand the code. I will leave it to your better judgement.

sed-i: Moved to https://github.com/canonical/observability-libs/pull/19 and refactored a bit.
Renaming/removal of the function(s) there too is still pending.

sed-i:
As-is, itests will fail until:
- https://github.com/canonical/microk8s-core-addons/pull/73
- https://github.com/juju/python-libjuju/pull/696

are merged and released.

PR #334, opened 2022-08-01
Report Prometheus version correctly

jnsgruk:
## Issue
The operator does not report the running version of Prometheus to Juju

## Solution
Report version during the `_configure` hook, and also add an `update-status` handler to report the version on each `update-status` event

### DIFF
--- a/src/charm.py
+++ b/src/charm.py
@@ -76,6 +76,7 @@ class PrometheusCharm(CharmBase):
         self.ingress = IngressPerUnitRequirer(self, relation_name="ingress", port=self._port)
 
         external_url = urlparse(self._external_url)
+        self._prometheus_server = Prometheus(web_route_prefix=external_url.path)
 
         # Exposes remote write endpoints
         self.remote_write_provider = PrometheusRemoteWriteProvider(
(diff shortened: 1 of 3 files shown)


rbarry82 (approved):
LGTM. We had initially hoped to do this with a version built into the charm. Mostly, there wasn't a clear story around "should we report the version of the _charm_ or the version of the _workload_"? 

For bug/issue reporting, knowing which version of the _charm_ is deployed in any given place potentially provides more useful context, but I'm ok with either.

jnsgruk:
(looks like the integration test failure is because python-libjuju isn't up to date and breaks the traefik integration test, so should be safe to merge)

src/charm.py
```diff
@@ -178,9 +178,13 @@ def _configure(self, _):
         # Make sure that if the remote_write endpoint changes, it is reflected in relation data.
         self.remote_write_provider.update_endpoint()
         self.grafana_source_provider.update_source(self._external_url)
-
+        self.unit.set_workload_version(self._prometheus_server.version())
         self.unit.status = ActiveStatus()
 
+    def _update_status(self, event):
+        """Fired intermittently by the Juju agent."""
+        self.unit.set_workload_version(self._prometheus_server.version())
```
sed-i, line 186: Is `update-status` added in case the workload app is ready on `pebble-ready`?

sed-i (approved):
Neat. This OF feature was implemented 2yr ago. We should use it!

src/charm.py
jnsgruk: Yeh, it's essentially a catch all in case for whatever reason Prometheus isn't quite ready at the end of the pebble-ready hook. The implementation of `version` on the `Prometheus` server provides no guarantees

simskij (approved):

PR #335, opened 2022-08-02
Enable override of interface name

jnsgruk:
## Issue
The Prometheus `scrape_targets` config structure is used by quite a few different applications. The lib in its current form means that we'll always need to use the `prometheus_scrape` interface name for this library to work, which may introduce unforseen bugs if there are multiple apps using the same interface structure, and could get wired up wrong at runtime.

## Solution
This is submitted for discussion - and would enable the ability for charm developers to use the library for a custom relation name. The changes only cover `MetricsEndpointProvider` and `MetricsEndpointConsumer` at this point. I don't forsee the `MetricsEndpointAggregator` being universally useful in the same way.

### DIFF
--- a/lib/charms/prometheus_k8s/v0/prometheus_scrape.py
+++ b/lib/charms/prometheus_k8s/v0/prometheus_scrape.py
@@ -21,17 +21,17 @@ object from this charm library. For the simplest use cases, using the
 typically in the constructor of your charm (the one which exposes a
 metrics endpoint). The `MetricsEndpointProvider` constructor requires
 the name of the relation over which a scrape target (metrics endpoint)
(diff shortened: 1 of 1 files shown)


jnsgruk:
Debating in my mind if a better plan here is to fork the lib?

sed-i:
Iiuc, the working assumptions so far were:
- libs manage relations.
- we conduct ourselves as if relation interface names are unique throughout the ecosystem, and "owned" by one lib.

> multiple apps using the same interface structure

Would you mind clarifying the context?

jnsgruk:
Yeh, I'm gonna rethink this.

PR #336, opened 2022-08-03
Remove incorrect warning from lib

jnsgruk:
## Issue
Incorrect warning is issued in the logs on machine charms warning about containers. I think the warning was designed for Kubernetes charms, but the symptom it warns about is mostly mitigated by the inclusion of pebble_ready as the default refresh event.

## Solution
Remove the log entry!

### DIFF
--- a/lib/charms/prometheus_k8s/v0/prometheus_scrape.py
+++ b/lib/charms/prometheus_k8s/v0/prometheus_scrape.py
@@ -339,7 +339,7 @@ LIBAPI = 0
 
 # Increment this PATCH version before using `charmcraft publish-lib` or reset
 # to 0 if you are raising the major API version
-LIBPATCH = 20
+LIBPATCH = 21
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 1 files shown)


rbarry82:
This was actually added to check for podspec charms, too, rather than defaulting to `pebble_ready`.

Now that it's being used on machine charms (👍 by the way!), the "goal" of the `refresh_event` argument is, basically, "if the charm which instantiated this library is restarted/rescheduled/has an IP change, please update the endpoint". The goals there should probably be updated, or we can pass `refresh_event`:

```python
        if not refresh_event:
            if len(self._charm.meta.containers) == 1:
                if "kubernetes" in self._charm.meta.series:
                    # This is a podspec charm
                    refresh_event = [self._charm.on.update_status]
                else:
                    # This is a sidecar/pebble charm
```

We weren't really able to find anything better than `update_status` for podspec charms, but if there's a reasonable pattern for machine charms ( `on.start`, `on.install`, etc), it would be better to add that. 

Really, `prometheus_scrape` is the first one of the libraries doing this, since scraping is a pretty general target, but it's a pattern, even a small one, which we should probably extract out so it can be reused by other observability (and potentially other) libraries as a generalized "my IP address may have changed, go update peer data/relation data/etc to for gossip/cluster/scrape"

rbarry82 (changes requested):
Let's see if we can determine the right event for machine charms, or `getattr` for `pebble_ready` regardless of whether or not `charm.meta.containers` has a set value or not.

PR #340, opened 2022-08-08
Fix validation logic and other things

sed-i:
This PR was split into:
- #348 
- #349; as well as:
  - #370
  - #373
- Another TBD PR with the AlertRulesProvider merge.


sed-i:
> We explicitly synthesized a temporary file because the reldata was broken
> We should probably test that. I don't remember which charm was creating the double nesting in the first place. Maybe grafana agent, and my dedupe patch squashd it
> we [need to] build a list of what different rulesets may look like, because the built-in rules clearly aren't handled the same way, and we should go through a matrix to figure out which cases we need to test.
> -- @rbarry82 

I'm thinking about these cases for provider's rule files:
- one file with one rule per file, and another file with a number of groups and a number of alerts
- one rule per file, and the alertname same as filename.
- multiple rules per file, and with the same alertname, groupname and filename.

Anything else?

dstathis (approved):

src/charm.py
```diff
@@ -73,19 +73,27 @@ def __init__(self, *args):
             self._name,
             resource_reqs_func=self._resource_reqs_from_config,
         )
+        self.framework.observe(self.resources_patch.on.patch_failed, self._on_k8s_patch_failed)
+
+        self.ingress = IngressPerUnitRequirer(self, relation_name="ingress", port=self._port)
+        self.framework.observe(self.ingress.on.ready_for_unit, self._on_ingress_ready)
+        self.framework.observe(self.ingress.on.revoked_for_unit, self._on_ingress_revoked)
```
Abuelodelanada, line 80: Just to make the code more legible... what about placing all the `self.framework.observe` all together?

Abuelodelanada (approved):

src/charm.py
sed-i: There was some back and forth about this.
Iirc we concluded to split them and have the `observe` right next to the lib that emits those events.

rbarry82 (changes requested):
As a general comment, this is... fine. It very much looks like the scope exploded.

I'm not sure why `_is_valid_unit_address` moved to a top-level function, since it's only used by one class, and only in one place. If anything, as a restructure, `_is_official_alert_rule_format` and `_is_single_alert_rule_format` could be moved under the appropriate class (also only used in one place) and `_is_valid_unit_address` could stay under the class to make the code more literate and decrease the amount of "bouncing around" a patch author, reviewer, or anyone reading the code needs to do in order to follow the flow of the code.

`PrometheusScrapeRelationData` is over-engineered and over-abstracted. Use of `@classmethod` when none of the methods are factories returning an instantiated `cls` object nor are they accessing class-level variables is an antipattern. 

These would be more suitable as `@staticmethod` and dropping the `cls` argument in that case. 

It is not at all clear why any of them are returning tuples, or why we have new methods abstracted into a class which are called only once, and whose callers discard half of the return value with `_`. All of these methods could/should simply take a flat argument (as an enum, const, or whatever) which says "return either app or unit data", which makes it clear to the caller what we are hoping to get back, and return only one element of the tuple.

But as above, moving these out forces the reader, even one who is otherwise familiar with the codebase, to jump to a different place in the file, mentally map any class attributes into method arguments, then back, in order to keep track of what this class is actually doing, what it's returning, and what we're going to do with it. It sacrifices clarity and legibility, an I'm not sure what the payoff is.

I'm sure that there is one, conceptually, but I cannot see, as a reviewer, what benefit I gain from the rewritten `_set_reldata_scrape_host`, for example.

I now need to go to the constructor (if I don't remember) to see what may have been set for `_scrape_jobs` or `_external_hostname`, go back to `_set_reldata_scrape_host`, keep track of the approximate values to follow conditionals in a method which, bluntly, would fit into a single terminal window on a PDP in order to figure out what is actually being set in relation data, because `for k, v in unit_data.items():` doesn't give me any indication of what is is in `unit_data` or why we threw away one of the return values unless I do this.

This really seems like it should be split into two (or even three) separate PRs. 
* Fix validation logic (the changes in `CosTool`), which is isolated, tested, and fixes a clear issue. More or less the changes to `prometheus_remote_write` also applied to `prometheus_scrape`
* Update `IngressPerUnitRequirer` logic. This seems to be most of the changes in `src/charm.py`
  -Eeven there, I would suggest that "call `_configure|_common_exit_hook` hoping it will resolve `BlockedStatus` if an event was emitted before ingress is ready is a sledgehammer which is bad considering we already have an issue about `/-/reload` being called to often, what whatever logic it would take for "is `BlockedStatus` set during `_update_status`? Is it still valid? If it is, clear it, if not, move on) would be more explicit, cleaner, and more legible.
  - Even seeing the comment and reading `_configure`, I have no idea why this would be true. Either `current_services == new_layer.services` and it's never restarted at all, so we never try to connect, or the change to `set_workload_version` by calling the binary rather than an endpoint means that the presumptive bypass by calling `self._prometheus_server.version()` will no longer call the endpoint, and this won't actually be set.
  - In reading through this again, it also seems wrong that `# Reload may fail if ingress ...` wold lead to a `BlockedStatus(CORRUPT_PROMETHEUS_CONFIG_MESSAGE)`. It's not corrupt, though. Ingress is not ready, or Prometheus has taken too long to come up. This could potentially lead administrators down a "blind alley"
  - The "right" fix for this is likely:
    - Add a `can_connect()` method to `prometheus_server.Prometheus()`. If `update_status` fires and it's `BlockedStatus` with a message (a _new_ message which indicates the actual state) about timeouts:
```python
def _update_status(self, event):
  ...
  if isinstance(self.unit.status, BlockedStatus) and self.unit.status.message == PROMETHEUS_TIMEOUT:
    if self._prometheus_server.can_connect():
```
    - In addition, `prometheus_server.Prometheus.*` should provide distinctions between timeouts and `ConnectionError` to flag the "real" status. Custom exceptions instead of a boolean are a trivial, Pythonic way to do this
* Whatever is happening with the refactor, so it can be addressed separately. As above, the `@classmethod` is an antipattern, and the abstraction dramatically impacts how literate the code is. The intention seems to be to provide a "stock" way for `cos-config` to interact with the library, but there are a lot of changes, and it's difficult to address them without knowing what the end goal is
  - `PrometheusRulesProvider` is a single-client class which is 55 lines long. Even if it were extracted into a separate file for `cos-config` to use which needed to `from ...prometheus_scrape import AlertRules, InvalidAlertRulePathError, _resolve_dir_against_charm_path`, it would be dramatically simpler with less side effects
  - Even with all of these changes, `cos-config` still presumably needs to use a "backdoor". Couldn't it now pass the git sync events into `refresh_event`? In combination with the change to `_scrape_jobs` from this patch be 99% there?
  - More generally, I'm concerned about a behavior change here. `_set_reldata_scrape_host` does not do the same thing as `_set_scrape_job_spec` did, and that may matter a lot. In particular, `_set_scrape_job_spec`, if the unit was the leader, also re-set alertrules/scrape jobs/etc. In the case of machine charms or other cases which could actually be setting a more explicit scrape job which is built from dynamic data in the charm code, this is an important distinction, and it's not clear why this was broken out into multiple methods which could adversely affect behavior either.
  
For the last case, from the outside, even from the case of "remove `PrometheusRulesProvider` and provide a way for `cos-proxy` to still operate", it very much seems like the change to `MetricsEndpointProvider._scrape_jobs` to not always return `DEFAULT_JOB` if it's empty as well as the suggested change in the constructor would be nearly sufficient. Then if we had something very trivial such as:

`cos-config/src/rules_provider.py`
```python
from charms.observability_libs.v0.juju_topology import JujuTopology
from charms.prometheus_k8s.v0.prometheus_scrape import MetricsEndpointProvider

class CosConfigIdentifier(JujuTopology):
  group_name_prefix = ""
  @property
  def identifier(self) -> str:
    return self.group_name_prefix
    
class CosRulesProvider(MetricsEndpointProvider):
  def __init__(self, charm: CharmBase, refresh_event=..., group_name_prefix: str = ""):
    super().__init__(self, charm, refresh_event=...)
    self.topology = CosConfigIdentifier.from_dict({
      "model": "",
      "model_uuid": str(uuid.uuid4()),
      "application": ""
    })
    self.topology.group_name_prefix = group_name_prefix
```

And this provides 100% of what's needed to replace `PrometheusRulesProvider` for its only actual user while also keeping exactly the same behavior as `cos-config` currently has.

Until when/if a time exists that someone/anyone else will actually need the functionality, it sandboxes away this specific use case. 

If we as a team would rather modify `MetricsEndpointProvider` and `AlertRules` via `group_name_prefix` to also support this use case, then let's do it, but the existence of and changes around `PrometheusScrapeRelationData` should be carefully considered for the reasons above. To reiterate, though:
* it's not quite a dataclass, but sort-of is
* it doesn't actually have any generalized logic -- it's all operations on attributes/variables which exist in the caller
* it's impossible to determine from the caller (without reading it) what you are getting back and why, nor is it possible to determine from the callee why either of `app_data` or `unit_data` should be generated and whether either will be used
* it forces authors/readers/reviewers to add mental load to keep track of the flow of the code and what may be set in relation data
* addition of a new top-level method with a single caller which could trivially stay bound to the class it "belongs" to
* the usage of `@classmethod` is an antipattern, and it could be `@staticmethod` if anything
* all methods contained within are called only from a single place. why are they somewhere disconnected in the codebase?
* the changes around what happens to `refresh_event` are potentially significant for non-trivial use cases of `MetricsEndpointProvider` which currently work as expected; this would be a change in what we provide, without a clear reason for why they are being separated

lib/charms/prometheus_k8s/v0/prometheus_remote_write.py
```diff
@@ -1024,9 +1008,15 @@ def validate_alert_rules(self, rules: dict) -> Tuple[bool, str]:
                 return True, ""
             except subprocess.CalledProcessError as e:
                 logger.debug("Validating the rules failed: %s", e.output)
-                return False, ", ".join([line for line in e.output if "error validating" in line])
+                return False, ", ".join(
+                    [
+                        line
+                        for line in e.output.decode("utf8").strip().split("\n")
```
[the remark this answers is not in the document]
sed-i: The strip removes the extra `''` element at the end of the list, but indeed as you suggest, `splitlines` obviates it.
Fixed.

sed-i:
Merging `AlertRulesProvider` (as well as `MetricsAggregator`?) into `MetricsEndpointProvider` can be considered for `v1`. 
Closing for now.

PR #343, opened 2022-08-17
Add additional unit_address arg provided by Ingress

Abuelodelanada:
## Issue

Charms that expose a metrics endpoint via a `MetricsEndpointProvider` AND that use an ingress, must have ways to enforce the usage of the ingress-provided URL in the targets that the Prometheus server receives.

In terms of a scrape job definition, and within the current confines of the prometheus_scrape library, which does not (yet) support authentication and encryption, the elements of a scrape job that are affected by the charm providing the metrics endpoint to be proxied by an ingress are:

1. `hostname` in the targets
2. `port` in the targets
3. `metrics_path` in the scrape job

## Solution

A new option argument was added to `MetricsEndpointProvider`

## Context

This PR will not work correctly unti we fix traefic's ingress_per_unit lib since its design requires charm to "know" not to use `.url` after a `revoked` event.


lib/charms/prometheus_k8s/v0/prometheus_scrape.py
```diff
@@ -1345,6 +1345,7 @@ def __init__(
         jobs=None,
         alert_rules_path: str = DEFAULT_ALERT_RULES_RELATIVE_PATH,
         refresh_event: Optional[Union[BoundEvent, List[BoundEvent]]] = None,
+        unit_address: str = None,
```
sed-i, line 1348: - Elsewhere we call it `external_url`, might be better for clarity.
- Seems like this design relies on charm-reinit every event. Is it a valid assumption?

Abuelodelanada: * Seems reasonable... but I think in this case the name should be `external_address` because it is not an `URL`
* The design is relying in the `refresh_event` parameter. I'm testing this modifications with the following code in  `__init__` method in Loki charm:

```python
        self.parsed_external_url = urlparse(self._external_url)
        scrape_jobs = [
            {
                "metrics_path": f"{self.parsed_external_url.path}/metrics",
                "static_configs": [{"targets": [f"*:{self.parsed_external_url.port}"]}],
            }
        ]
        self.metrics_provider = MetricsEndpointProvider(
            self,
            jobs=scrape_jobs,
            refresh_event=[
                self.ingress_per_unit.on.ready_for_unit,
                self.ingress_per_unit.on.revoked_for_unit,
                self.on.ingress_relation_departed,
            ],
            unit_address=self.parsed_external_url.hostname,
        )
```

Please, let me know your thoughts!

sed-i: - In that case we should probably call it `external_hostname`, just [like traefik does](https://github.com/canonical/traefik-k8s-operator/blob/494741059d931a9827e8214401f61cde8585582e/config.yaml#L5). "Address" might be ambiguous.
- A refresh event seems great.

rbarry82 (approved):

dstathis (approved):

PR #346, opened 2022-08-23
Consolidate CI

simskij:
Move away from duplicating CI files across all repos. Instead, use workflow_call to have them be shared from the `canonical/observability` repo. Unfortunately, this can't really be tested without merging, so once merged, we need to run through the workflows and make sure they work before proceeding to the next charm.


.github/workflows/pull-request.yaml
```diff
@@ -6,92 +6,10 @@ on:
...
-      - name: Install dependencies
-        run: python -m pip install tox
-      - name: Run tests
-        run: tox -e unit
-
-  integration-test:
-    name: Integration tests (microk8s)
-    uses: ./.github/workflows/integration-tests.yml
+  pull-request:
+    uses: canonical/observability/.github/workflows/pull-request.yaml@main
```
Abuelodelanada, line 10: Should this file be present in [this directory](https://github.com/canonical/observability-libs/tree/main/.github/workflows)?
🤔

Abuelodelanada (approved):

.github/workflows/pull-request.yaml
simskij: No, `observability` - not `observability-libs` :)

Abuelodelanada: small detail! ;-)

dstathis (approved):

Issue #342, opened 2022-08-16
Reload is called unnecessarily


sed-i:
### Bug Description

Currently, every `_configure`, the charm would either reload configuration

https://github.com/canonical/prometheus-k8s-operator/blob/e75eb0b1a7669c945dabb6b4ba8cfe583697533b/src/charm.py#L272-L273

or replan:

https://github.com/canonical/prometheus-k8s-operator/blob/e75eb0b1a7669c945dabb6b4ba8cfe583697533b/src/charm.py#L281-L282

This means the prometheus service will be unavailable for potentially significant duration (e.g. big wal replay).

### To Reproduce

NA

### Environment

NA

### Relevant log output

```shell
NA
```

### Additional context

- This came up while adding a call to `_configure` in `update-status` as a workaround for https://github.com/canonical/traefik-k8s-operator/issues/78.
- Note:  WAL replay can take longer than timeout duration.

sed-i:
Also as part of this change we need to consider comments by @rbarry82 on #349:

> "call `_configure|_common_exit_hook` hoping it will resolve BlockedStatus if an event was emitted before ingress is ready is a sledgehammer which is bad considering we already have an issue about `/-/reload` being called to often

Agreed, but to limit the scope of this change, and given #342 will be addressed very soon, can live with this for now?

> what whatever logic it would take for "is BlockedStatus set during _update_status? Is it still valid? If it is, clear it, if not, move on) would be more explicit, cleaner, and more legible.

This would be in direct contrast to the common exit hook pattern. I'd rather not mix both approaches in the same charm. Also, once #342 is fixed this will be much less painful to look at.

> seems wrong that # Reload may fail if ingress ... wold lead to a `BlockedStatus(CORRUPT_PROMETHEUS_CONFIG_MESSAGE)`

Agreed. I don't particularly like that status comparison pattern and it should be fixed as part of #342.

> prometheus_server.Prometheus.* should provide distinctions between timeouts and ConnectionError to flag the "real" status.

Sounds presumptuous at first glance. And what will the charm do with the various exceptions? Just choose one message over another? Maybe, if we have insight, expose a `retry` setting the charm could use?

_Originally posted by @sed-i in https://github.com/canonical/prometheus-k8s-operator/pull/349#discussion_r961335529_

rbarry82:
> Agreed, but to limit the scope of this change, and given #342 will be addressed very soon, can live with this for now?

[...]

PR #349, opened 2022-09-02
Fix ingress usage

sed-i:
(Note: this PR was split out of #340.)

## Issue
The self-scrape job is "down" when prometheus is related to traefik.The external hostname is not passed to prom_scrape constructor for the self-scraping endpoint.

## Solution
Use external_url in constructing the self-scraping job.

Note: the quality of the solution in this PR is affected by https://github.com/canonical/traefik-k8s-operator/issues/78.

## Testing Instructions
Deploy prom and relate to traefik. Then look at prom's `/api/v1/targets`.

## Release Notes
Fix ingress usage.

### DIFF
--- a/src/charm.py
+++ b/src/charm.py
@@ -89,25 +89,31 @@ class PrometheusCharm(CharmBase):
             resource_reqs_func=self._resource_reqs_from_config,
         )
 
+        self.ingress = IngressPerUnitRequirer(self, relation_name="ingress", port=self._port)
+
         self._topology = JujuTopology.from_charm(self)
 
+        external_url = urlparse(self.external_url)
+
+        # TODO when https://github.com/canonical/traefik-k8s-operator/issues/78 is fixed, change
+        #  the refresh event to self.ingress.on.ready_for_unit.
         self._scraping = MetricsEndpointProvider(
             self,
             relation_name="self-metrics-endpoint",
             jobs=self.self_scraping_job,
+            external_hostname=external_url.netloc,
+            refresh_event=self.on.update_status,  # needed for ingress
         )
         self.grafana_dashboard_provider = GrafanaDashboardProvider(charm=self)
         self.metrics_consumer = MetricsEndpointConsumer(self)
-        self.ingress = IngressPerUnitRequirer(self, relation_name="ingress", port=self._port)
 
-        external_url = urlparse(self.external_url)
         self._prometheus_server = Prometheus(web_route_prefix=external_url.path)
 
         self.remote_write_provider = PrometheusRemoteWriteProvider(
             charm=self,
             relation_name=DEFAULT_REMOTE_WRITE_RELATION_NAME,
             endpoint_address=external_url.hostname or "",
-            endpoint_port=external_url.port or self._port,
+            endpoint_port=external_url.port or 80,
             endpoint_schema=external_url.scheme,
             endpoint_path=f"{external_url.path}/api/v1/write",
         )
@@ -154,9 +160,27 @@ class PrometheusCharm(CharmBase):
         self.framework.observe(self.on.validate_configuration_action, self._on_validate_config)
 
     @property
+    def metrics_path(self):
+        """The metrics path, adjusted by ingress path (if any)."""
+        return urlparse(self.external_url).path.rstrip("/") + "/metrics"
+
+    @property
     def self_scraping_job(self):
         """The scrape job used by Prometheus to scrape itself during self-monitoring."""
-        return [{"static_configs": [{"targets": [f"*:{self._port}"]}]}]
+        # TODO https://github.com/canonical/prometheus-k8s-operator/issues/360
+        # A single scrape job cannot accommodate more than one unit with ingress-per-unit.
+        # With ingress, each unit is expected to have a different path. This means that each
+        # unit needs to have its own static_config, because `metrics_path` is on the same
+        # hierarchical level as `static_configs` (for this reason, *-notation cannot be used).
+        # However, scrape jobs are placed in app data, so only the leader unit can set it, and the
+        # leader unit must be able to obtain addresses of all units.
+        port = urlparse(self.external_url).port or 80
+        return [
+            {
+                "metrics_path": self.metrics_path,
+                "static_configs": [{"targets": [f"*:{port}"]}],
+            }
+        ]
 
     @property
     def log_level(self):
@@ -180,7 +204,7 @@ class PrometheusCharm(CharmBase):
             "job_name": "prometheus",
             "scrape_interval": "5s",
             "scrape_timeout": "5s",
-            "metrics_path": "/metrics",
+            "metrics_path": self.metrics_path,
             "honor_timestamps": True,
             "scheme": "http",
             "static_configs": [
@@ -410,10 +434,13 @@ class PrometheusCharm(CharmBase):
         return True
 
     def _update_layer(self, container) -> bool:
-        current_services = container.get_plan().services
+        current_planned_services = container.get_plan().services
         new_layer = self._prometheus_layer
 
-        if current_services == new_layer.services:
+        current_services = container.get_services()  # mapping from str to ServiceInfo
+        all_svcs_running = all(svc.is_running() for svc in current_services.values())
+
+        if current_planned_services == new_layer.services and all_svcs_running:
             return False
 
         container.add_layer(self._name, new_layer, combine=True)
--- a/src/charm.py
+++ b/src/charm.py
@@ -95,14 +95,16 @@ class PrometheusCharm(CharmBase):
 
         external_url = urlparse(self.external_url)
 
-        # TODO when https://github.com/canonical/traefik-k8s-operator/issues/78 is fixed, change
-        #  the refresh event to self.ingress.on.ready_for_unit.
         self._scraping = MetricsEndpointProvider(
             self,
             relation_name="self-metrics-endpoint",
             jobs=self.self_scraping_job,
             external_url=self.external_url,
-            refresh_event=self.on.update_status,  # needed for ingress
+            refresh_event=[  # needed for ingress
+                self.ingress.on.ready_for_unit,
+                self.ingress.on.revoked_for_unit,
+                self.on.update_status,
+            ],
         )
         self.grafana_dashboard_provider = GrafanaDashboardProvider(charm=self)
         self.metrics_consumer = MetricsEndpointConsumer(self)
--- a/src/charm.py
+++ b/src/charm.py
@@ -101,7 +101,7 @@ class PrometheusCharm(CharmBase):
             self,
             relation_name="self-metrics-endpoint",
             jobs=self.self_scraping_job,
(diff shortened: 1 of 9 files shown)


src/charm.py
```diff
@@ -308,6 +311,11 @@ def _update_status(self, event):
         """Fired intermittently by the Juju agent."""
         self.unit.set_workload_version(self._prometheus_server.version())
 
+        # Call `_configure`` to resolve "BlockedStatus" when the ready-for-unit event is processed
+        # before ingress is in fact ready.
+        # https://github.com/canonical/traefik-k8s-operator/issues/78
+        self._configure(event)
```
sed-i, line 317: Copying comment by @rbarry82 from https://github.com/canonical/prometheus-k8s-operator/pull/340#pullrequestreview-1093589121:

> -   Update `IngressPerUnitRequirer` logic. This seems to be most of the changes in `src/charm.py`  
>     - Even there, I would suggest that "call `_configure|_common_exit_hook` hoping it will resolve `BlockedStatus` if an event was emitted before ingress is ready is a sledgehammer which is bad considering we already have an issue about `/-/reload` being called to often, what whatever logic it would take for "is `BlockedStatus` set during `_update_status`? Is it still valid? If it is, clear it, if not, move on) would be more explicit, cleaner, and more legible.
>     - Even seeing the comment and reading `_configure`, I have no idea why this would be true. Either `current_services == new_layer.services` and it's never restarted at all, so we never try to connect, or the change to `set_workload_version` by calling the binary rather than an endpoint means that the presumptive bypass by calling `self._prometheus_server.version()` will no longer call the endpoint, and this won't actually be set.
>     - In reading through this again, it also seems wrong that `# Reload may fail if ingress ...` wold lead to a `BlockedStatus(CORRUPT_PROMETHEUS_CONFIG_MESSAGE)`. It's not corrupt, though. Ingress is not ready, or Prometheus has taken too long to come up. This could potentially lead administrators down a "blind alley"
>     - The "right" fix for this is likely:
>         - Add a `can_connect()` method to `prometheus_server.Prometheus()`. If `update_status` fires and it's `BlockedStatus` with a message (a _new_ message which indicates the actual state) about timeouts:
> 
> ```
> def _update_status(self, event):
>   ...
>   if isinstance(self.unit.status, BlockedStatus) and self.unit.status.message == PROMETHEUS_TIMEOUT:
>     if self._prometheus_server.can_connect():
>       self.unit.status = ActiveStatus()
> 
> ```
> - In addition, `prometheus_server.Prometheus.*` should provide distinctions between timeouts and `ConnectionError` to flag the "real" status. Custom exceptions instead of a boolean are a trivial, Pythonic way to do this

sed-i: > "call `_configure|_common_exit_hook` hoping it will resolve BlockedStatus if an event was emitted before ingress is ready is a sledgehammer which is bad considering we already have an issue about `/-/reload` being called to often

Agreed, but to limit the scope of this change, and given #342 will be addressed very soon, can live with this for now?

> what whatever logic it would take for "is BlockedStatus set during _update_status? Is it still valid? If it is, clear it, if not, move on) would be more explicit, cleaner, and more legible.

This would be in direct contrast to the common exit hook pattern. I'd rather not mix both approaches in the same charm. Also, once #342 is fixed this will be much less painful to look at.

> seems wrong that # Reload may fail if ingress ... wold lead to a `BlockedStatus(CORRUPT_PROMETHEUS_CONFIG_MESSAGE)`

Agreed. I don't particularly like that status comparison pattern and it should be fixed as part of #342.

> prometheus_server.Prometheus.* should provide distinctions between timeouts and ConnectionError to flag the "real" status.

Sounds presumptuous at first glance. And what will the charm do with the various exceptions? Just choose one message over another? Maybe, if we have insight, expose a `retry` setting the charm could use?

sed-i: I mentioned this on #342.

rbarry82: > > This would be in direct contrast to the common exit hook pattern. I'd rather not mix both approaches in the same charm. Also, once #342 is fixed this will be much less painful to look at.

I think that's a little bit the point -- the common exit hook cannot cleanly address this, but we actually do need to do so. There are multiple non-`_configure` places in the charm where we set `BlockedStatus` (in `_generate_command`, `_on_k8s_...`). "Un-setting" it isn't a massive divergence.

The odds of this being needed inside this patch are low, since ingress shouldn't fail/be unavailable long enough for us to want to avoid blocking a charm in `executing` for some timeout (2-5 seconds or whatever), though.
 
> Agreed. I don't particularly like that status comparison pattern and it should be fixed as part of #342.

Why would it be changed there? It also affects this. There seems to be three cases where the charm may not be accessible:

* Bad configuration (we can check with `promtool` if there's an exception raised trying to connect)
* Timeout because it's replaying/`mmap`ing HEAD chunks which haven't been written to the TSDB
* Not reachable due to ingress

`/-/reload` being called to often doesn't have anything to do with the last case, but this patch does.

> Sounds presumptuous at first glance. And what will the charm do with the various exceptions? Just choose one message over another? Maybe, if we have insight, expose a `retry` setting the charm could use?

Using exceptions for flow control is Pythonic, and a narrow custom exception which is caught and sets an appropriate message is much less presumptuous/magic than returning a boolean from `prometheus_server.Prometheus` which may set a meaningless message which will hopefully automagically clear itself the next time something fires and hits the common exit hook.

The exceptions could have a property with a message, or a lookup table in the charm, or whatever. The charm would set a status message to tell the administrator what is happening and why Prometheus may not be reachable which is more useful than a "configuration is invalid" message which may not be true or an `ActiveStatus` when Prometheus is actually not active/reachable.

A `retry` mechanism can only be so useful. For ingress, if we retry for 2-3 seconds, it may be enough. Traefik shouldn't take that long to restart. For `/-/reload`, though, it won't be, and how we address that can certainly be discussed elsewhere (we already "know" that we'll probably have to sidestep it with @manadart's work to trigger async hooks), but we're well enough aware of the "basic" failure cases, and we need a way to flag the administrator about what the actual status is. Since this patch is quite small/contained, and is the logical place to indicate what "goes wrong" when Ingress isn't available yet (timeout or default traefik landing page, I'm not sure which), the initial scaffolding of "raise an exception of the appropriate type" seems that it should go here and not be deferred.

tests/integration/test_external_url.py
```diff
@@ -0,0 +1,99 @@
...
+        # )
+        ops_test.model.add_relation(f"{prometheus_app_name}:ingress", "traefik"),
+        ops_test.model.wait_for_idle(
+            apps=app_names,
+            status="active",
+            wait_for_units=num_units,
+            timeout=600,
+            idle_period=idle_period,
+        ),
+    )
```
dstathis, line 84: Shouldn't these be run sequentially rather than in parallel? When run in parallel, `wait_for_idle` could in theory return before `add_relation` starts running.

tests/integration/test_external_url.py
```diff
@@ -0,0 +1,99 @@
...
+    logger.info("Attmpting to fetch targets from these urls: %s", prom_urls)
+    targets_per_unit = [
+        urllib.request.urlopen(url, None).read().decode("utf8") for url in prom_urls
+    ]
+    assert all('"health":"down"' not in response for response in targets_per_unit)
+
+
+@pytest.mark.abort_on_fail
+async def test_self_scrape_job_is_up_with_config_option_overriding_traefik(ops_test: OpsTest):
+    pass
```
dstathis, line 99: I think you forgot to remove this placeholder.

dstathis (approved):

tests/integration/test_external_url.py
sed-i: Done.

tests/integration/test_external_url.py
sed-i: Done.

src/charm.py
```diff
@@ -159,11 +171,14 @@ def log_level(self):
     @property
     def _default_config(self):
         """Default configuration for the Prometheus workload."""
+        base_path = urlparse(self.external_url).path
+        metrics_path = base_path.rstrip("/") + "/metrics"
```
rbarry82, line 175: This is almost identical to what's in `self_scraping_job` but structured slightly differently. If there a reason not to just use one code variant and make it a property?

tox.ini
```diff
@@ -71,9 +71,13 @@ deps =
     lib: ops
     charm: responses==0.20.0
     charm: httpcore==0.14.7
+    unit: {[testenv:unit]deps}
+    integration: {[testenv:integration]deps}
 commands =
     charm: mypy {[vars]src_path} {posargs}
     lib: mypy --python-version 3.5 {[vars]lib_path} {posargs}
+    unit: mypy {[vars]tst_path}/unit {posargs}
+    integration: mypy {[vars]tst_path}/integration {posargs}
```
rbarry82, line 80: Why are these added back?

sed-i: I think those were removed by mistake. The correct change is
```diff
-envlist = lint, static-{charm,lib,unit,integration}, unit
+envlist = lint, static-{charm,lib}, unit
```
I've done it in this PR here.

As for the envs themselves - I use them! :)

src/charm.py
sed-i: Good catch! Fixed in next commit.

src/charm.py
```diff
@@ -88,25 +88,31 @@ def __init__(self, *args):
...
+        external_url = urlparse(self.external_url)
+
+        # TODO when https://github.com/canonical/traefik-k8s-operator/issues/78 is fixed, change
+        #  the refresh event to self.ingress.on.ready_for_unit.
         self._scraping = MetricsEndpointProvider(
             self,
             relation_name="self-metrics-endpoint",
             jobs=self.self_scraping_job,
+            external_hostname=external_url.netloc,
+            refresh_event=self.on.update_status,  # needed for ingress
```
Abuelodelanada, line 104: Shouldn't we set more refresh_events, like??

```python
            refresh_event=[
                self.ingress_per_unit.on.ready_for_unit,
                self.ingress_per_unit.on.revoked_for_unit,
                self.on.ingress_relation_departed,
                self.on.update_status,
            ],
```

Abuelodelanada (approved):

src/charm.py
sed-i: Fixed, thanks!

Abuelodelanada (approved):

tests/unit/test_endpoint_consumer.py
```diff
@@ -586,3 +590,150 @@ def wildcard_targets(jobs, wildcard_ports):
...
+            "remote-app/1",
+            {
+                "prometheus_scrape_unit_address": "11.11.11.11",
+                "prometheus_scrape_unit_path": external_url_path.substitute(unit=1),
+                "prometheus_scrape_unit_name": "remote-app/1",
+            },
+        )
+
+    def setUp(self):
+        metadata_file = open("metadata.yaml")
```
beliaev-maksim, line 644: this is dependent on the invocation directory

PR #350, opened 2022-09-02
get workload version from the binary rather than the api

dstathis:
## Issue
When using the api to get workload version we rely on `update_status`.

## Solution
Use `/bin/prometheus --version`.

## Testing Instructions
Deploy and check version is set without errors.

## Release Notes
prometheus-k8s no longer relies on update_status to set workload version

### DIFF
--- a/src/charm.py
+++ b/src/charm.py
@@ -7,7 +7,7 @@ import logging
 import os
 import re
 import socket
-from typing import Dict, cast
+from typing import Dict, Optional, cast
 from urllib.parse import urlparse
 
 import yaml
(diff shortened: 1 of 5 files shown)


src/charm.py
```diff
@@ -301,12 +301,20 @@ def _configure(self, _):
...
-    def _update_status(self, event):
-        """Fired intermittently by the Juju agent."""
-        self.unit.set_workload_version(self._prometheus_server.version())
+    def _on_pebble_ready(self, event) -> None:
+        """Pebble ready hook.
+
+        This runs after the the workload container starts.
+        """
+        self._configure(event)
+        if version := self._prometheus_version:
```
rbarry82, line 312: Walrus again

rbarry82 (approved):

rbarry82:
LGTM other than the unit tests.

src/charm.py
dstathis: Pretty sure this code only ever runs on modern python.

PR #351, opened 2022-09-06
Add tls, basic_auth keys to allowlist

### LINKED ISSUE #347: Add extra options to scrape config (`scheme`, `basic_auth`, `tls_config`)
### Enhancement Proposal

I am trying to integrate prometheus-k8s with the Juju [controller charm](https://github.com/juju/juju-controller), to make it easy for COS Lite to observe Juju controllers. This involves implementing the prometheus_scrape library on the controller charm.

The Juju controller's metrics endpoint requires the following `scrape-jobs` config (see e.g. [here](https://juju.is/docs/olm/collecting-juju-metrics)):
```yaml
- job_name: juju
  metrics_path: /introspection/metrics
  scheme: https
  static_configs:
    - targets:
      - '<controller-ip>:17070'
      labels:
        juju_model: controller
        host: controller-0
  basic_auth:
    # Need a user "foo" that has read access to controller
    username: user-foo
    password: 1234
  tls_config:
    ca_file: /path/to/juju-ca.crt
```

However, I notice that the `MetricsEndpointProvider` doesn't accept the fields `scheme`, `basic_auth` or `tls_config` (these are not in `ALLOWED_KEYS` and so will get removed from the passed-in `jobs`). I have tested manually with `prometheus2` and verified that all these keys are necessary.

Hence, I would like to request an implementation of `scheme`, `basic_auth` and `tls_config` in the `jobs` config, as per the yaml above.

sed-i:
## Issue
Fixes #347.

## Solution
Add keys to allowlist.

## Context
NTA

## Testing Instructions
Manually deploy and relate this and the matching PR in [scrape-target](https://github.com/canonical/prometheus-scrape-target-k8s-operator/pull/21).

```bash
juju deploy ./prometheus-scrape-target-k8s_ubuntu-20.04-amd64.charm target
juju deploy ./prometheus-k8s_ubuntu-20.04-amd64.charm prom --resource prometheus-image=ubuntu/prometheus:2.33-22.04_beta

TARGET=$(juju controllers --format json | jq -r '.controllers.uk8s."api-endpoints"[0]')
CERT=$(openssl s_client -connect $TARGET -showcerts </dev/null 2>/dev/null | openssl x509 -outform PEM)
juju config target targets=$TARGET scheme=https metrics_path=/introspection/metrics basic_auth=user-prometheus:prom tls_config_insecure_skip_verify=true tls_config_ca_file=$CERT

juju relate prom target

PROM=$(juju status --format=json | jq -r '.applications.prom.units."prom/0".address')
curl $PROM:9090/api/v1/targets | jq '.data.activeTargets[].health'
```

## Release Notes
Add tls, basic_auth keys to allowlist.

### DIFF
--- a/src/charm.py
+++ b/src/charm.py
@@ -7,7 +7,7 @@ import logging
 import os
 import re
 import socket
-from typing import Dict, Optional, cast
+from typing import Dict, Optional, Tuple, cast
 from urllib.parse import urlparse
 
 import yaml
@@ -260,8 +260,11 @@ class PrometheusCharm(CharmBase):
             self.unit.status = MaintenanceStatus("Configuring Prometheus")
             return
 
-        prometheus_config = self._prometheus_config()
+        prometheus_config, certs = self._prometheus_config()
         container.push(PROMETHEUS_CONFIG, prometheus_config, make_dirs=True)
+        for filename, contents in certs.items():
+            container.push(filename, contents, make_dirs=True)
+
         logger.info("Pushed new configuration")
 
         self._set_alerts(container)
@@ -550,11 +553,11 @@ class PrometheusCharm(CharmBase):
         alerting_config = {"alertmanagers": [{"static_configs": [{"targets": alertmanagers}]}]}
         return alerting_config
 
-    def _prometheus_config(self) -> str:
+    def _prometheus_config(self) -> Tuple[str, dict]:
         """Construct Prometheus configuration.
 
         Returns:
-            Prometheus config file in YAML (string) format.
+            A 2-tuple of prometheus config file in YAML (string) format and a dict of cert per job.
         """
         prometheus_config = {
             "global": self._prometheus_global_config(),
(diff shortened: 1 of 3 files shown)


sed-i:
Hi @barrettj12, I was able to have prom scrape the controller only with `insecure_skip_verify: true` (see testing instructions above). Would you mind testing this on your end?

sed-i:
Seems like the FQDN needs to presents in the common names of the certificate.
Currently it is autogenerated by juju and is limited to:
```
$ openssl s_client -connect 10.152.183.217:17070 -showcerts </dev/null 2>/dev/null | openssl x509 -text -noout -subject - | grep DNS
                DNS:juju-apiserver, DNS:juju-mongodb, DNS:anything, DNS:localhost
```

src/charm.py
```diff
@@ -542,11 +545,11 @@ def _alerting_config(self) -> dict:
         alerting_config = {"alertmanagers": [{"static_configs": [{"targets": alertmanagers}]}]}
         return alerting_config
 
-    def _prometheus_config(self) -> str:
+    def _prometheus_config(self) -> Tuple[str, dict]:
```
rbarry82, line 548: This return type doesn't seem logically consistent or Pythonic. It isn't returning "just" the Prometheus config.

Considering that the only thing done with the returned certs is pushing them to the container, it would probably make more sense to refactor `_prometheus_config` to `_generate_prometheus_config` or similar and handle the writing there, so "get the cert and write it to disk" are logically in the same place, especially since we don't do any other processing or handling of it whatsover

sed-i: So `self._prometheus_config()` is only "converting" reldata to useful stuff, and then both the config yaml and the certs are written together. Kind of separation of concerns. From first glance it seems that this would have a lower spaghetti factor.

I agree it's confusing the "config" on the left hand side means the yaml, but on the right hand side it's the "entire config":

`prometheus_config, certs = self._prometheus_config()`

sed-i: Oh you mean move everything into the same function. Sure, that's probably better.

rbarry82 (approved):

Abuelodelanada (approved):

PR #352, opened 2022-09-07
Less frequent reloads and restarts

### LINKED ISSUE #342: Reload is called unnecessarily

sed-i:
## Issue
Before this change, reload/restart could result in unnecessary down time. 

## Solution
- Only restart if cli args changed
- Only reload if config changed
- Block if config is invalid
- Fix error + status messages
- Drive-by fixes

Fixes #342.

## Context
NTA.

## Testing Instructions
Afaict, the feature is fully captured in the newly added utests. Please review those for completeness!

### DIFF
--- a/src/charm.py
+++ b/src/charm.py
@@ -3,6 +3,7 @@
 # See LICENSE file for licensing details.
 
 """A Juju charm for Prometheus on Kubernetes."""
+import hashlib
 import logging
 import os
 import re
@@ -40,6 +41,7 @@ from lightkube import Client
 from lightkube.core.exceptions import ApiError as LightkubeApiError
 from lightkube.resources.core_v1 import PersistentVolumeClaim, Pod
 from ops.charm import ActionEvent, CharmBase
+from ops.framework import StoredState
 from ops.main import main
 from ops.model import ActiveStatus, BlockedStatus, MaintenanceStatus, WaitingStatus
 from ops.pebble import ChangeError, ExecError, Layer
@@ -49,16 +51,26 @@ from utils import convert_k8s_quantity_to_legacy_binary_gigabytes
 
 PROMETHEUS_CONFIG = "/etc/prometheus/prometheus.yml"
 RULES_DIR = "/etc/prometheus/rules"
-CORRUPT_PROMETHEUS_CONFIG_MESSAGE = "Failed to load Prometheus config"
 
 logger = logging.getLogger(__name__)
+
+
+def sha256(hashable) -> str:
+    """Use instead of the builtin hash() for repeatable values."""
+    if isinstance(hashable, str):
+        hashable = hashable.encode("utf-8")
+    return hashlib.sha256(hashable).hexdigest()
 
 
 class PrometheusCharm(CharmBase):
     """A Juju Charm for Prometheus."""
 
+    _stored = StoredState()
+
     def __init__(self, *args):
         super().__init__(*args)
+
+        self._stored.set_default(config_hash=None, alerts_hash=None)
 
         self._name = "prometheus"
         self._port = 9090
@@ -249,6 +261,15 @@ class PrometheusCharm(CharmBase):
         line arguments). If the Pebble layer has changed then Prometheus
         is restarted.
         """
+        cfg_load_fail = BlockedStatus(
+            "Prometheus failed to reload the configuration (WAL replay or ingress may be in progress); see debug logs"
+        )
+        restart_fail = BlockedStatus(
+            "Prometheus failed to restart (config valid?); see debug logs"
+        )
+        push_fail = BlockedStatus("Failed to push updated config/alert files; see debug logs")
+        early_return_status_list = [cfg_load_fail, restart_fail, push_fail]
+
         container = self.unit.get_container(self._name)
 
         if not self.resources_patch.is_ready():
@@ -260,25 +281,18 @@ class PrometheusCharm(CharmBase):
             self.unit.status = MaintenanceStatus("Configuring Prometheus")
             return
 
-        self._generate_prometheus_config(container)
-        logger.info("Pushed new configuration")
+        try:
+            # Need to reload if config or alerts changed.
+            # (Both functions need to run so cannot use the short-circuiting `or`.)
+            should_reload = any([self._update_config(container), self._set_alerts(container)])
+        except ChangeError as e:
+            logger.error("Failed to push updated config/alert files:  %s", str(e))
+            self.unit.status = push_fail
+            return
 
-        self._set_alerts(container)
+        should_restart = self._update_layer(container)
 
-        current_services = container.get_plan().services
-        new_layer = self._prometheus_layer
-
-        if current_services == new_layer.services:
-            reloaded = self._prometheus_server.reload_configuration()
-            if not reloaded:
-                logger.error("Prometheus failed to reload the configuration")
-                self.unit.status = BlockedStatus(CORRUPT_PROMETHEUS_CONFIG_MESSAGE)
-                return
-
-            logger.info("Prometheus configuration reloaded")
-
-        else:
-            container.add_layer(self._name, new_layer, combine=True)
+        if should_restart:
             try:
                 # If a config is invalid then prometheus would exit immediately.
                 # This would be caught by pebble (default timeout is 30 sec) and a ChangeError
@@ -289,13 +303,19 @@ class PrometheusCharm(CharmBase):
                 logger.error(
                     "Failed to replan; pebble plan: %s; %s", container.get_plan().to_dict(), str(e)
                 )
-                self.unit.status = BlockedStatus(CORRUPT_PROMETHEUS_CONFIG_MESSAGE)
+                self.unit.status = restart_fail
                 return
 
-        if (
-            isinstance(self.unit.status, BlockedStatus)
-            and self.unit.status.message != CORRUPT_PROMETHEUS_CONFIG_MESSAGE
-        ):
+        elif should_reload:
+            reloaded = self._prometheus_server.reload_configuration()
+            if not reloaded:
+                logger.error(cfg_load_fail.message)
+                self.unit.status = cfg_load_fail
+                return
+
+            logger.info("Prometheus configuration reloaded")
+
+        if self.unit.status in early_return_status_list:
             return
 
         self.remote_write_provider.update_endpoint()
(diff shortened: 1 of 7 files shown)


Abuelodelanada (approved):

src/charm.py
```diff
@@ -293,6 +297,15 @@ def _configure(self, _):
                 self.unit.status = BlockedStatus(CORRUPT_PROMETHEUS_CONFIG_MESSAGE)
                 return
 
+        elif reload:
+            reloaded = self._prometheus_server.reload_configuration()
+            if not reloaded:
+                logger.error("Prometheus failed to reload the configuration")
+                self.unit.status = BlockedStatus(CORRUPT_PROMETHEUS_CONFIG_MESSAGE)
```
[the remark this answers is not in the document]
rbarry82: This was one of the conversations had earlier (which we should have moved upstream). This message probably doesn't reflect the actual state of the charm.

It returns a `False` on `ConnectionError`, `ConnectTimeout`, or `ReadTimeout`, but these could be due to other causes (`mmap` replaying, ingress not available, etc). `prometheus_server.reload_configuration()` probably needs to be more explicit about what it's doing and raise an exception which indicates what the actual state is, and then, here, we should use `promtool` (similar to the action) to validate whether or not the config really is corrupt. 

With the code at present, there is no way to know what the "actual" state of the charm is, only that something went wrong when the endpoint was `POST`ed to.

src/charm.py
```diff
@@ -316,18 +329,59 @@ def _on_pebble_ready(self, event) -> None:
                 "Cannot set workload version at this time: could not get Alertmanager version."
             )
 
-    def _set_alerts(self, container):
+    def _update_config(self, container) -> bool:
+        """Pushes new config, if needed.
+
+        Returns a boolean indicating if reload is needed.
```
rbarry82, line 335:
```suggestion
        Returns a boolean indicating if a new configuration was pushed.
```

src/charm.py
```diff
@@ -316,18 +329,59 @@ def _on_pebble_ready(self, event) -> None:
...
+
+    def _set_alerts(self, container) -> bool:
         """Create alert rule files for all Prometheus consumers.
 
         Args:
             container: the Prometheus workload container into which
                 alert rule files need to be created. This container
                 must be in a pebble ready state.
+
+        Returns: A boolean indicating if a reload is needed.
```
rbarry82, line 371:
```suggestion
        Returns: A boolean indicating if new alert rules were pushed.
```

src/charm.py
```diff
@@ -260,26 +273,17 @@ def _configure(self, _):
...
-        if current_services == new_layer.services:
-            reloaded = self._prometheus_server.reload_configuration()
-            if not reloaded:
-                logger.error("Prometheus failed to reload the configuration")
-                self.unit.status = BlockedStatus(CORRUPT_PROMETHEUS_CONFIG_MESSAGE)
-                return
+        try:
+            # Need to reload if config or alerts changed.
+            # (Both functions need to run so cannot use the short-circuiting `or`.)
+            reload = bool(sum([self._update_config(container), self._set_alerts(container)]))
```
rbarry82, line 279:
```suggestion
            should_reload = any([self._update_config(container), self._set_alerts(container)])
```

src/charm.py
```diff
@@ -293,6 +297,15 @@ def _configure(self, _):
                 self.unit.status = BlockedStatus(CORRUPT_PROMETHEUS_CONFIG_MESSAGE)
                 return
 
+        elif reload:
```
rbarry82, line 300:
```suggestion
        elif should_reload:
```

rbarry82 (changes requested):
Can we please at least set a different string if the reload returns an error? Or check whether the config is valid, and if it is, then set a message about failing to connect to the endpoint?

src/charm.py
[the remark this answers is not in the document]
sed-i: Improved the text of the message in the next commit. Further refactoring should probably wait for another PR.

sed-i:
@dstathis here's a summary of StoredState vs pull for determining whether a reload is needed.

|                | StoredState            | Pull                                                                         |
|----------------|------------------------|------------------------------------------------------------------------------|
| Multiple files | Easy to calculate hash | Need to pull and sort all data (config AND alerts) (sort before pushing too) |
| Tinkering      | Manual changes persist | Manual changes disappear every _configure                                    |
| Robustness     | Only push              | Pull (multiple?) then push                                                   |

Given the above I prefer StoredState. Did I miss anything?

dstathis:
> @dstathis here's a summary of StoredState vs pull for determining whether a reload is needed.
> 
> |                | StoredState            | Pull                                                                         |
> |----------------|------------------------|------------------------------------------------------------------------------|
> | Multiple files | Easy to calculate hash | Need to pull (and sort?) all data (config, certs AND alerts) |
> | Tinkering      | Manual changes persist | Manual changes disappear every _configure                                    |
> | Robustness     | Only push              | Pull (multiple files) then push                                                   |
> 
> Given the above I prefer StoredState. Did I miss anything?

I'm not sure I understand why we would lose local changes. I will go ahead and approve though.

dstathis (approved):

rbarry82 (approved):

Abuelodelanada (approved):

src/charm.py
```diff
@@ -249,52 +263,96 @@ def _configure(self, _):
...
+            ),
+            "push_fail": BlockedStatus(
+                "Failed to push updated config/alert files; see debug logs"
+            ),
+            "layer_fail": BlockedStatus("Failed to update prometheus service; see debug logs"),
+            "config_invalid": BlockedStatus("Invalid prometheus configuration; see debug logs"),
+            "validation_fail": BlockedStatus(
+                "Failed to validate prometheus config; see debug logs"
+            ),
+        }
```
rbarry82, line 281: I like this. It seems slightly out of place in a Goldilocks way, though.

If we're going to have constant strings this way, they could be class-level rather than instance level, or just plain ALL-CAPS constants the way the old status was (just to make it clear what they are).

Alternatively, this seems like a great place for a tiny meta-config class/method, where:

```python
if some_bad_thing:
  self.set_unit_status("push_fail") # or even an Enum at the top to make it self.set_unit_status(PUSH_FAIL)
```

Would be a nice abstraction which keeps the list of conditions->statuses in a single place, almost like a `excepthook`

src/charm.py
```diff
@@ -249,52 +263,96 @@ def _configure(self, _):
...
+            self.unit.status = early_return_statuses["layer_fail"]
+            return
 
-        if current_services == new_layer.services:
-            reloaded = self._prometheus_server.reload_configuration()
-            if not reloaded:
-                logger.error("Prometheus failed to reload the configuration")
-                self.unit.status = BlockedStatus(CORRUPT_PROMETHEUS_CONFIG_MESSAGE)
+        try:
+            output, err = self._promtail_check_config()
```
rbarry82, line 314: I worry that this is one of those places where unit testing won't actually approximate what happens.

Specifically, if we add a layer _and_ a config which has a config problem and fails to start, we're going to return just above this with `layer_fail`. Logically, by the time we hit this block at all, it's already succeeded if there are any layer changes, and we don't know whether the config is valid or not (which could be a good/valid reason why the layer won't start)

So, the steps are:

* `generate_config` (and push it), which is not obvious unless you read the method.
* `_update_layer` if the hash differs

It seems that we should do something like:

* `generate_config`, but don't push it yet. Write it to a temporary path, try to check it with promtool. If it fails, set the message. If it passes, move it into the right place and continue. If it fails, do as we are doing now (since checking the config will fail, and `should_restart` won't trigger the "real" restart). The current codepath has a dead zone where an invalid config has been generated and pushed to the container but actually applied yet.
  - Alternatively, generate it, push it, try to restart anyway, and if there's an exception failing to replan, validate the config there and raise this message.

tests/unit/test_charm.py
```diff
@@ -513,12 +509,188 @@ def test_charm_writes_meaningful_alerts_filename_4(self, *_):
...
+
+        # AND workload is restarted
+        self.assertTrue(self.service.is_running())
+        self.assertTrue(restart.called or start.called or replan.called)
+        restart.reset_mock()
+        start.reset_mock()
+        replan.reset_mock()
+
+        # BUT reload is not invoked
+        reload_config.assert_not_called()
```
rbarry82, line 652: Maybe I'm missing something, but aren't all of these essentially checking the same behavior?

src/charm.py
```diff
@@ -408,23 +512,26 @@ def _generate_command(self) -> str:
...
         return " ".join(command)
 
-    def _on_validate_config(self, event: ActionEvent) -> None:
-        """Runs `promtool check config` inside the workload."""
-        container = self.unit.get_container(self._name)
-
-        if not container.can_connect():
-            event.fail("Could not connect to the Prometheus workload!")
-            return
+    def _promtail_check_config(self) -> tuple:
```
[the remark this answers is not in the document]
sed-i: Fixed in next commit.

rbarry82 (commented):
The subtests are a nice addition. Patching of a bunch of `ops.testing._TestingPebbleClient` internals gives me bad feelings, though. Do we really need this huge amount of mock patching?

src/charm.py
sed-i: `_update_layer` only calls `container.add_layer`, so iiuc no restart should take place just yet.

tests/unit/test_charm.py
sed-i: Yes, knowing how the charm is written then these are redundant.
OTOH, feature-wise, the test goes through the "lifecycle" of a config option: set-change-unset. We want to make sure this promise is kept in future changes.

src/charm.py
sed-i: Yep. But this would be a larger refactoring (a good one!) that we should probably discuss separately.

sed-i:
> Patching of a bunch of `ops.testing._TestingPebbleClient` internals gives me bad feelings, though. Do we really need this huge amount of mock patching?

That's the least evil I could come up with at the time. Open for other suggestions.

Abuelodelanada (approved):

Issue #58, opened 2021-06-18
Validate behaviour of cross model relations


balbirthomas:
## Introduction

Even though the Prometheus charm may just work, when used with cross model relations, its behaviour in this case has not yet been tested and validated.

PR #354, opened 2022-09-12
update README.md

beliaev-maksim:
* issue #58 is closed as completed, remove the section from the doc.
* minor fixed

### DIFF
--- a/README.md
+++ b/README.md
@@ -40,9 +40,10 @@ $ juju relate kube-state-metrics prometheus-k8s
 In a similar manner any charm that exposes a scrape target may be
 related to the Prometheus charm.
 
-> Note: At present it is expected that all relations the Prometheus Operator 
-> partakes in are within the same Juju model. For alternative, please set up a 
-> grafana agent in the remote model and use `remote_write` to get metrics into Prometheus.
+> Note: At present it is expected that all relations that Prometheus are 
+> part of are within the same Juju model. For cross-model relations, 
+> please set up a Grafana Agent in the remote model and use `remote_write` 
+> to get the metrics into Prometheus.
 
 ## Dashboard
 
--- a/README.md
+++ b/README.md
@@ -40,7 +40,7 @@ $ juju relate kube-state-metrics prometheus-k8s
 In a similar manner any charm that exposes a scrape target may be
 related to the Prometheus charm.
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -41,15 +40,10 @@ $ juju relate kube-state-metrics prometheus-k8s
 In a similar manner any charm that exposes a scrape target may be
 related to the Prometheus charm.
 
-At present it is expected that all relations the Prometheus Operator
-partakes in are within the same Juju model.
-[Further development](https://github.com/canonical/prometheus-operator/issues/58)
-may extend this to allow cross model scrape targets.
```
simskij, line 47: This is still the case. Prometheus might in some cases (like same-cluster, different models) be able to scrape across model boundaries, but usually what we recommend is for users to set up a grafana agent in the remote model and use `remote_write` to get their metrics into Prometheus.

The reference to the issue can go, however, as we're not expecting to change this behavior.

simskij (changes requested):
Thanks Maksim! I've added a small clarification that I think we should address before merging. 👍🏼

README.md
beliaev-maksim: @simskij 
what is about adding such a note?


> Note: At present it is expected that all relations the Prometheus Operator 
> partakes in are within the same Juju model. For alternative, please set up a 
> grafana agent in the remote model and use `remote_write` to get metrics into Prometheus.

Abuelodelanada (approved):

README.md
simskij: > At present it is expected that all relations that Prometheus are part of are within the same Juju model. For cross-model relations, please set up a Grafana Agent in the remote model and use remote_write to get the metrics into Prometheus.

some minor grammar changes, but all in all - looks good! 👍🏼

beliaev-maksim: done

README.md
```diff
@@ -41,15 +40,15 @@ $ juju relate kube-state-metrics prometheus-k8s
...
+> Note: At present it is expected that all relations that Prometheus are
```
sed-i, line 43:
```suggestion
> Note: At present it is expected that all relations that Prometheus is
```

sed-i (approved):

simskij (approved):

PR #358, opened 2022-09-13
Add architecture decision log [do not merge: experiment]

sed-i:
## Issue
There is currently no centralized place to document architecturally-significant decisions.

## Solution
(Proposal) Use a lean [Y-statements](https://adr.github.io/#sustainable-architectural-decisions)-like format to log the arch. decision records, all in the same file.

In short, the Y-statement is as follows:

> In the context of `<use case/user story u>`, facing `<concern c>` we decided for `<option o>` to achieve `<quality q>`, accepting `<downside d>`.

The long form of it is as follows (extra section “because”):

> In the context of `<use case/user story u>`, facing `<concern c>` we decided for `<option o>` and neglected `<other options>`, to achieve `<system qualities/desired consequences>`, accepting `<downside d/undesired consequences>`, because `<additional rationale>`.

## Release Notes
Add architecture decision log.

### DIFF
--- a/adl.md
+++ b/adl.md
@@ -1,0 +1,19 @@
+# Architecture decision log
+
+## 2022-09-13 Use stored state for keeping track of config hash
+
+In the context of preventing unnecessary reloads
+([#352](https://github.com/canonical/prometheus-k8s-operator/pull/352)), facing the concerns
+outlined in the table below, we decided for using stored state for keeping track of the config
+hash.
+
+Note that the charm's storage backend is the charm (`use_juju_for_storage` is not set), so files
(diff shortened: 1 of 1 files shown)


simskij:
This is something we've discussed prior, but using the terminology ADRs (Log -> Record). While I do like the premise, and have been using them extensively in other projects, I do wonder if we should:
 
  1. adopt the practice, but 
  2. put it with the rest of our specs. 

WDYT?

beliaev-maksim:
great idea but I would go for an explicit file name rather than another acronym

sed-i:
I think we should revisit this when we start working on v1. Closing.

PR #362, opened 2022-09-19
cleanup tests

beliaev-maksim:
* remove patch for UUID and replace with uuid v4 real value
* remove duplicated test. @sed-i, please check it, comments look different, but the code itself is identical

### DIFF
--- a/tests/unit/test_endpoint_aggregator.py
+++ b/tests/unit/test_endpoint_aggregator.py
@@ -3,7 +3,6 @@
 
 import json
 import unittest
-from unittest.mock import patch
 
 from charms.prometheus_k8s.v0.prometheus_scrape import MetricsEndpointAggregator
 from ops.charm import CharmBase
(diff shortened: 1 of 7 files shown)


tests/unit/test_endpoint_aggregator.py
```diff
@@ -78,11 +77,10 @@ def __init__(self, *args, **kwargs):
         )
 
 
-@patch("charms.observability_libs.v0.juju_topology.JujuTopology.is_valid_uuid", lambda *args: True)
 class TestEndpointAggregator(unittest.TestCase):
     def setUp(self):
         self.harness = Harness(EndpointAggregatorCharm, meta=AGGREGATOR_META)
-        self.harness.set_model_info(name="testmodel", uuid="1234567")
+        self.harness.set_model_info(name="testmodel", uuid="12de4fae-06cc-4ceb-9089-567be09fec78")
```
Abuelodelanada, line 83: I like this PR Maksim!
I was thinking... what about generating the UUID every time we run the test? Seems it will be more robust.

Abuelodelanada (approved):

tests/unit/test_endpoint_aggregator.py
beliaev-maksim: then we will need to regenerate model names as well, since they contain part of the UUID.

eg
https://github.com/canonical/prometheus-k8s-operator/blob/07074183ec918d9a8db10f09b1bb7133d7140eff/tests/unit/test_endpoint_aggregator.py#L114

I think logic is clear, we can just trim UUID to the first 7 chars, if I got it right

but before going there, I would like to see what other think

@rbarry82 , @simskij @dstathis

rbarry82: I tend to agree that just trimming is probably enough. That's what `observability_libs.JujuTopology` does anyway, and the added complexity in dynamic creation of the rest of the attributes (model name, etc) isn't worth it from my POV considering that the unit tests will never _truly_ be able to robustly make assertions against Juju anyway. That requires integration tests.

From my side, I'm ok with unit tests which demonstrate that the essential logical flows work with reasonable values, even if they're static/const.

sed-i (approved):
Nice!

beliaev-maksim:
@sed-i 
great, can you merge it then?

PR #363, opened 2022-09-20
add new landing page integration

simskij:
This PR introduces a new relation interface called `catalog` for connecting Prometheus to a landing page. Over this relation, Prometheus provides an application entry, which then will be used by the landing page to display where users may go.

## Context
See https://github.com/canonical/landing-page-k8s-operator/ for the other end of this relation.

## Testing Instructions
- Build a charm file from this PR branch
- Deploy the build output on Juju
- Deploy `landing-page-k8s` from `edge`
- Relate the two charms together
- Visit the landing page on port 80

## Release Notes
Implement the `catalog` interface to allow for integration with `landing-page-k8s`.

### DIFF
--- a/lib/charms/landing_page_k8s/v0/landing_page.py
+++ b/lib/charms/landing_page_k8s/v0/landing_page.py
@@ -1,5 +1,7 @@
 """Charm for providing landing pages to bundles."""
 
+import ipaddress
+import socket
 from ops.framework import BoundEvent, EventBase, EventSource, Object, ObjectEvents
 from ops.charm import CharmBase
 
@@ -7,7 +9,7 @@ import logging
 
 LIBID = "7e5cd3b1e1264c2689f09f772c9af026"
 LIBAPI = 0
-LIBPATCH = 1
+LIBPATCH = 2
 
 DEFAULT_RELATION_NAME = "landing-page"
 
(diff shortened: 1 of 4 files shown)


sed-i (commented):
Very nice!
This suggests adding two new terms: landing page, catalog.
Can we add only 1.5 new terms? E.g.
- catalog + catalog item; or
- index + index entry;
- etc

lib/charms/landing_page_k8s/v0/landing_page.py
```diff
@@ -0,0 +1,142 @@
...
+        return [
+            {
+                "name": relation.data[relation.app].get("name", ""),
+                "url": relation.data[relation.app].get("url", ""),
+                "icon": relation.data[relation.app].get("icon", ""), 
+                "description": relation.data[relation.app].get("description", "")
+            }
+            for relation in self._charm.model.relations[self._relation_name]
+            if relation.app and relation.units
+        ]
```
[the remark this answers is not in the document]
simskij: That is what line 141 does, isn't it? If you think adding name makes a difference, by all means, I can add it :D

sed-i (approved):

src/charm.py
```diff
@@ -116,13 +117,28 @@ def __init__(self, *args):
...
+                self.on.prometheus_pebble_ready,
+                self.on["ingress"].relation_joined,
+            ],
+            item=CatalogueItem(
+                name="Prometheus",
+                icon="chart-line-variant",
+                url=self.external_url,
+                description=(
+                    "Prometheus collects, stores and serves metrics as time series data, "
+                    ",alongside optional key-value pairs called labels."
```
dstathis, line 137: Wont this render as:
"Prometheus collects, stores and serves metrics as time series data, ,alongside optional key-value pairs called labels."

simskij: it will! thanks

simskij, line 137:
```suggestion
                    "alongside optional key-value pairs called labels."
```

PR #366, opened 2022-09-21
make all tests to be independent of invocation folder

beliaev-maksim:
now all the tests could be executed from any folder


sed-i (commented):
Seems like this would break if we reorg the tests a bit.
Just curious what use case prompted this change?

beliaev-maksim:
unittest discover and execution should work from any folder. For example, user will need to execute a particular test while in the `unit` folder. This will not work.

it is a good practice, that tests should be independent of the current working directory

dstathis (approved):

Abuelodelanada (approved):

PR #367, opened 2022-09-21
Resolve duplication

beliaev-maksim:
wrap duplicated code fragments into functions

### DIFF
--- a/tests/unit/test_endpoint_provider.py
+++ b/tests/unit/test_endpoint_provider.py
@@ -574,7 +574,7 @@ class TestAlertRulesWithMultipleRulesPerFile(unittest.TestCase):
     def test_load_multiple_rules_per_file(self):
         """Test official format with multiple alert rules per group in multiple groups."""
         rules_file_dict = {"groups": [self.gen_group(1), self.gen_group(2)]}
-        self._compare_rules(rules_file_dict)
+        self._compare_rules(rules_file_dict, "1", "2")
 
     def _generate_rules_file_dict(self, rules_file_dict):
         sandbox = TempFS("rule_files", auto_clean=True)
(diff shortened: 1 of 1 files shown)


tests/unit/test_endpoint_provider.py
```diff
@@ -278,36 +278,32 @@ def test_provider_unit_sets_fqdn_if_not_address_on_relation_joined(self):
         self.assertEqual(data["prometheus_scrape_unit_address"], "some.host")
         self.assertIn("prometheus_scrape_unit_name", data)
 
-    @patch_network_get()
-    def test_provider_supports_multiple_jobs(self):
+    def _get_data(self, target):
```
sed-i, line 281: Can we generalize this a function?
Seems odd to me that"
- `provider` is hard-coded
- a non-test function is making assertions
- this specifically addresses app data (cf. https://github.com/canonical/operator/issues/834)

On one hand it reduces duplication but on the other hand the tests become less straightforward.

sed-i (commented):
Most changes look good.
I have reservations about `_get_data`.

tests/unit/test_endpoint_provider.py
```diff
@@ -585,25 +574,30 @@ def gen_group(self, name):
     def test_load_multiple_rules_per_file(self):
         """Test official format with multiple alert rules per group in multiple groups."""
         rules_file_dict = {"groups": [self.gen_group(1), self.gen_group(2)]}
+        self._compare_rules(rules_file_dict, "1", "2")
+
+    def _generate_rules_file_dict(self, rules_file_dict):
```
dstathis, line 579: This method seems to take a dict, write it to disk, read it from disk, and return it. What is the difference between this and
```
def generate_rules_file_dict(self, rules_file_dict):
    return rules_file_dict
```

dstathis (commented):
Unfortunately, I think this PR sacrifice a lot of readability for a small amount of code deduplication.

tests/unit/test_endpoint_provider.py
beliaev-maksim: checks that alert rules are properly reading it?

PS
it is not a code that I wrote initially :)

beliaev-maksim: also, I think it will insert juju topology in the resulting dict

Issue #371, opened 2022-09-25
Reload of config during Prometheus startup can cause permanent blocked state.


dstathis:
### Bug Description

When Prometheus is starting, we can get stuck in a permanent blocked state that is impossible to fix.

### To Reproduce

`kubectl delete pod prometheus-0`
repeat until blocked state is reached

### Environment

microk8s

### Relevant log output

```shell
None
```

PR #372, opened 2022-09-25
Fix reload

dstathis:
## Issue
resolves https://github.com/canonical/prometheus-k8s-operator/issues/371

## Solution
Make sure to observe update_status so Prometheus can be cleaned up.

## Testing Instructions
Kill a Prometheus pod and make sure you get a waiting status (or active status) instead of blocked status.

## Release Notes
fixed a bug where prometheus-k8s would go in to blocked status if it took too long to start


dstathis:
Fixes #371

simskij (approved):

PR #373, opened 2022-09-29
Metrics path per unit

sed-i:
## Issue
With #370 implemented, now wildcard targets are expanded to individual jobs.
However, the provider's constructor currently takes a hostname rather than a url, and (ingress) path cannot be inferred.

## Solution
Change the provider constructor to take a url. This is a non-breaking change because prometheus_scrape wasn't published since the `external_hostname` arg was introduced.

## Context
Depends on #370.

## Testing Instructions
Afaict, existing utests and itest provide sufficient coverage.

## Release Notes
Add support for a custom metrics path per unit.

### DIFF
--- a/lib/charms/prometheus_k8s/v0/prometheus_scrape.py
+++ b/lib/charms/prometheus_k8s/v0/prometheus_scrape.py
@@ -324,11 +324,13 @@ import subprocess
 import tempfile
 from pathlib import Path
 from typing import Dict, List, Optional, Tuple, Union
+from urllib.parse import urlparse
 
 import yaml
 from charms.observability_libs.v0.juju_topology import JujuTopology
 from ops.charm import CharmBase, RelationRole
 from ops.framework import BoundEvent, EventBase, EventSource, Object, ObjectEvents
+from ops.model import Relation
 
 # The unique Charmhub library identifier, never change it
 LIBID = "bc84295fef5f4049878f07b131968ee2"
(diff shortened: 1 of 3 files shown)


simskij (approved):

lib/charms/prometheus_k8s/v0/prometheus_scrape.py
```diff
@@ -1191,31 +1195,22 @@ def _static_scrape_config(self, relation) -> list:
...
-        Args:
-            relation: An `ops.model.Relation` object for which the unit name to
-                address mapping is required.
-
-        Returns:
-            A dictionary that maps unit names to unit addresses for
-            the specified relation.
-        """
+    def _relation_hosts(self, relation: Relation) -> Dict[str, Tuple[str, str]]:
+        """Returns a mapping from unit names to (address, path) tuples, for the given relation."""
```
Abuelodelanada, line 1213: I know our linting is no checking this, but just in case some day we decide to check it, we may get:

`D401: First line should be in imperative mood (perhaps 'Return', not 'Returns') [D401]`

Abuelodelanada (approved):

lib/charms/prometheus_k8s/v0/prometheus_scrape.py
rbarry82: That sounds like a bug in the linter, or it's interpreting a single line as a Google-format docstring for some reason. Maybe you should open a bug on the linter that reported it, because "Returns" is the correct English here. I doubt if they're going to pull in NLTK to try to get English right, though ;)

rbarry82 (approved):

lib/charms/prometheus_k8s/v0/prometheus_scrape.py
sed-i: https://github.com/canonical/observability-libs/issues/27

PR #376, opened 2022-10-05
Remotewrite

### LINKED ISSUE #353: Add support for remote writing from one Prometheus to another
### Enhancement Proposal

We need to support use cases where users want to tier their Prometheus deployments to avoid excessive pressure. By adding `send-remote-write` to Prometheus, and remote writing whatever we receive, we've effectively accomplished this.

beliaev-maksim:
## Issue

closes #353

### DIFF
--- a/src/charm.py
+++ b/src/charm.py
@@ -23,7 +23,7 @@ from charms.observability_libs.v0.kubernetes_compute_resources_patch import (
 )
 from charms.observability_libs.v0.kubernetes_service_patch import KubernetesServicePatch
 from charms.prometheus_k8s.v0.prometheus_remote_write import (
-    DEFAULT_RELATION_NAME as DEFAULT_REMOTE_WRITE_RELATION_NAME,
+    DEFAULT_RELATION_NAME as DEFAULT_REMOTE_WRITE_RELATION_NAME, PrometheusRemoteWriteConsumer,
 )
 from charms.prometheus_k8s.v0.prometheus_remote_write import (
     PrometheusRemoteWriteProvider,
@@ -40,7 +40,7 @@ from charms.traefik_k8s.v1.ingress_per_unit import (
 from lightkube import Client
 from lightkube.core.exceptions import ApiError as LightkubeApiError
 from lightkube.resources.core_v1 import PersistentVolumeClaim, Pod
-from ops.charm import ActionEvent, CharmBase
+from ops.charm import ActionEvent, CharmBase, RelationChangedEvent
 from ops.framework import StoredState
 from ops.main import main
 from ops.model import ActiveStatus, BlockedStatus, MaintenanceStatus, WaitingStatus
@@ -111,6 +111,8 @@ class PrometheusCharm(CharmBase):
             endpoint_path=f"{external_url.path}/api/v1/write",
         )
 
+        self._remote_write_consumer = PrometheusRemoteWriteConsumer(self)
+
         self.grafana_source_provider = GrafanaSourceProvider(
             charm=self,
             source_type="prometheus",
@@ -131,6 +133,10 @@ class PrometheusCharm(CharmBase):
         self.framework.observe(self.on.receive_remote_write_relation_created, self._configure)
         self.framework.observe(self.on.receive_remote_write_relation_changed, self._configure)
         self.framework.observe(self.on.receive_remote_write_relation_broken, self._configure)
+        self.framework.observe(
+            self._remote_write_consumer.on.endpoints_changed,
+            self._on_remote_write_endpoints_changed,
+        )
         self.framework.observe(self.metrics_consumer.on.targets_changed, self._configure)
         self.framework.observe(self.alertmanager_consumer.on.cluster_changed, self._configure)
         self.framework.observe(self.resources_patch.on.patch_failed, self._on_k8s_patch_failed)
@@ -251,6 +257,10 @@ class PrometheusCharm(CharmBase):
 
     def _on_k8s_patch_failed(self, event: K8sResourcePatchFailedEvent):
         self.unit.status = BlockedStatus(event.message)
+
+    def _on_remote_write_endpoints_changed(self, event: RelationChangedEvent) -> None:
+        """Event handler for the remote write endpoint changed event."""
+        self._configure(event)
 
     def _configure(self, _):
         """Reconfigure and either reload or restart Prometheus.
--- a/src/charm.py
+++ b/src/charm.py
@@ -694,7 +694,6 @@ class PrometheusCharm(CharmBase):
         Returns:
             list of dicts consisting of the remote write endpoints configuration for Prometheus.
         """
-
         endpoints = self._remote_write_consumer.endpoints
 
         if not endpoints:
--- a/src/charm.py
+++ b/src/charm.py
@@ -150,6 +150,7 @@ class PrometheusCharm(CharmBase):
         self.framework.observe(self.on.receive_remote_write_relation_created, self._configure)
         self.framework.observe(self.on.receive_remote_write_relation_changed, self._configure)
         self.framework.observe(self.on.receive_remote_write_relation_broken, self._configure)
+        self.framework.observe(self.on.send_remote_write_relation_broken, self._configure)
         self.framework.observe(
             self._remote_write_consumer.on.endpoints_changed,
             self._on_remote_write_endpoints_changed,
--- a/src/charm.py
+++ b/src/charm.py
@@ -8,7 +8,7 @@ import logging
 import os
 import re
 import socket
-from typing import Dict, Optional, cast, List
+from typing import Dict, List, Optional, cast
 from urllib.parse import urlparse
 
 import yaml
@@ -24,9 +24,10 @@ from charms.observability_libs.v0.kubernetes_compute_resources_patch import (
 )
 from charms.observability_libs.v0.kubernetes_service_patch import KubernetesServicePatch
 from charms.prometheus_k8s.v0.prometheus_remote_write import (
-    DEFAULT_RELATION_NAME as DEFAULT_REMOTE_WRITE_RELATION_NAME, PrometheusRemoteWriteConsumer,
+    DEFAULT_RELATION_NAME as DEFAULT_REMOTE_WRITE_RELATION_NAME,
 )
 from charms.prometheus_k8s.v0.prometheus_remote_write import (
+    PrometheusRemoteWriteConsumer,
     PrometheusRemoteWriteProvider,
 )
 from charms.prometheus_k8s.v0.prometheus_scrape import (
--- a/src/charm.py
+++ b/src/charm.py
@@ -277,6 +277,7 @@ class PrometheusCharm(CharmBase):
 
     def _on_remote_write_endpoints_changed(self, event: RelationChangedEvent) -> None:
         """Event handler for the remote write endpoint changed event."""
+        logger.debug("Remote write endpoints were changed")
         self._configure(event)
 
     def _configure(self, _):
@@ -696,6 +697,8 @@ class PrometheusCharm(CharmBase):
 
         if not endpoints:
             logger.debug("No remote write endpoints available")
+        else:
+            logger.debug("Remote write endpoints are: %s", endpoints)
 
         return endpoints
 
--- a/src/charm.py
+++ b/src/charm.py
@@ -8,7 +8,7 @@ import logging
 import os
 import re
 import socket
-from typing import Dict, Optional, cast
+from typing import Dict, Optional, cast, List
 from urllib.parse import urlparse
 
 import yaml
@@ -683,6 +683,22 @@ class PrometheusCharm(CharmBase):
         alerting_config = {"alertmanagers": [{"static_configs": [{"targets": alertmanagers}]}]}
         return alerting_config
 
+    def _remote_write_config(self) -> List[Dict[str, str]]:
+        """Construct Prometheus endpoints configuration for remote write.
+
+        See https://prometheus.io/docs/prometheus/latest/configuration/configuration/#remote_write
+
+        Returns:
+            list of dicts consisting of the remote write endpoints configuration for Prometheus.
+        """
+
+        endpoints = self._remote_write_consumer.endpoints
+
+        if not endpoints:
+            logger.debug("No remote write endpoints available")
+
+        return endpoints
+
     def _generate_prometheus_config(self, container) -> bool:
         """Construct Prometheus configuration and write to filesystem.
 
@@ -697,6 +713,9 @@ class PrometheusCharm(CharmBase):
         alerting_config = self._alerting_config()
         if alerting_config:
             prometheus_config["alerting"] = alerting_config
+
+        if remote_write_config := self._remote_write_config():
+            prometheus_config["remote_write"] = remote_write_config
 
         prometheus_config["scrape_configs"].append(self._default_config)  # type: ignore
         certs = {}
--- a/tests/unit/test_charm.py
+++ b/tests/unit/test_charm.py
@@ -10,6 +10,7 @@ from unittest.mock import patch
 
 import ops
 import yaml
+from charms.prometheus_k8s.v0.prometheus_remote_write import DEFAULT_CONSUMER_NAME
 from helpers import k8s_resource_multipatch, prom_multipatch
 from ops.model import ActiveStatus, BlockedStatus, MaintenanceStatus
 from ops.testing import Harness
@@ -231,6 +232,30 @@ class TestCharm(unittest.TestCase):
             if job["job_name"] != "prometheus":
                 self.assertIn("honor_labels", job)
                 self.assertTrue(job["honor_labels"])
+
+    @k8s_resource_multipatch
+    @patch("lightkube.core.client.GenericSyncClient")
+    def test_send_remote_write_endpoints(self, *unused):
+        rel_id = self.harness.add_relation(DEFAULT_CONSUMER_NAME, "prometheus-read")
+        unit_name = "prometheus-read/0"
+        self.harness.add_relation_unit(rel_id, unit_name)
+
+        self.harness.update_relation_data(
+            rel_id,
+            unit_name,
+            {"remote_write": json.dumps({"url": "http://1.1.1.1:9090/api/v1/write"})},
+        )
+
+        container = self.harness.charm.unit.get_container(self.harness.charm._name)
+        config = container.pull(PROMETHEUS_CONFIG)
+        prometheus_scrape_config = yaml.safe_load(config)
+
+        self.assertIn("remote_write", prometheus_scrape_config)
+        self.assertEqual(len(prometheus_scrape_config["remote_write"]), 1)
+        self.assertDictEqual(
+            prometheus_scrape_config["remote_write"][0],
+            {"url": "http://1.1.1.1:9090/api/v1/write"},
+        )
 
     @k8s_resource_multipatch
     @patch("lightkube.core.client.GenericSyncClient")
--- a/metadata.yaml
+++ b/metadata.yaml
@@ -47,6 +47,8 @@ requires:
   ingress:
     interface: ingress_per_unit
     limit: 1
+  send-remote-write:
+    interface: prometheus_remote_write
 
 peers:
   prometheus-peers:


src/charm.py
```diff
@@ -696,6 +709,23 @@ def _alerting_config(self) -> dict:
...
+
+        See https://prometheus.io/docs/prometheus/latest/configuration/configuration/#remote_write
+
+        Returns:
+            list of dicts consisting of the remote write endpoints configuration for Prometheus.
+        """
+        endpoints = self._remote_write_consumer.endpoints
+
+        if not endpoints:
+            logger.debug("No remote write endpoints available")
```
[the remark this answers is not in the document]
beliaev-maksim: Yes, it will be logged without a relation. It is a debug flag, don't think it will really harm

src/charm.py
```diff
@@ -289,6 +297,11 @@ def _on_ingress_revoked(self, event: IngressPerUnitRevokedForUnitEvent):
     def _on_k8s_patch_failed(self, event: K8sResourcePatchFailedEvent):
         self.unit.status = BlockedStatus(event.message)
 
+    def _on_remote_write_endpoints_changed(self, event: RelationChangedEvent) -> None:
+        """Event handler for the remote write endpoint changed event."""
+        logger.debug("Remote write endpoints were changed")
```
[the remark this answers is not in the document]
beliaev-maksim: To my understanding Prometheus should work as usual + send to requested endpoints

@simskij, can you confirm?

src/charm.py
```diff
@@ -120,6 +121,8 @@ def __init__(self, *args):
             endpoint_path=f"{external_url.path}/api/v1/write",
         )
 
+        self._remote_write_consumer = PrometheusRemoteWriteConsumer(self)
```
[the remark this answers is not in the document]
simskij: 1. yes
2. i'm not sure i follow?
3. likely, seems we don't understand each other 😅

src/charm.py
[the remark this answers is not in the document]
simskij: if it can work as normal - excellent. I was under the impression that enabling the remote write sender automatically put it into agent mode, but @beliaev-maksim's tests seem to suggest otherwise.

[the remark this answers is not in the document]
beliaev-maksim: it goes to the agent mode, but it just replicates the data, like normal mode + push to the endpoint.

_Sure, but the amazing part is that, even with Remote Write, Prometheus still uses a pull model to gather metrics from applications, which gives us an understanding of those different failure modes. After that, we batch samples and series and export, replicate (push) data to the Remote Write endpoints, limiting the number of monitoring unknowns that the central point has!_

source: https://prometheus.io/blog/2021/11/16/agent/

src/charm.py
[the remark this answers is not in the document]
sed-i: Seems like this is the target use-case?

```mermaid
graph LR

subgraph COS1
  am1[Alertmanager]
  prom1[Prometheus]
  am1 --- prom1
end

subgraph COS2
  am2[Alertmanager]
  prom2[Prometheus]
  am2 --- prom2
end

subgraph COS[Main COS]
  am[Alertmanager]
  prom[Prometheus]
  prom --- am
end

pd1[PagerDuty] --- am1
pd2[PagerDuty] --- am2
am --- pd[PagerDuty]

prom1 --- |remote write, rules| prom
prom2 --- |remote write, rules| prom
```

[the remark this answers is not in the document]
sed-i: - Agreed. BTW I've been adding such diagrams to [INTEGRATING.md](https://github.com/canonical/prometheus-k8s-operator/blob/main/INTEGRATING.md) under "Deployment scenarios" (new section).
- Can you please confirm that you were able to observe new alerts coming out of COS1/COS2 while doing remote-write to MainCOS?
- I would add an itest that the same alert is firing in all three prometheuses. See `/api/v1/alerts`.

Also, we need to think if this functionality could unintentionally bork the system (and if we care to mitigate), for example:
```mermaid
graph LR
prom1 --- |self metrics| prom2
prom1 --- |remote write| prom2

prom1 --- prom3
prom2 --- prom3
```

beliaev-maksim:
sum up last change on alert rules:
# 1
deployed stack
image

# 2
now COS1 which is related to zinc and COS Main has an alert from Zinc
image

# 3
this is forwarded to COS Main via remote writing to it along with COS1 prometheus alert rules, so, we monitor zinc+cos1 prom:
image

# 4
then relating COS1 prom to alert manager in COS1 confirms that it fires up an alert:
image

# 5
relating COS Main prom to alert manager in COS Main confirms that it fires up an alert for zinc as well:
image

simskij:
bump @sed-i @rbarry82

tests/integration/test_receive_remote_write.py
```diff
@@ -0,0 +1,94 @@
...
+        for rule in prom_rules:
+            if ava_rule := rule.get("rules", {}):
+                if ava_rule[0]["name"] == "AlwaysFiringDueToNumericValue":
+                    assert ava_rule[0]["state"] == "firing"
+                    break
+        else:
+            raise AssertionError("Rule was not fired")
+
+
+async def has_metric(ops_test, query: str, app_name: str) -> bool:
```
sed-i, line 87: Should we move this into helpers?

tests/integration/test_receive_remote_write.py
```diff
@@ -0,0 +1,94 @@
...
+        assert await has_metric(
+            ops_test,
+            f'up{{juju_model="{ops_test.model_name}",juju_application="{avalanche}"}}',
+            app,
+        )
+
+        prom_rules = await get_prometheus_rules(ops_test, app, 0)
+        for rule in prom_rules:
+            if ava_rule := rule.get("rules", {}):
+                if ava_rule[0]["name"] == "AlwaysFiringDueToNumericValue":
```
sed-i, line 80: With the zero are we assuming there is only one alert per group? Might break if alerts get restructured a bit.

tests/integration/test_receive_remote_write.py
```diff
@@ -0,0 +1,94 @@
...
+
+    # check that both Prometheus have avalanche metrics and both fire avalanche alert
+    for app in [prom_read, prom_write]:
+        assert await has_metric(
+            ops_test,
+            f'up{{juju_model="{ops_test.model_name}",juju_application="{avalanche}"}}',
+            app,
+        )
+
+        prom_rules = await get_prometheus_rules(ops_test, app, 0)
```
sed-i, line 77:
```suggestion
        # Note: the following depends on an avalnche alert coming from the avalanche charm
        # https://github.com/canonical/avalanche-k8s-operator/blob/main/src/prometheus_alert_rules
        prom_rules = await get_prometheus_rules(ops_test, app, 0)
```

tests/integration/test_receive_remote_write.py
```diff
@@ -0,0 +1,94 @@
...
+        for rule in prom_rules:
+            if ava_rule := rule.get("rules", {}):
+                if ava_rule[0]["name"] == "AlwaysFiringDueToNumericValue":
+                    assert ava_rule[0]["state"] == "firing"
+                    break
+        else:
+            raise AssertionError("Rule was not fired")
```
sed-i, line 84:
```suggestion
            raise AssertionError(f"The 'AlwaysFiringDueToNumericValue' avalanche alert was not found in prometheus '{app}'")
```

tests/unit/test_charm.py
```diff
@@ -232,6 +233,30 @@ def test_honor_labels_is_always_set_in_scrape_configs(self, *unused):
                 self.assertIn("honor_labels", job)
                 self.assertTrue(job["honor_labels"])
 
+    @k8s_resource_multipatch
+    @patch("lightkube.core.client.GenericSyncClient")
+    def test_send_remote_write_endpoints(self, *unused):
+        rel_id = self.harness.add_relation(DEFAULT_CONSUMER_NAME, "prometheus-read")
+        unit_name = "prometheus-read/0"
```
sed-i, line 240: We should probably stick to the existing terminology: send/recv.

tests/unit/test_charm.py
```diff
@@ -232,6 +233,30 @@ def test_honor_labels_is_always_set_in_scrape_configs(self, *unused):
...
+        container = self.harness.charm.unit.get_container(self.harness.charm._name)
+        config = container.pull(PROMETHEUS_CONFIG)
+        prometheus_scrape_config = yaml.safe_load(config)
+
+        self.assertIn("remote_write", prometheus_scrape_config)
+        self.assertEqual(len(prometheus_scrape_config["remote_write"]), 1)
+        self.assertDictEqual(
+            prometheus_scrape_config["remote_write"][0],
+            {"url": "http://1.1.1.1:9090/api/v1/write"},
+        )
```
sed-i, line 258:
```suggestion
        self.assertEqual(
            prometheus_scrape_config["remote_write"],
            [{"url": "http://1.1.1.1:9090/api/v1/write"}],
        )
```

sed-i (commented):
Great itest!

tests/unit/test_charm.py
beliaev-maksim: in this case will not that be an implicit error "Index out of range" if there is no config?

tests/integration/test_receive_remote_write.py
beliaev-maksim: done

tests/integration/test_receive_remote_write.py
```diff
@@ -0,0 +1,94 @@
...
+    run_promql,
+)
+from pytest_operator.plugin import OpsTest
+
+logger = logging.getLogger(__name__)
+
+avalanche = "avalanche"
+# prometheus that will consume from the app and write to the remote API
+prom_write = "prometheus-write"
+prom_read = "prometheus-read"  # prometheus that provides `/api/v1/write` API endpoint
```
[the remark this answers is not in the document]
beliaev-maksim: done

tests/integration/test_receive_remote_write.py
beliaev-maksim: yep, let me redo it

lib/charms/prometheus_k8s/v0/prometheus_remote_write.py
```diff
@@ -621,6 +628,7 @@ def __init__(
         self._charm = charm
         self._relation_name = relation_name
         self._alert_rules_path = alert_rules_path
+        self.extra_rules_list = []  # type: list
```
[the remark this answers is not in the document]
beliaev-maksim: This is used later in the library and is a public attribute that could be updated to provide extra alerts

tests/integration/test_receive_remote_write.py
```diff
@@ -0,0 +1,98 @@
...
+            app,
+        )
+
+        # Note: the following depends on an avalanche alert coming from the avalanche charm
+        # https://github.com/canonical/avalanche-k8s-operator/blob/main/src/prometheus_alert_rules
+        prom_rules_list = await get_prometheus_rules(ops_test, app, 0)
+        for rules_dict in prom_rules_list:
+            if rules_list := rules_dict.get("rules", []):
+                for rule in rules_list:
+                    if rule["name"] == "AlwaysFiringDueToNumericValue":
```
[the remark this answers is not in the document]
beliaev-maksim: List comprehension with nested structure is not well readable.
Them also `black` will break it to many lines making even worse

+ extra loops that we don't need

tests/integration/test_remote_write_grafana_agent.py
```diff
@@ -10,6 +10,7 @@
     check_prometheus_is_ready,
     get_prometheus_rules,
     get_rules_for,
+    has_metric,
     oci_image,
     run_promql,
```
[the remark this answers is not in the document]
beliaev-maksim: Yes,
It is used in the file

lib/charms/prometheus_k8s/v0/prometheus_remote_write.py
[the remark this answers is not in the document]
sed-i: Yes, the variable is later appended but within the lib it is always blank.
I'm no sure about (ab)using it as a public interface for providing extra alerts.
Currently we do not use this pattern anywhere.
Any particular reason you prefer it over a regular constructor arg?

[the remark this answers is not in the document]
beliaev-maksim: another thing we can do is either
* add an argument to `reload_alerts` to consume extra alert rules `(def reload_alerts(self, *, extra_rules: list = None)`
* add a brand new function that does reload with extra alerts

that will also work without abusing attributes

I, personally, would go then with extra argument to `reload_alerts`. Let me know with which option you are more comfortable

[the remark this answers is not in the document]
sed-i: > Not sure about doing this, we might fall into the trap that during class init you do not have alerts and we will not reload alert rules which will lead to the empty databag

I think that should be ok: if alerts suddenly show up on reldata, it means the charm re-init'ed due to some relation-changed, so _this time_, the `extra_rules` arg passed to `PrometheusRemoteWriteConsumer` _would_ have alerts. You will need to init metrics_consumer before remote write though.

> add an argument to `reload_alerts`

Atm, I do not understand why that method is needed.
If we can solve the `extra_rules` use-case fully with a constructor arg then I think we should.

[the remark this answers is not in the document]
rbarry82: I agree that a setter method would be more obvious (and could call `reload_rules` anyway). The public method cannot be solved via a constructor arg, though. It's becoming increasingly obvious that this pattern cannot solve everything, and this case is "Grafana agent receives alert rules from another charm, then needs to send them across". This is after the constructor has been invoked, and it's a different relation in a different library entirely. It needs to stay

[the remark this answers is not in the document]
sed-i: @rbarry82 could you please elaborate that scenario?
I.e. how we'd receive new alert rules without a relation event causing a re-init.

src/charm.py
```diff
@@ -165,11 +170,26 @@ def __init__(self, *args):
...
         self.framework.observe(self.alertmanager_consumer.on.cluster_changed, self._configure)
         self.framework.observe(self.resources_patch.on.patch_failed, self._on_k8s_patch_failed)
         self.framework.observe(self.on.validate_configuration_action, self._on_validate_config)
 
+    def _parse_metric_consumer_alerts(self) -> list:
+        """Parse alert rules from metrics consumer."""
+        alerts = self.metrics_consumer.alerts()
+        consumer_rules_list = []
+        for topology_identifier, alert_rule_groups in alerts.items():
+            consumer_rules_list.extend(alert_rule_groups.get("groups", []))
```
sed-i, line 188: This conversion seems like it could be a library concern.
Would it make sense to change the definition of the `extra_alert_rules` so it readily takes the output of `.alerts()`?

lib/charms/prometheus_k8s/v0/prometheus_remote_write.py
```diff
@@ -671,6 +684,12 @@ def _push_alerts_to_relation_databag(self, relation: Relation) -> None:
 
         alert_rules_as_dict = alert_rules.as_dict()
 
+        if self.extra_alert_rules:
+            if alert_rules_as_dict:
+                alert_rules_as_dict["groups"].extend(self.extra_alert_rules)
+            elif self.extra_alert_rules:
+                alert_rules_as_dict["groups"] = self.extra_alert_rules
```
sed-i, line 691: Could this be simplified a bit without breaking existing contracts?
Maybe this would work?
```python
alert_rules_as_dict = {"groups": alert_rules_as_dict.get("groups", []) + self.extra_alert_rules}
```

beliaev-maksim: then `alert_rules_as_dict` always will be True, causing next condition in the code to fire and fill relation databag with confusing data

src/charm.py
beliaev-maksim: then we limit it to the functionality of this PR only. While there might be some unforeseeable use cases where use may want just to feed with alerts directly. 
or?

sed-i: I think in general we try to do transactions with standard data structures - the prom alert rule format.
I imagine this would be most future proof. For example, if they add another top-level key to the yaml besides "groups".

lib/charms/prometheus_k8s/v0/prometheus_remote_write.py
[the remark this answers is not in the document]
rbarry82: It's [right here](https://github.com/canonical/grafana-agent-k8s-operator/blob/main/src/charm.py#L85). With the "user" [here](https://github.com/canonical/grafana-agent-k8s-operator/blob/main/src/charm.py#L138). 

* Some Loki/Prometheus related to Grafana Agent sends updated alert rules to Grafana Agent.
* Grafana Agent has already initialized `RemoteWriteConsumer`, and the paths (which do not contain the new alert rules) have been read.
* It then writes the new rules to disk, and needs a way to tell `RemoteWriteConsumer` to re-read the alert rule paths so they can be appropriately forwarded to Prom/Loki.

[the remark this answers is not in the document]
sed-i: Ah yes. Then seems like `reload_alerts` with `extra_rules` arg is a necessity.
Thanks for the details @rbarry82!

src/charm.py
beliaev-maksim: done, now reload_alerts can consume alerts as from `metrics_consumer`

lib/charms/prometheus_k8s/v0/prometheus_remote_write.py
```diff
@@ -671,12 +685,23 @@ def _push_alerts_to_relation_databag(self, relation: Relation) -> None:
...
+
         if alert_rules_as_dict:
             relation.data[self._charm.app]["alert_rules"] = json.dumps(alert_rules_as_dict)
 
-    def reload_alerts(self) -> None:
+    def reload_alerts(self, *, extra_alerts: Optional[dict] = None) -> None:
         """Reload alert rules from disk and push to relation data."""
-        self._push_alerts_to_all_relation_databags(None)
+        extra_alerts_list = []
+        if extra_alerts:
```
sed-i, line 700: Based on the function name/doc it seems like a reload should occur with or without `extra_alerts` present.

If that is not the case then we could simplify indentation by using a guard:
```python
if not extra_alerts:
    return
```

lib/charms/prometheus_k8s/v0/prometheus_remote_write.py
```diff
@@ -23,7 +23,14 @@
...
+from ops.charm import (
+    CharmBase,
+    HookEvent,
+    RelationBrokenEvent,
+    RelationEvent,
+    RelationMeta,
+    RelationRole,
+)
 from ops.framework import EventBase, EventSource, Object, ObjectEvents
 from ops.model import Relation
```
sed-i, line 35: Bump LIBPATCH?

sed-i (approved):

lib/charms/prometheus_k8s/v0/prometheus_remote_write.py
beliaev-maksim: in this case there are interal functions that re-read from disk. Don't think we should exit.
That also will break back-compatibility

lib/charms/prometheus_k8s/v0/prometheus_remote_write.py
beliaev-maksim: let's do in another PR once ready to release. Do not want to rerun CI only for this :D

lib/charms/prometheus_k8s/v0/prometheus_remote_write.py
```diff
@@ -671,12 +685,23 @@ def _push_alerts_to_relation_databag(self, relation: Relation) -> None:
 
         alert_rules_as_dict = alert_rules.as_dict()
 
+        if extra_alerts_list:
+            if alert_rules_as_dict:
```
rbarry82, line 689: Did we decide against something like:

```python
alert_rules_as_dict["groups"] = alert_rules_as_dict.get("groups", []) + extra_alerts_list
```

This seems to fulfill all of the requirements without the conditionals.

lib/charms/prometheus_k8s/v0/prometheus_remote_write.py
```diff
@@ -658,11 +668,15 @@ def _handle_endpoints_changed(self, event: RelationEvent) -> None:
     def _push_alerts_on_relation_joined(self, event: RelationEvent) -> None:
         self._push_alerts_to_relation_databag(event.relation)
 
-    def _push_alerts_to_all_relation_databags(self, _: Optional[HookEvent]) -> None:
+    def _push_alerts_to_all_relation_databags(
+        self, _: Optional[HookEvent], *, extra_alerts_list: Optional[list] = None
+    ) -> None:
```
rbarry82, line 673: I mean, in general, I think this whole patch is ok.

Conceptually, though, `_push_alerts_to_relation_databags` also triggers on `leader-elected`, `upgrade-charm`, and `relation-joined`. These do not trigger `src.PrometheusCharm._push_alerts_to_remote_write`.

It seems entirely possible for some other event (a new leader election after scaling the application, refreshing the charm, joining a new consumer)  to then invoke `_push_alerts_to_relation_databag` _without_ anything in `extra_alerts_list`, and they will silently disappear. I would suspect this could even be verified in `test_receive_remote_write` with an additional testcase which performed one of these (refreshing the charm, for example, since there are... problem with scaling in integration tests right now).

We really don't always love `StoredState`, or forcing the existence of a peer relation, but those are about the only two good mechanisms of persistence between events with dynamic data this way.

Alternatively, we did something for Kubeflow endpoint discovery where `extra_alerts_list` could be something like `lookaside_alerts_callable`, set in the constructor, and that callable (which could as easily be `PrometheusCharm.metrics_consumer.alerts`) is invoked if it exists and is callable, which provides kind of an abstraction which doesn't _require_ users to break through encapsulation by directly twiddling `metrics_consumer` from here, but presents the option to do so. Or nothing. Or a lambda. Or whatever.

lib/charms/prometheus_k8s/v0/prometheus_remote_write.py
sed-i: I don't follow. Afaict, current code is equivalent to having
```python
if not extra_alerts:
    return
```
at the top.

lib/charms/prometheus_k8s/v0/prometheus_remote_write.py
beliaev-maksim: in the case if `alert_rules_as_dict` and `extra_alerts_list` we current workflow the relation databag will be empty. We will not create `alert_rules` key.

While with proposed implementation databag key will always exist, but will be in some cases empty

lib/charms/prometheus_k8s/v0/prometheus_remote_write.py
```diff
@@ -595,6 +595,8 @@ def __init__(
         charm: CharmBase,
         relation_name: str = DEFAULT_CONSUMER_NAME,
         alert_rules_path: str = DEFAULT_ALERT_RULES_RELATIVE_PATH,
+        *,
+        extra_alerts_callable: Optional[Callable] = None,
```
sed-i, line 599: We should probably mention motivation+usage of this arg somewhere in the docstring.

I'd also add typing. I think it's
```python
extra_alerts_callable: Optional[Callable[[], dict]] = None,
```
?

lib/charms/prometheus_k8s/v0/prometheus_remote_write.py
```diff
@@ -681,9 +684,37 @@ def _push_alerts_to_relation_databag(self, relation: Relation) -> None:
...
             relation.data[self._charm.app]["alert_rules"] = json.dumps(alert_rules_as_dict)
 
+    def _parse_and_add_extra_alerts(self, alert_rules_as_dict: dict) -> dict:
+        """Parse alerts from `self.extra_alerts_callable`.
+
+        Call `self.extra_alerts_callable`, parse alert groups and
+        extend (mutate) `alert_rules_as_dict` with new alerts.
+
+        Returns:
+            (dict) Modified `alert_rules_as_dict`. Return is done only for explicit mutation.
```
sed-i, line 699:
```suggestion
            The original `alert_rules_as_dict`, augmented with the return value of `self.extra_alerts_callable`.
```

lib/charms/prometheus_k8s/v0/prometheus_remote_write.py
```diff
@@ -681,9 +684,37 @@ def _push_alerts_to_relation_databag(self, relation: Relation) -> None:
...
+        """
+        if callable(self.extra_alerts_callable):
+            extra_alerts_list = []
+            extra_alerts = self.extra_alerts_callable()
+            if extra_alerts:
+                for topology_identifier, alert_rule_groups in extra_alerts.items():
+                    extra_alerts_list.extend(alert_rule_groups.get("groups", []))
+
+                if alert_rules_as_dict:
+                    alert_rules_as_dict["groups"].extend(extra_alerts_list)
```
sed-i, line 709: This is an in-place modification ("by-ref"). Should probably make a copy first just in case? Or make it clear it's modifying in-place (i.e. do not return anything)?

lib/charms/prometheus_k8s/v0/prometheus_remote_write.py
```diff
@@ -681,9 +684,37 @@ def _push_alerts_to_relation_databag(self, relation: Relation) -> None:
 
         alert_rules_as_dict = alert_rules.as_dict()
 
+        alert_rules_as_dict = self._parse_and_add_extra_alerts(alert_rules_as_dict)
+
         if alert_rules_as_dict:
             relation.data[self._charm.app]["alert_rules"] = json.dumps(alert_rules_as_dict)
 
+    def _parse_and_add_extra_alerts(self, alert_rules_as_dict: dict) -> dict:
```
sed-i, line 692: - Should we refactor this into a classmethod?
- Tests would be nice.

lib/charms/prometheus_k8s/v0/prometheus_remote_write.py
```diff
@@ -681,9 +684,37 @@ def _push_alerts_to_relation_databag(self, relation: Relation) -> None:
...
+                for topology_identifier, alert_rule_groups in extra_alerts.items():
+                    extra_alerts_list.extend(alert_rule_groups.get("groups", []))
+
+                if alert_rules_as_dict:
+                    alert_rules_as_dict["groups"].extend(extra_alerts_list)
+                else:
+                    alert_rules_as_dict["groups"] = extra_alerts_list
+
+                logger.debug(
+                    "Extra rules were pushed to remote write endpoint. %s", extra_alerts_list
```
sed-i, line 714: Not sure if we want to dump all this, even if it's just debug. Perhaps:
```suggestion
                    "%s extra rules were pushed to remote write endpoint", len(extra_alerts)
```
?

src/charm.py
```diff
@@ -131,6 +132,10 @@ def __init__(self, *args):
             endpoint_path=f"{external_url.path}/api/v1/write",
         )
 
+        self._remote_write_consumer = PrometheusRemoteWriteConsumer(
+            self, extra_alerts_callable=self.metrics_consumer.alerts
+        )
```
sed-i, line 137: I'm a bit concerned about behaviour in unintended scenarios:
- What happens to the `self.remote_write_provider` of a middle-prom in a tiered scenario? Should we have a config option to switch between modes?
- Do we need to include here alerts coming from `self.remote_write_provider`?

lib/charms/prometheus_k8s/v0/prometheus_remote_write.py
rbarry82: Why would it be a classmethod? It's not a factory returning an instance of the class. It could be static, but so could a lot of things in the codebase.

lib/charms/prometheus_k8s/v0/prometheus_remote_write.py
rbarry82: "Augmented with" is very tortured for non-native speakers compared to "with the return value of `PrometheusRemoteWriteConsumer.extra_alerts_callable` merged into the dict" or similar. 

"Augmented with" (even as a native speaker) seems to imply that the actual contents of the existing `alert_rules_as_dict` somehow may have extra fields added (ala label injection), which this is not doing.

lib/charms/prometheus_k8s/v0/prometheus_remote_write.py
```diff
@@ -681,9 +684,37 @@ def _push_alerts_to_relation_databag(self, relation: Relation) -> None:
...
+
+        Returns:
+            (dict) Modified `alert_rules_as_dict`. Return is done only for explicit mutation.
+        """
+        if callable(self.extra_alerts_callable):
+            extra_alerts_list = []
+            extra_alerts = self.extra_alerts_callable()
+            if extra_alerts:
+                for topology_identifier, alert_rule_groups in extra_alerts.items():
+                    extra_alerts_list.extend(alert_rule_groups.get("groups", []))
```
rbarry82, line 706: This still seems kind of unnecessary.

We could do:

```python
for alert_rule_groups in extra_alerts.values():
  extra_alerts_list.extend(...)
```

It could also be a comprehension, but even I'll admit that would be hard to reason about if it had to be extended/changed later.

lib/charms/prometheus_k8s/v0/prometheus_remote_write.py
rbarry82: I don't know why we'd make a copy "just in case" which we'll just garbage collect later. It's not performance sensitive, but an extra stack allocation for no reason probably isn't worth it.

On the other hand, I would still say that we could just move this up and take:

There are a couple of ways in which all of this could be rearranged to be slightly cleaner, but I'm not going to stand on ceremony about nits.

lib/charms/prometheus_k8s/v0/prometheus_remote_write.py
rbarry82: :+1:

src/charm.py
rbarry82: Is having a middle prometheus in a tiered scenario even a valid use case right now? Grafana Agent sort of does this (scrape one and push to the other), remote write provider alerts are already handled via `on.receive_remote_write`, too.

In particular, the "tiered prom" use case should not be recommended. Grafana Agent does this, or remote writing with Mimir eventually. Bad Things ™️ happen to Prometheus once it starts remote writing instead of something else scraping it (via `/federate` or other). Primarily that it starts sort-of discarding alerts. Grafana Agent sidesteps that.

rbarry82 (approved):

lib/charms/prometheus_k8s/v0/prometheus_remote_write.py
beliaev-maksim: yes, this could be static. However, what is the scenario where you will use it outside of instance?

lib/charms/prometheus_k8s/v0/prometheus_remote_write.py
beliaev-maksim: why `len(extra_alerts)` ?
this is a dict and number of keys will represent number of groups and not number of rules, or?

lib/charms/prometheus_k8s/v0/prometheus_remote_write.py
beliaev-maksim: valid point, thanks!

lib/charms/prometheus_k8s/v0/prometheus_remote_write.py
beliaev-maksim: fixed including suggestion from Ryan

lib/charms/prometheus_k8s/v0/prometheus_remote_write.py
beliaev-maksim: I think we need to specify `Any` as well. Done

lib/charms/prometheus_k8s/v0/prometheus_remote_write.py
beliaev-maksim: probably?

```python
logger.debug(
                    "%s extra rules were pushed to remote write endpoint. %s", len(extra_alerts_list)
                )
```

lib/charms/prometheus_k8s/v0/prometheus_remote_write.py
beliaev-maksim: fixed

lib/charms/prometheus_k8s/v0/prometheus_remote_write.py
beliaev-maksim: fixed

lib/charms/prometheus_k8s/v0/prometheus_remote_write.py
beliaev-maksim: function is refactored to be cleaner. 

However, without a copy. Importing deepcopy, etc does not make a value here. It is not a loop but one time extension, so we do not reference mutable item several times

lib/charms/prometheus_k8s/v0/prometheus_remote_write.py
sed-i: It's not a big deal. As Ryan says we have a lot of this in the codebase.
I meant to better convey intent: is this better as a free function or static?

(In the past we had a debate about staticmethod vs classmethod and after being provided with a [reference](https://mail.python.org/pipermail/python-ideas/2016-July/041189.html) for Guido saying staticmethod was a mistake we said internally we'd always do classmethod.)

src/charm.py
sed-i: By "middle prom" I was referring to either COS1::Prometheus or COS2::Prometheus in the deployment scenario:
What happens if, in addition to alertmanager, someone relates to prom over remote-write?
In the POV of the local COS cluster, they probably expect their prom to be able to intake remote write.

```mermaid
graph LR
subgraph COS1
  am1[Alertmanager]
  prom1[Prometheus]
  am1 --- prom1
end
subgraph COS2
  am2[Alertmanager]
  prom2[Prometheus]
  am2 --- prom2
end
subgraph COS[Main COS]
  am[Alertmanager]
  prom[Prometheus]
  prom --- am
end
pd1[PagerDuty] --- am1
pd2[PagerDuty] --- am2
am --- pd[PagerDuty]
prom1 --- |remote write, rules| prom
prom2 --- |remote write, rules| prom
```

beliaev-maksim: do you mean like this?
I am a bit lost :)

```mermaid
graph LR
subgraph COS1
  am1[Alertmanager]
  prom1[Prometheus]
  am1 --- prom1
end
subgraph COS1.2
  am1.2[Alertmanager]
  prom1.2[Prometheus]
  am1.2 --- prom1.2
end
subgraph COS2
  am2[Alertmanager]
  prom2[Prometheus]
  am2 --- prom2
end
subgraph COS[Main COS]
  am[Alertmanager]
  prom[Prometheus]
  prom --- am
end
pd1[PagerDuty] --- am1
pd2[PagerDuty] --- am2
am --- pd[PagerDuty]
prom1 --- |remote write, rules| am1.2
prom1.2 --- |remote write, rules| prom
prom2 --- |remote write, rules| prom
```

lib/charms/prometheus_k8s/v0/prometheus_remote_write.py
beliaev-maksim: how I understand static method is that `it is a function that can work without a class instance and attached to a class just to show where it logically belongs to`

in our case, yes, we can rework it to static method or even to a separate function. But what the use case will be?

src/charm.py
sed-i: Like this:

```mermaid
graph LR
subgraph COS1
  am1[Alertmanager]
  prom1[Prometheus]
  am1 --- prom1
end

subgraph COS[Main COS]
  am[Alertmanager]
  prom[Prometheus]
  prom --- am
end
pd1[PagerDuty] --- am1
app1 ---|scrape| prom1
app2 ---|remote-write| prom1

am --- pd[PagerDuty]
prom1 --- |remote write, rules| prom
```

lib/charms/prometheus_k8s/v0/prometheus_remote_write.py
sed-i: Ideally I'd like to have completely separate code for manipulating those data structures.
Augmenting a `groups` yaml is outside the responsibility of the charm class, imo.
It would then also be easier to test.
But this would probably be a big refactoring and outside the scope of this PR.
Resolving.

rbarry82: Even if Guido thinks it was a mistake, plenty of others have found good uses for it with `@staticmethod` members on classes. It really doesn't make sense on a private method though.

`@classmethod` is really useful for factories, or accessing class attributes from subclasses (or other). This code isn't returning an instance of the class, and it isn't accessing class-level variables (which is only `on` anyway, and actually requires a "real" instance due to the way OF dispatches things to object handles), so it doesn't have a good reason to be a classmethod regardless.

src/charm.py
rbarry82: No, this is exactly the case we do not want. Grafana Agent does this in a much better way. Remote write does not function this way without something like this patch. Prometheus evaluates alert rules itself, and speaks directly to Alertmanager. The entire experience is unintuitive and terrible. 

The default expectation from Prometheus is that, in your scenario, COS1 Prometheus will evaluate all rules it knows about, write the evaluation status to disk (which is not part of the WAL AFAIK and won't be sent across remote write -- it'll directly trigger any Alertmanagers it knows about). 

As a "general" expectation, the "Main COS" in a remote-write scenario would not receive any alert rules at all, and would need "fresh" ones (or recording rules could be sent across remote write, because those are in the WAL; or if it's Thanos/Cortex/Mimir, it can use a Ruler to ask Prometheus what's firing and go from there).

As a "general" expectation, the Prometheus sending remote write would not be connected to any Alertmanagers at all (because those Alertmanagers are isolated and can't dedupe, so whomever gets paged will get the same thing multiple times). It's handling evaluation only, and the "Main COS" would do all of the "actual" talking to whatever API the alerts are sent from.

I mentioned that Grafana Agent (or `/federate` in the "old days") is a common intermediary because it can relabel to adjust the granularity and mitigate this somewhat (so at least the evaluation may be different). If we're going with a design such as "alert rules are forwarded when Prom is related over remote_write", adding toggles to complicate this relation further is very strictly worse and higher complexity than using a small integrator charm for "remote write with alerts" or "remote write without alerts", depending on preference, or doing it for Grafana agent, or... whatever.

sed-i:
More diagrams for future reference:

### COS duplication (hypothetical; symmetric to prometheus)
```mermaid
graph LR
subgraph COS1
  am1[Alertmanager]
  prom1[Prometheus]
  graf1[Grafana]
  loki1[Loki]
  am1 --- prom1
end

subgraph COS[Main COS]
  am[Alertmanager]
  prom[Prometheus]
  graf[Grafana]
  loki[Loki]
  prom --- am
end
pd1[PagerDuty] --- am1

am --- pd[PagerDuty]
prom1 --- |remote write, rules| prom
loki1 -.- |logs, rules - not impl'd| loki
graf1 -.- |dashboards - not impl'd| graf
```

### COS duplication (hypothetical; via grafana-agent)

```mermaid
graph LR
subgraph COS1
  am1[Alertmanager]
  prom1[Prometheus]
  graf1[Grafana]
  loki1[Loki]
  am1 --- prom1

  agent1[Grafana Agent]
  prom1 ---|metrics, rules| agent1
  graf1 -.-|dashboards - not impl'd| agent1
  loki1 ---|logs, rules| agent1
end

subgraph COS[Main COS]
  am[Alertmanager]
  prom[Prometheus]
  graf[Grafana]
  loki[Loki]
  prom --- am
end
pd1[PagerDuty] --- am1

am --- pd[PagerDuty]

agent1 ---|metrics, rules| prom
agent1 -.-|dashboards - not impl'd| graf
agent1 ---|logs, rules| loki
```

PR #381, opened 2022-10-07
Add auto fetch-lib action

sed-i:
## Issue
Need to keep charm libs up to date.

## Solution
Periodic action for auto-opening PRs.

Refs:
- https://docs.github.com/en/actions/quickstart
- https://github.com/peter-evans/create-pull-request/blob/main/docs/examples.md
- https://github.com/canonical/observability/blob/main/.github/workflows/_quality-checks.yaml
- https://github.com/canonical/charming-actions/blob/main/dist/check-libraries/index.js

## Release Notes
Add auto fetch-lib action


sed-i:
Apparently @ca-scribner is ahead of the game!
https://github.com/canonical/charmed-kubeflow-workflows/blob/main/.github/workflows/update-libs.yaml

.github/workflows/fetch-lib.yml
```diff
@@ -0,0 +1,40 @@
...
+        with:
+            commit-message: Update dependencies (fetch-lib)
+            title: Update dependencies (fetch-lib)
+            body: |
+              - Dependency updates
+  
+              Auto-generated by [create-pull-request][1]
+  
+              [1]: https://github.com/peter-evans/create-pull-request
+            branch: fetch-lib
```
ca-scribner, line 40: As is, this action is allowed to create PRs but these PRs will not trigger the repo's CI.  This is because the action uses the `GITHUB_TOKEN` to create the pull request, but that token does not have permission to trigger workflow events.  There's more documentation [here].

If you want to trigger the repo's CI (which I'm guessing you do), you need to create and pass a PAT that has `repo` access on the repo where the PR is being created.  So something like:

```suggestion
            branch: fetch-lib
            token: "${{ secrets.SOME_PAT_TOKEN }}"
```

.github/workflows/fetch-lib.yml
```diff
@@ -0,0 +1,40 @@
+name: Update Dependencies (fetch-lib)
+
+on:
+  schedule:
+    - cron:  '0 10 * * 1'
```
PietroPasotti, line 5: just asking: is there a way to put a limit on the number of retries? Like, if this action fails X times in a row and there have been no code changes, stop trying?

Asking mostly out of curiosity, because one such script sent me 20 emails over the weekend at some point.

PietroPasotti (approved):

.github/workflows/fetch-lib.yml
PietroPasotti: Also, should this not be in the master COS CI?

.github/workflows/fetch-lib.yml
sed-i: @jnsgruk that looks great. I'm not clear on the `tibdex/github-app-token` details but would like to give it a try.

sed-i: Could this be part of template-operator?

sed-i:
[Zinc's udpate-libs](https://github.com/jnsgruk/zinc-k8s-operator/blob/main/.github/workflows/update-libs.yaml) fully addresses the need. Closing.

PR #382, opened 2022-10-10
add the juju_topology lib requirement to the docs

lucabello:
## Issue

Implementing a Prometheus relation by using this library also requires the developer to fetch the [charms.observability_libs.v0.juju_topology](https://github.com/canonical/observability-libs/blob/main/lib/charms/observability_libs/v0/juju_topology.py) library.

The library is required [here](https://github.com/canonical/prometheus-k8s-operator/blob/main/lib/charms/prometheus_k8s/v0/prometheus_scrape.py#L331).

## Solution
Documentation update to communicate to the library user the need to fetch that library to be able to use this one.

### DIFF
--- a/lib/charms/prometheus_k8s/v0/prometheus_scrape.py
+++ b/lib/charms/prometheus_k8s/v0/prometheus_scrape.py
@@ -1,6 +1,6 @@
 # Copyright 2021 Canonical Ltd.
 # See LICENSE file for licensing details.
-"""## Overview.
+"""## Overview
 
 This document explains how to integrate with the Prometheus charm
 for the purpose of providing a metrics endpoint to Prometheus. It
@@ -10,6 +10,11 @@ currently integrated charms. Finally this document is the
 authoritative reference on the structure of relation data that is
 shared between Prometheus charms and any other charm that intends to
 provide a scrape target for Prometheus.
+
+## Requirements
+
+Using this library requires to fetch the juju_topology library from the
+[observability-libs]((https://github.com/canonical/observability-libs/blob/main/lib/charms/observability_libs/v0/juju_topology.py).
 
 ## Provider Library Usage
 


lib/charms/prometheus_k8s/v0/prometheus_scrape.py
```diff
@@ -11,6 +11,11 @@
 shared between Prometheus charms and any other charm that intends to
 provide a scrape target for Prometheus.
 
+## Requirements
+
+Using this library requires to fetch the juju_topology library from the
+[observability-libs]((https://github.com/canonical/observability-libs/blob/main/lib/charms/observability_libs/v0/juju_topology.py).
```
dstathis, line 17:
```suggestion
Using this library requires you to fetch the juju_topology library from
[observability-libs]((https://github.com/canonical/observability-libs/blob/main/lib/charms/observability_libs/v0/juju_topology.py).

`charmcraft fetch-lib charms.observability_libs.v0.juju_topology`
```

dstathis (changes requested):

lib/charms/prometheus_k8s/v0/prometheus_scrape.py
```diff
@@ -11,6 +11,13 @@
 shared between Prometheus charms and any other charm that intends to
 provide a scrape target for Prometheus.
 
+## Requirements
```
sed-i, line 14:
```suggestion
## Dependencies
```

lib/charms/prometheus_k8s/v0/prometheus_scrape.py
```diff
@@ -11,6 +11,13 @@
...
+
+Using this library requires you to fetch the juju_topology library from
+[observability-libs]((https://github.com/canonical/observability-libs/blob/main/lib/charms/observability_libs/v0/juju_topology.py).
```
sed-i, line 17: This doc is rendered on charmhub and by convention we try to have all links point to charmhub resources as well, when appropriate.

```suggestion
[observability-libs](https://charmhub.io/observability-libs/libraries/juju_topology).
```

lib/charms/prometheus_k8s/v0/prometheus_scrape.py
```diff
@@ -1,6 +1,6 @@
 # Copyright 2021 Canonical Ltd.
 # See LICENSE file for licensing details.
-"""## Overview.
+"""## Overview
```
sed-i, line 3: The easy way out of 
```
D415 First line should end with a period, question mark, or exclamation point
```
is to have the period in there.
Alternative suggestions welcome!

lucabello: Oh yeah how did I not see it; I think ideally that line should be a one-line description of the library, though it feels a little repetitive with the overview right below it.

I'm applying this suggestion in the meantime, seeing if anyone has better ideas!

PR #385, opened 2022-10-13
Always refresh metrics endpoint reldata on update-status

### LINKED ISSUE #368: External hostname is not updated if an ingress is added *after* relating a charm to Prometheus
### Bug Description

See title. If you first relate to Prometheus and then to Traefik, it all works as expected. The other way around, no cigar.

### To Reproduce

-

### Environment

-

### Relevant log output

```shell
-
```

### Additional context

[We could have used the ingress established/revoked events, but these are unfortunately fired prematurely](https://github.com/canonical/traefik-k8s-operator/issues/78)

sed-i:
## Issue
- #368: metrics endpoint is not updated to ingressed address after relating to traefik.

## Solution
Add [`self._set_scrape_job_spec`](https://github.com/canonical/prometheus-k8s-operator/blob/0e410410da9dd6f64fbd63c2b35383ecc7d2b9cb/lib/charms/prometheus_k8s/v0/prometheus_scrape.py#L1577) as an observer to `update-status`.

Fixes #368.

## Drawbacks
The user would have to wait for the update-status hook interval to elapse.

## Context
This is an alternative to #384.

## Release Notes
Always refresh metrics endpoint reldata on update-status.


sed-i:
#384 is probably the better solution. Closing this one.

PR #390, opened 2022-10-17
Pre commit

beliaev-maksim:
switch styler to pre-commit to simplify the process

backwards compatible

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,3 +1,4 @@
+# exclude library folder from checks, except prometheus library itself
 exclude: ^((?!lib/charms/prometheus_k8s)lib).*$
 repos:
   - repo: https://github.com/PyCQA/autoflake
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,0 +1,33 @@
+exclude: ^((?!lib/charms/prometheus_k8s)lib).*$
+repos:
+  - repo: https://github.com/PyCQA/autoflake
+    rev: v1.7.6
+    hooks:
+      - id: autoflake
+  - repo: https://github.com/psf/black
+    rev: 22.3.0
+    hooks:
+      - id: black
+  - repo: https://github.com/asottile/blacken-docs
+    rev: v1.12.1
+    hooks:
+      - id: blacken-docs
+        additional_dependencies: [black]
+  - repo: https://github.com/pycqa/isort
+    rev: 5.10.1
+    hooks:
+      - id: isort
+        name: isort (python)
+
+  # official flake8 doesn't support pyproject.toml configuration.
+  - repo: https://github.com/csachs/pyproject-flake8
+    rev: v5.0.4
+    hooks:
+      - id: pyproject-flake8
+        name: pyproject-flake8
+
+  - repo: https://github.com/Lucas-C/pre-commit-hooks-nodejs
+    rev: v1.1.2
+    hooks:
+      - id: markdown-toc
+        name: Table of Contents
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -33,6 +33,11 @@ copyright-check = "True"
 copyright-author = "Canonical Ltd."
 copyright-regexp = "Copyright\\s\\d{4}([-,]\\d{4})*\\s+%(author)s"
 
+[tool.autoflake]
+check = true
+remove-all-unused-imports = true
+in-place = true
+
 [tool.mypy]
 pretty = true
 python_version = 3.8
--- a/lib/charms/prometheus_k8s/v0/prometheus_scrape.py
+++ b/lib/charms/prometheus_k8s/v0/prometheus_scrape.py
@@ -1784,8 +1784,8 @@ class MetricsEndpointAggregator(Object):
         {
             "prometheus": "monitoring",
             "scrape_target": "prometheus-target",
-            "alert_rules": "prometheus-rules"
-        }
+            "alert_rules": "prometheus-rules",
+        },
     )
     ```
 
--- a/tox.ini
+++ b/tox.ini
@@ -42,6 +42,7 @@ deps =
     flake8-copyright
     flake8-builtins
     pep8-naming
+    pre-commit
 commands =
     codespell {[vars]lib_path}
     codespell . --skip .git --skip .tox --skip build --skip lib --skip venv --skip .mypy_cache
--- a/tox.ini
+++ b/tox.ini
@@ -30,32 +30,23 @@ passenv =
 [testenv:fmt]
 description = Apply coding style standards to code
 deps =
-    black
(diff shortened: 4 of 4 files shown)


pyproject.toml
```diff
@@ -33,6 +33,11 @@ copyright-check = "True"
 copyright-author = "Canonical Ltd."
 copyright-regexp = "Copyright\\s\\d{4}([-,]\\d{4})*\\s+%(author)s"
 
+[tool.autoflake]
+check = true
+remove-all-unused-imports = true
```
sed-i, line 38: Great addition!

sed-i (commented):
Iiuc, this change is suggesting:

- To add a non-ubiquitous dependency (pre-commit) on top of tox, and migrate from tox `commands` section (bash) to a new kind of yaml syntax, currently unfamiliar to us.
- To force devs into the pre-commit workflow.

1. Is the above correct?
2. How is this better than keep what we have and each dev creating (or not!) a `~/.githooks/pre-push` file to their liking?

pyproject.toml
beliaev-maksim: pre-commit is a common practice is lots of repos. Nobody is pushed to use pre-commit if they do not want to, but if you want, then you just go for `pre-commit install`.

every linting tool provides a pre-commit hook, I think it is also a positive sign

the only change is that we provide an extra dependency, but I do not know how much of that is a drawback.

pre-commit provides a native support as hooks, you do not need to configure anything extra. It will run only on files added to staging. Do not see a point to reinvent the wheel when there is already a great community solution.

dstathis: I would prefer to just have a pre-commit git hook that you can copy in if you want. I see no reason to install extra software to do this.

beliaev-maksim: what is the sense to spend time creating and debugging hooks if we have out of the box solution?

sed-i:
I'm rethinking this.

## How is this better than `git config core.hooksPath`?
Git itself already supports a custom hooks dir, e.g. `~/.githooks` via [`core.hooksPath`](https://git-scm.com/docs/git-config#Documentation/git-config.txt-corehooksPath).
However, this would need to be a single hook that fits all repos. Specifically for o11y charms that is ok because we always have `static, unit` tox envs, etc.
Also, our CI is currently centralized (always checks static-charm and static-lib but never static-unit). With pre-commit we could customize CI per repo while still keeping all the gh yamls centralized.

However, in the general case, pre-commit is probably better.

## How is this better than `.git/info/exclude`?
Devs can create their per-repo pre-commit settings and add it to `.git/info/exclude` so it's not committed.
However, the dev would need to manually keep track of all those files. If the local repo is deleted, then the pre-commit config is gone.

## How to improve the current proposal?
- Keep tox as the centrepiece: keep all the logic inside tox.ini, and call `tox -e env1,env2` from pre-commit.
- Phrase a community guideline (under CONTRIBUTING.md?) that clarifies why pre-commit yes, but .vscode no.

beliaev-maksim:
@sed-i 
following our discussion during sprint.

The topic was either `tox` should call `pre-commit` or `pre-commit` calls `tox`. 
I conclude that as designed by the product it should be `tox` calling `pre-commit`.

Here are the examples of projects:
https://github.com/pallets/flask
https://github.com/apache/superset
https://github.com/getsentry/responses
https://github.com/home-assistant/core

side finding:
while I was going through the list of most starred projects on GH I could find that there are also lots of projects that use pre-commit as in a way it is done in this PR, but they do not use tox at all


so, it is not just my personal preference (since otherwise I will not have this strong opinion) but the way how the tool is designed and adopted by the community

.pre-commit-config.yaml
```diff
@@ -0,0 +1,34 @@
...
+        name: isort (python)
+
+  # official flake8 doesn't support pyproject.toml configuration.
+  - repo: https://github.com/csachs/pyproject-flake8
+    rev: v5.0.4
+    hooks:
+      - id: pyproject-flake8
+        name: pyproject-flake8
+
+  - repo: https://github.com/Lucas-C/pre-commit-hooks-nodejs
```
sed-i, line 30: > Does not require to have NodeJS installed: pre-commit will fetch & install it under the hood.

:scream: 

But also, we don't use `<!-- toc -->` comments yet. Probably better to leave this one off for now.

sed-i (commented):
Thanks @beliaev-maksim for the examples.

The hooks collection seems like a continuously growing and insightful set of [quality checks](https://github.com/pre-commit/pre-commit-hooks/tree/main/pre_commit_hooks). However:
- they do involve an additional level of indirection: using hooks, rather than the tools themselves, means depending on even more repos and versioning the hooks rather than the tools
- the hooks are not _that_ sophisticated - after little investigation we can easily figure out which cli args to use with the tools themselves.

The desired scope of "pre-commit" actions varies per proj/dev:
- responses and flask do not do typing checks in pre-commit
- superset and homeassistant use mypy hook in pre-commit and pylint directly in tox.
- I personally run fmt, lint, static-charm, static-lib and unit before committing. @dstathis and @rbarry82 preferences are probably different.

Seems like the main decision here is between
- Simplicity&flexibility but discipline (use the tools directly but need to remember to tox);  vs.
- More deps but convenience (use the hooks, and if pre-commit is installed, don't need to remember to tox)

WDYT @Abuelodelanada @dstathis @lucabello @PietroPasotti @rbarry82 @simskij

.pre-commit-config.yaml
beliaev-maksim: JFYI

- hooks are provided by the software provider
- hooks do not require CLI args, you can, if you want, but as current PR shows, tools will rely on `pyproject.toml` if no argument is supplied


Overall, PR is backwards compatible. For those who do not want to use it, just use `tox -e fmt` and it will do the same thing. For those who want to autofix style in each commit, install via `pre-commit install`

PS
I just sent you a list of projects that use `tox` + `pre-commit` to show use case, but have a look for amount of projects that use only `pre-commit`. Almost every huge python project

lucabello:
Personally, I'm not a big fan of pre-commit hooks, especially when the amount of things they are doing is very small (hence they are not particularly necessary).

I believe there is an argument to be made on the fact that: hiding the explicit verification steps that should happen before a commit (testing, running `tox` in general) is transforming the process to "I just write code and commit", and that can be an unhealthy approach. I like the affirmative action of verifying that your code is good **before** committing, especially when it's very simple to do so.

Considering that it's something that it can be setup easily for an individual person that wants it, I don't think it should be the default; the "you can always disable it" argument is a slippery slope IMHO; but it's just my two cents 🌈

sed-i:
This is an insightful record of how CI could look like with pre-commit hooks.
Thanks @beliaev-maksim for the effort you put it in making it all tick.
I'm inclined to close this for now until a wider consensus is reached.

Issue #388, opened 2022-10-14
Error on application remove


PietroPasotti:
### Bug Description

My guess is that attempting to `config-get` after some point during the teardown sequence gives permission errors.
If we could postpone that `self.external_url`-dependent logic to some specific hook, that would probably fix this. 

### To Reproduce

I think it might be flaky, but repro is:

`juju deploy prometheus-k8s`

wait for thing to settle

`juju remove-application prometheus-k8s # --> error soon after`

### Environment

microk8s v1.25.2
juju 3.0-beta4 controller

### Relevant log output

```shell
Traceback (most recent call last):
  File "./src/charm.py", line 762, in <module>
    main(PrometheusCharm)
  File "/var/lib/juju/agents/unit-prom-1/charm/venv/ops/main.py", line 426, in main
    charm = charm_class(framework)
  File "./src/charm.py", line 97, in __init__
    external_url = urlparse(self.external_url)
  File "./src/charm.py", line 245, in external_url
    if web_external_url := self.model.config.get("web_external_url"):
  File "/usr/lib/python3.8/_collections_abc.py", line 660, in get
    return self[key]
  File "/var/lib/juju/agents/unit-prom-1/charm/venv/ops/model.py", line 547, in __getitem__
    return self._data[key]
  File "/var/lib/juju/agents/unit-prom-1/charm/venv/ops/model.py", line 531, in _data
    data = self._lazy_data = self._load()
  File "/var/lib/juju/agents/unit-prom-1/charm/venv/ops/model.py", line 1092, in _load
    return self._backend.config_get()
  File "/var/lib/juju/agents/unit-prom-1/charm/venv/ops/model.py", line 2289, in config_get
    out = self._run('config-get', return_output=True, use_json=True)
  File "/var/lib/juju/agents/unit-prom-1/charm/venv/ops/model.py", line 2179, in _run
    raise ModelError(e.stderr)
ops.model.ModelError: b'ERROR permission denied\n'
```

PR #392, opened 2022-10-19
move external_url dependees out of constructor

lucabello:
## Issue
This PR addresses #388.

## Solution
Moving things depending on `self.external_url` outside of the constructor, so they are not called on removing the application.

## Release Notes
I moved the logic depending on `self.external_url` to a separate function and invoked that in `_configure()` instead of `__init__()`.  
This way, that code is executed in every part of the lifecycle except the removal.

### DIFF
--- a/charmcraft.yaml
+++ b/charmcraft.yaml
@@ -12,6 +12,9 @@ parts:
   charm:
     build-packages:
       - git
+    charm-python-packages:
+      - wheel==0.37.1
+      - setuptools==45.2.0
   cos-tool:
     plugin: dump
     source: .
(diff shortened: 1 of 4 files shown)


sed-i (commented):
Typically we keep all constructions under `__init__`.
What do you think about an event guard in `__init__`?
```python
if os.getenv("JUJU_HOOK_NAME") in ["stop", "remove"]:
    return  # or sys.exit(0)?
```
Or try-except?
```python
# inside external_url()
try:
    # config-get, etc.
except ops.model.ModelError as e:
    logger.error("Failed obtaining external url: %s. Shutting down?", e)
    # return fqdn?
```

lucabello:
An event guard probably make more sense!

rbarry82:
I was slow on the uptake for the review, but please don't directly check `JUJU_HOOK_NAME`. This isn't a promise Juju gives us, and even OF doesn't try to be that clever. catching `ModelError` is much less brittle

simskij (approved):

PR #395, opened 2022-11-01
Add the ability to use lookaside caches for targets

rbarry82:
## Issue
Give `MetricsEndpointProvider` the ability to have a lookaside source for targets which is not always in relations/relation data

## Solution
Invoke a `Callable` to get an a point-in-time list of jobs, regardless of where the object is instantiated. Make the method to set the scrape job spec public.

## Release Notes
Add the ability to use lookaside caches for targets


dstathis (approved):

sed-i:
Just curious why not leave it for the user to do in charm code:
```python
self._scraping = MetricsEndpointProvider(
            self,
            relation_name="...",
            jobs=self.the_usual_scrape_job + self.that_lookaside_callable(),
            ...,
        )
```

PR #398, opened 2022-11-15
Fix #386 - alert rules disappear if k8s cluster gets rebooted

### LINKED ISSUE #386: alert rules disappear if k8s cluster gets rebooted
### Bug Description

During testing microk8s cluster went down. Due to some issue cluster entered CrashLoopBackOff and juju restarted it. Once units went up again, alert rules were empty

### To Reproduce

restart k8s cluster with deployed prometheus
check alert rules

### Environment

-

### Relevant log output

```shell
-
```

### Additional context

_No response_

Abuelodelanada:
## Issue

This PR fixes #386 

## Testing Instructions

- Deploy Prometheus
- Deploy a charm that implements `MetricsEndpointProvider` interface (Zinc)
- Relate both charms
- Check that the alert rules provided by Zinc are in Prometheus
- restart microk8s
- Check that the alert rules provided by Zinc are in Prometheus ;-)

### DIFF
--- a/src/charm.py
+++ b/src/charm.py
@@ -469,6 +469,8 @@ class PrometheusCharm(CharmBase):
         if alerts_hash == self._stored.alerts_hash:
             alert_rules_changed = False
 
+        # Pushing files every time for situations such as cluster restart:
+        # Relation data and stored state match, but files haven't been written yet.
         container.remove_path(RULES_DIR, recursive=True)
         self._push_alert_rules(container, self.metrics_consumer.alerts())
         self._push_alert_rules(container, self.remote_write_provider.alerts())
--- a/src/charm.py
+++ b/src/charm.py
@@ -411,7 +411,7 @@ class PrometheusCharm(CharmBase):
             self.unit.set_workload_version(version)
         else:
             logger.debug(
-                "Cannot set workload version at this time: could not get Alertmanager version."
+                "Cannot set workload version at this time: could not get Prometheus version."
             )
 
     def _update_config(self, container) -> bool:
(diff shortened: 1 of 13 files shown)


src/charm.py
```diff
@@ -459,20 +459,21 @@ def _set_alerts(self, container) -> bool:
...
         """
+        alert_rules_changed = True
         metrics_consumer_alerts = self.metrics_consumer.alerts()
         remote_write_alerts = self.remote_write_provider.alerts()
-
         alerts_hash = sha256(str(metrics_consumer_alerts) + str(remote_write_alerts))
+
         if alerts_hash == self._stored.alerts_hash:
-            return False
+            alert_rules_changed = False
```
sed-i, line 470: Seems like the logic around `alert_rules_changed` could be replaced with:

```python
alert_rules_changed = alerts_hash != self._stored.alerts_hash
```

src/charm.py
```diff
@@ -459,20 +459,21 @@ def _set_alerts(self, container) -> bool:
...
 
         container.remove_path(RULES_DIR, recursive=True)
```
sed-i, line 472: Should we add a comment here?

```suggestion
        # Pushing files every time for situations such as cluster restart:
        # Relation data and stored state match, but files haven't been written yet.
        container.remove_path(RULES_DIR, recursive=True)
```

sed-i (approved):

src/charm.py
Abuelodelanada: do you mean something like this?

```python
    def _set_alerts(self, container) -> bool:
        """Create alert rule files for all Prometheus consumers.

        Args:
            container: the Prometheus workload container into which
                alert rule files need to be created. This container
                must be in a pebble ready state.

        Returns: A boolean indicating if new or different alert rules were pushed.
        """
        metrics_consumer_alerts = self.metrics_consumer.alerts()
        remote_write_alerts = self.remote_write_provider.alerts()
        alerts_hash = sha256(str(metrics_consumer_alerts) + str(remote_write_alerts))

        container.remove_path(RULES_DIR, recursive=True)
        self._push_alert_rules(container, self.metrics_consumer.alerts())
        self._push_alert_rules(container, self.remote_write_provider.alerts())
        self._stored.alerts_hash = alerts_hash
        return alerts_hash != self._stored.alerts_hash
```

sed-i: Not quite: at the return statement they would always match because the stored value was updated one line above.

src/charm.py
```diff
@@ -420,7 +420,7 @@ def _on_pebble_ready(self, event) -> None:
             self.unit.set_workload_version(version)
         else:
             logger.debug(
-                "Cannot set workload version at this time: could not get Alertmanager version."
+                "Cannot set workload version at this time: could not get Prometheus version."
             )
 
     def _update_config(self, container) -> bool:
```
sed-i, line 426: Seems like `_update_config` is not used anywhere. This is odd.

Abuelodelanada: Let's follow this issue here: https://github.com/canonical/prometheus-k8s-operator/issues/399

PR #406, opened 2022-11-29
move has_metric

beliaev-maksim:
refactoring

### DIFF
--- a/tests/integration/helpers.py
+++ b/tests/integration/helpers.py
@@ -237,8 +237,7 @@ def get_podspec(ops_test: OpsTest, app_name: str, container_name: str):
 
 
 async def has_metric(ops_test, query: str, app_name: str) -> bool:
-    # Throws if the query does not return any time series within 5 minutes,
-    # and as a consequence, fails the test
+    """Returns True if the query returns any time series; False otherwise."""
     for timeseries in await run_promql(ops_test, query, app_name):
         if timeseries.get("metric"):
             return True
--- a/tests/integration/helpers.py
+++ b/tests/integration/helpers.py
@@ -234,3 +234,13 @@ def get_podspec(ops_test: OpsTest, app_name: str, container_name: str):
     pod = client.get(Pod, name=f"{app_name}-0", namespace=ops_test.model_name)
     podspec = next(iter(filter(lambda ctr: ctr.name == container_name, pod.spec.containers)))
     return podspec
+
+
+async def has_metric(ops_test, query: str, app_name: str) -> bool:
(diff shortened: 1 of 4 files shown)


Abuelodelanada (approved):

beliaev-maksim:
@sed-i 
I believe that it is unrelated itest failure. test looks to be unstable

how can we update the test to make it working ?

tests/integration/helpers.py
```diff
@@ -234,3 +234,13 @@ def get_podspec(ops_test: OpsTest, app_name: str, container_name: str):
     pod = client.get(Pod, name=f"{app_name}-0", namespace=ops_test.model_name)
     podspec = next(iter(filter(lambda ctr: ctr.name == container_name, pod.spec.containers)))
     return podspec
+
+
+async def has_metric(ops_test, query: str, app_name: str) -> bool:
+    # Throws if the query does not return any time series within 5 minutes,
+    # and as a consequence, fails the test
```
sed-i, line 241:
```suggestion
    """Returns True if the query returns any time series; False otherwise."""
```

tox.ini
```diff
@@ -51,7 +51,7 @@ deps =
     isort
 commands =
     codespell {[vars]lib_path}
-    codespell . --skip .git --skip .tox --skip build --skip lib --skip venv --skip .mypy_cache
+    codespell . --skip .git --skip .tox --skip build --skip lib --skip venv* --skip .mypy_cache
```
sed-i, line 54: Just curious which cases the `venv*` covers.

sed-i:
> @sed-i I believe that it is unrelated itest failure. test looks to be unstable
> 
> how can we update the test to make it working ?

unit-1 of prom has an outdated url: fqdn instead of ingress:
```
"[http://prometheus-1.prometheus-endpoints.test-external-url-imkm.svc.cluster.local:80/metrics](http://prometheus-1.prometheus-endpoints.test-external-url-imkm.svc.cluster.local/metrics)"
```
(unit-0 is had the correct ingress url).

This might be due to https://github.com/canonical/traefik-k8s-operator/pull/107

tox.ini
beliaev-maksim: I have `venv37` and `venv310` and run interactive debugger with different versions

beliaev-maksim:
> 
> This might be due to [canonical/traefik-k8s-operator#107](https://github.com/canonical/traefik-k8s-operator/pull/107)

@sed-i 
if I got it right, nothing we can do in this repo for now, right?

sed-i (approved):

PR #408, opened 2022-12-01
fetch-lib

sed-i:
## Issue
https://github.com/canonical/catalogue-k8s-operator/issues/3

## Solution
Fetch-lib to obtain changes from https://github.com/canonical/traefik-k8s-operator/pull/107.

## Context
NTA.

## Testing Instructions
CI now, load test later.

## Release Notes
Fetch-lib.


rbarry82:
We should really wait for the Grafana patch to merge or we'll just have to do this again immediately.

sed-i:
> We should really wait for the Grafana patch to merge or we'll just have to do this again immediately.

Good point. I exported the load test dashboards from grafana 8 but the dropdowns fix is certainly needed. I'll switch to draft for now.

rbarry82:
> > We should really wait for the Grafana patch to merge or we'll just have to do this again immediately.
> 
> Good point. I exported the load test dashboards from grafana 8 but the dropdowns fix is certainly needed. I'll switch to draft for now.

I mean, it kind of really isn't at all for the Prometheus self-monitoring dashboard (which is the only thing it would apply to here, and honestly, the library fix for Grafana is entirely on the consumer side anyway), just that waiting avoids a follow-up fetch-lib within 24h

Abuelodelanada (approved):

PR #421, opened 2023-01-13
The readme is not consistent with the actual usage

### LINKED ISSUE #413: The readme is not consistent with the actual usage
The readme is not consistent with the actual usage. When charmhub got rid of the Docs tab, the content from discourse was moved to the Descriptions tab, making it super confusing as it now contains TWO overviews with entirely different content. 

We should
 
- Make sure the overview taken from the readme is up to date
- Incorporate everything valuable from the `Docs - Overview` discourse post and then delete that post.
- Incorporate everything valuable from the `Relations` discourse post and then delete that post.

Abuelodelanada:
## Issue

This PR fixes #413

> The readme is not consistent with the actual usage. When charmhub got rid of the Docs tab, the content from discourse was moved to the Descriptions tab, making it super confusing as it now contains TWO overviews with entirely different content. 
>
> We should
>  
> - Make sure the overview taken from the readme is up to date
> - Incorporate everything valuable from the `Docs - Overview` discourse post and then delete that post.
> - Incorporate everything valuable from the `Relations` discourse post and then delete that post.
>

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,5 +1,9 @@
 # Prometheus Charmed Operator for Kubernetes
 
+
+[![Release to Edge](https://github.com/canonical/prometheus-k8s-operator/actions/workflows/release-edge.yaml/badge.svg)](https://github.com/canonical/prometheus-k8s-operator/actions/workflows/release-edge.yaml)
+
+
 ## Description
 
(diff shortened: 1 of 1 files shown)


simskij (approved):
Excellent work! Good job! 🎉

PR #429, opened 2023-01-20
Allow optional (default false) DNS lookups for aggregator targets

rbarry82:
## Issue
Some LMA1 dashboards and workflows rely on the presence of a `dns_name` label, which is set by the `prometheus2` charm when a new target is received. After investigation, the remaining set of labels is a non-overlapping set, with no usable substitute available via relabeling.

## Solution
Do hostname lookups, optionally. There are too many edge cases with trying to do this as a default, and not all workflows need this label. Behave the same way as the old charm -- if a lookup fails, set the name with the passed input/hostname anyway.

## Context
See [here](https://github.com/canonical/cos-proxy-operator/issues/32)

## Release Notes
Allow optional (default false) DNS lookups for aggregator targets

### DIFF
--- a/lib/charms/prometheus_k8s/v0/prometheus_scrape.py
+++ b/lib/charms/prometheus_k8s/v0/prometheus_scrape.py
@@ -2131,7 +2131,7 @@ class MetricsEndpointAggregator(Object):
                         "juju_application": application_name,
                         "juju_unit": unit_name,
                         "host": target["hostname"],
-                        **self._build_extra_info(target),
+                        **self._static_config_extra_labels(target),
                     },
                 }
                 for unit_name, target in targets.items()
--- a/lib/charms/prometheus_k8s/v0/prometheus_scrape.py
+++ b/lib/charms/prometheus_k8s/v0/prometheus_scrape.py
@@ -368,7 +368,7 @@ LIBAPI = 0
 
 # Increment this PATCH version before using `charmcraft publish-lib` or reset
 # to 0 if you are raising the major API version
-LIBPATCH = 28
+LIBPATCH = 29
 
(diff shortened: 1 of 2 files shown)


lib/charms/prometheus_k8s/v0/prometheus_scrape.py
```diff
@@ -2130,6 +2142,20 @@ def _static_scrape_job(self, targets, application_name, **kwargs) -> dict:
 
         return job
 
+    def _build_extra_info(self, target: Dict[str, str]) -> Dict[str, str]:
```
sed-i, line 2145:
```suggestion
    def _static_config_extra_labels(self, target: Dict[str, str]) -> Dict[str, str]:
```

sed-i (approved):

rbarry82:
Thanks @sed-i ! 

I will readily admit that I'm just leaving this PR alone for now so we don't get more "Update Libraries" CI churn (for something which is already part of a released `cos-proxy`, the only code which will exercise it), but it's nice to have it ready :)

lib/charms/prometheus_k8s/v0/prometheus_scrape.py
```diff
@@ -2120,6 +2131,7 @@ def _static_scrape_job(self, targets, application_name, **kwargs) -> dict:
                         "juju_application": application_name,
                         "juju_unit": unit_name,
                         "host": target["hostname"],
+                        **self._build_extra_info(target),
```
rbarry82, line 2134:
```suggestion
                        **self._static_config_extra_labels(target),
```

Abuelodelanada (approved):

PR #431, opened 2023-01-20
Stop using remote storage

### LINKED ISSUE #399: `_update_config` is not used anywhere
> Seems like `_update_config` is not used anywhere. This is odd.

_Originally posted by @sed-i in https://github.com/canonical/prometheus-k8s-operator/pull/398#discussion_r1024723712_

### LINKED ISSUE #414: Scrape targets disappear after reprovisioning
### Bug Description

I had the load test running for a while, and then something went wrong and all the charms went through start -> pebble-ready.

```
unit-grafana-0: 2022-12-08 01:05:14 ERROR juju.worker.dependency "api-caller" manifold worker returned unexpected error: [40652e] "unit-grafana-0" cannot open api: cannot resolve "controller-service.controller-uk8s.svc.cluster.local": lookup controller-service.controller-uk8s.svc.cluster.local: i/o timeout
unit-grafana-0: 2022-12-08 01:05:32 WARNING juju.worker.proxyupdater unable to set snap core settings [proxy.http= proxy.https= proxy.store=]: exec: "snap": executable file not found in $PATH, output: ""
unit-grafana-0: 2022-12-08 01:05:46 WARNING juju.worker.uniter.context could not retrieve the cloud API version: unable to determine legacy status for namespace cos-lite-load-test: Get "https://10.152.183.1:443/api/v1/namespaces/cos-lite-load-test": net/http: TLS handshake timeout
unit-grafana-0: 2022-12-08 01:05:55 DEBUG unit.grafana/0.juju-log Operator Framework 1.5.4+2.g320e7e0 up and running.
unit-grafana-0: 2022-12-08 01:05:59 INFO unit.grafana/0.juju-log Running legacy hooks/start.
unit-grafana-0: 2022-12-08 01:06:05 DEBUG unit.grafana/0.juju-log Operator Framework 1.5.4+2.g320e7e0 up and running.
unit-grafana-0: 2022-12-08 01:06:05 DEBUG unit.grafana/0.juju-log Charm called itself via hooks/start.
unit-grafana-0: 2022-12-08 01:06:06 DEBUG unit.grafana/0.juju-log Legacy hooks/start exited with status 0.
unit-grafana-0: 2022-12-08 01:06:06 DEBUG unit.grafana/0.juju-log Reading alert rule from /var/lib/juju/agents/unit-grafana-0/charm/src/prometheus_alert_rules/unit_unavailable.rule
unit-grafana-0: 2022-12-08 01:06:06 DEBUG unit.grafana/0.juju-log Emitting Juju event start.
```

At this point, relation data contained scrape targets:
```
$ juju show-unit scrape-config/0
```
```yaml
scrape-config/0:
  workload-version: n/a
  opened-ports: []
  charm: ch:amd64/focal/prometheus-scrape-config-k8s-38
  leader: true
  life: alive
  relation-info:
  - relation-id: 24
    endpoint: configurable-scrape-jobs
    related-endpoint: metrics-endpoint
    application-data:
      scrape_jobs: '[{"job_name": "juju_cos-lite-load-test_40652ec_scrape-target_external_jobs",
        "static_configs": [{"targets": ["avalanche.us-central1-a.c.lma-light-load-testing.internal:9001",
        "avalanche.us-central1-a.c.lma-light-load-testing.internal:9002", "avalanche.us-central1-a.c.lma-light-load-testing.internal:9003", ...
```

But these are not included in the targets:
```
$ curl -s 10.1.79.219/cos-lite-load-test-prometheus-0/api/v1/targets | jq
```
```json
{
  "status": "success",
  "data": {
    "activeTargets": [
      {
        "discoveredLabels": {
          "__address__": "localhost:9090",
          "__metrics_path__": "/metrics",
          "__scheme__": "http",
          "__scrape_interval__": "15s",
          "__scrape_timeout__": "10s",
          "job": "prometheus"
        },
        "labels": {
          "instance": "localhost:9090",
          "job": "prometheus"
        },
        "scrapePool": "prometheus",
        "scrapeUrl": "http://localhost:9090/metrics",
        "globalUrl": "http://pd-ssd-4cpu-8gb.us-central1-a.c.lma-light-load-testing.internal:80/metrics",
        "lastError": "server returned HTTP status 404 Not Found",
        "lastScrape": "2022-12-08T18:59:12.275395831Z",
        "lastScrapeDuration": 0.000530596,
        "health": "down",
        "scrapeInterval": "15s",
        "scrapeTimeout": "10s"
      }
    ],
    "droppedTargets": []
  }
}
```

And the config file is the one baked in the image, not the one generated by the charm:

```
$ juju ssh --container prometheus prometheus/0 cat /etc/prometheus/prometheus.yml
```
```yaml
# my global config
global:
  scrape_interval: 15s # Set the scrape interval to every 15 seconds. Default is every 1 minute.
  evaluation_interval: 15s # Evaluate rules every 15 seconds. The default is every 1 minute.
  # scrape_timeout is set to the global default (10s).

# Alertmanager configuration
alerting:
  alertmanagers:
    - static_configs:
        - targets:
          # - alertmanager:9093

# Load rules once and periodically evaluate them according to the global 'evaluation_interval'.
rule_files:
  # - "first_rules.yml"
  # - "second_rules.yml"

# A scrape configuration containing exactly one endpoint to scrape:
# Here it's Prometheus itself.
scrape_configs:
  # The job name is added as a label `job=<job_name>` to any timeseries scraped from this config.
  - job_name: "prometheus"

    # metrics_path defaults to '/metrics'
    # scheme defaults to 'http'.

    static_configs:
      - targets: ["localhost:9090"]
```

### To Reproduce

Not sure. `kubectl delete` perhaps?

### Environment

Edge charms. Juju 2.9.34.

### Relevant log output

```shell
See above.
```

### Additional context

Related to #398.

### LINKED ISSUE #418: Review use of stored state to cache config file hashes
> the caching that you're doing seems like you need to push it again on some upgrades, as whenever the pod gets rescheduled then /etc/prometheus/prometheus.yml isn't going to be there
> so it [local storage instead of controller storage; see [ops/880](https://github.com/canonical/operator/issues/880)] actually helps you, because you actually should be pushing on every pebble_ready event

-- @jameinel

Related: #414

sed-i:
## Issue
- Incorrectly assumed that the charm uses controller storage, when in fact it was using local storage (see https://github.com/canonical/operator/issues/880)
- Config and alerts hash was originally intended to be written to controller storage, which is wrong because files wouldnt be written to container after an upgrade (because hash(relation-data) would match the hash in stored state).

## Solution
- Do not use remote storage (remove `_stored`).
- Save hashes to container filesystem.

Fixes #399 
Fixes #414 
Fixes #418

## Context
NTA.

## Testing Instructions
- See new tests.
- In a `prom --(self monitoring)-- prom` constellation, refresh one of the proms and make sure the config file, alert rules and scrape targets are as they were before the refresh (covered in itest).

## Release Notes
Stop using remote storage.

### DIFF
--- a/src/charm.py
+++ b/src/charm.py
@@ -428,23 +428,6 @@ class PrometheusCharm(CharmBase):
                 "Cannot set workload version at this time: could not get Prometheus version."
             )
 
-    def _update_config(self, container) -> bool:
-        """Pushes new config, if needed.
-
-        Returns a boolean indicating if a new configuration was pushed.
-        """
-        config = self._generate_prometheus_config(container)
-        config_hash = sha256(config)
-
-        if config_hash == self._stored.config_hash:
-            return False
-
-        logger.debug("Prometheus config changed")
-        container.push(PROMETHEUS_CONFIG, config, make_dirs=True)
-        self._stored.config_hash = config_hash
-        logger.info("Pushed new configuration")
-        return True
-
     def _update_layer(self, container) -> bool:
         current_planned_services = container.get_plan().services
         new_layer = self._prometheus_layer
--- a/src/charm.py
+++ b/src/charm.py
@@ -46,7 +46,6 @@ from lightkube import Client
 from lightkube.core.exceptions import ApiError as LightkubeApiError
 from lightkube.resources.core_v1 import PersistentVolumeClaim, Pod
 from ops.charm import ActionEvent, CharmBase
-from ops.framework import StoredState
 from ops.main import main
 from ops.model import (
     ActiveStatus,
@@ -63,6 +62,8 @@ from utils import convert_k8s_quantity_to_legacy_binary_gigabytes
 
 PROMETHEUS_CONFIG = "/etc/prometheus/prometheus.yml"
 RULES_DIR = "/etc/prometheus/rules"
+CONFIG_HASH_PATH = "/etc/prometheus/config.hash"
+ALERTS_HASH_PATH = "/etc/prometheus/alerts.hash"
 
 logger = logging.getLogger(__name__)
 
@@ -74,15 +75,22 @@ def sha256(hashable) -> str:
     return hashlib.sha256(hashable).hexdigest()
 
 
+def pull(container, path, *, default: Optional[str] = None) -> Optional[str]:
+    try:
+        return container.pull(path, encoding="utf8").read()
+    except FileNotFoundError:
+        return default
+
+
+def push(container, path, contents):
+    container.push(path, contents, make_dirs=True, encoding="utf8")
+
+
 class PrometheusCharm(CharmBase):
     """A Juju Charm for Prometheus."""
 
-    _stored = StoredState()
-
     def __init__(self, *args):
         super().__init__(*args)
-
-        self._stored.set_default(config_hash=None, alerts_hash=None)
 
         self._name = "prometheus"
         self._port = 9090
@@ -461,14 +469,15 @@ class PrometheusCharm(CharmBase):
         metrics_consumer_alerts = self.metrics_consumer.alerts()
         remote_write_alerts = self.remote_write_provider.alerts()
         alerts_hash = sha256(str(metrics_consumer_alerts) + str(remote_write_alerts))
-        alert_rules_changed = alerts_hash != self._stored.alerts_hash
+        alert_rules_changed = alerts_hash != pull(container, ALERTS_HASH_PATH)
 
         # Pushing files every time for situations such as cluster restart:
         # Relation data and stored state match, but files haven't been written yet.
+        # FIXME don't push every time
         container.remove_path(RULES_DIR, recursive=True)
-        self._push_alert_rules(container, self.metrics_consumer.alerts())
-        self._push_alert_rules(container, self.remote_write_provider.alerts())
-        self._stored.alerts_hash = alerts_hash
+        self._push_alert_rules(container, metrics_consumer_alerts)
+        self._push_alert_rules(container, remote_write_alerts)
+        push(container, ALERTS_HASH_PATH, alerts_hash)
         return alert_rules_changed
 
     def _push_alert_rules(self, container, alerts):
@@ -488,7 +497,7 @@ class PrometheusCharm(CharmBase):
 
             rules = yaml.safe_dump(rules_file)
 
-            container.push(path, rules, make_dirs=True)
+            push(container, path, rules)
             logger.debug("Updated alert rules file %s", filename)
 
     def _generate_command(self) -> str:
@@ -725,16 +734,16 @@ class PrometheusCharm(CharmBase):
         config_hash = sha256(
             yaml.safe_dump({"prometheus_config": prometheus_config, "certs": certs})
         )
-        if config_hash == self._stored.config_hash:
+        if config_hash == pull(container, CONFIG_HASH_PATH):
             return False
 
         logger.debug("Prometheus config changed")
 
-        container.push(PROMETHEUS_CONFIG, yaml.safe_dump(prometheus_config), make_dirs=True)
+        push(container, PROMETHEUS_CONFIG, yaml.safe_dump(prometheus_config))
         for filename, contents in certs.items():
-            container.push(filename, contents, make_dirs=True)
+            push(container, filename, contents)
 
-        self._stored.config_hash = config_hash
+        push(container, CONFIG_HASH_PATH, config_hash)
         logger.info("Pushed new configuration")
         return True
 
@@ -757,4 +766,4 @@ class PrometheusCharm(CharmBase):
 
 
 if __name__ == "__main__":
-    main(PrometheusCharm, use_juju_for_storage=True)
+    main(PrometheusCharm)
--- a/src/charm.py
+++ b/src/charm.py
@@ -82,14 +82,14 @@ def pull(container, path, *, default: Optional[str] = None) -> Optional[str]:
         File contents if exists; `default` otherwise.
     """
     try:
-        return container.pull(path, encoding="utf8").read()
-    except FileNotFoundError:
+        return container.pull(path, encoding="utf-8").read()
+    except (FileNotFoundError, PebbleError):
         return default
 
 
 def push(container, path, contents):
     """Push file to container, creating subdirs as necessary."""
-    container.push(path, contents, make_dirs=True, encoding="utf8")
+    container.push(path, contents, make_dirs=True, encoding="utf-8")
 
 
 class PrometheusCharm(CharmBase):
--- a/src/charm.py
+++ b/src/charm.py
@@ -453,13 +453,13 @@ class PrometheusCharm(CharmBase):
         metrics_consumer_alerts = self.metrics_consumer.alerts()
         remote_write_alerts = self.remote_write_provider.alerts()
         alerts_hash = sha256(str(metrics_consumer_alerts) + str(remote_write_alerts))
-        alert_rules_changed = alerts_hash != self.pull(ALERTS_HASH_PATH)
+        alert_rules_changed = alerts_hash != self._pull(ALERTS_HASH_PATH)
 
         if alert_rules_changed:
             self.container.remove_path(RULES_DIR, recursive=True)
             self._push_alert_rules(metrics_consumer_alerts)
             self._push_alert_rules(remote_write_alerts)
-            self.push(ALERTS_HASH_PATH, alerts_hash)
+            self._push(ALERTS_HASH_PATH, alerts_hash)
 
         return alert_rules_changed
 
@@ -476,7 +476,7 @@ class PrometheusCharm(CharmBase):
 
             rules = yaml.safe_dump(rules_file)
 
-            self.push(path, rules)
+            self._push(path, rules)
             logger.debug("Updated alert rules file %s", filename)
 
     def _generate_command(self) -> str:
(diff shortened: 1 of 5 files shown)


src/charm.py
```diff
@@ -74,16 +75,29 @@ def sha256(hashable) -> str:
...
+    Returns:
+        File contents if exists; `default` otherwise.
+    """
+    try:
+        return container.pull(path, encoding="utf8").read()
+    except FileNotFoundError:
+        return default
+
+
+def push(container, path, contents):
```
rbarry82, line 90: Why was this moved to the top level? Nothing else will use it, it forces readers/authors to jump around the code, consider whether or not this is doing something different than `ops.model.Container.push` (it is not), and couldn't meaningfully be used from elsewhere without constructing a [`Container`](https://github.com/canonical/operator/blob/main/ops/model.py#L1748) object by hand, which expects a `_ModelBackend` and to be able to find either an `ops.pebble.Client` object or the socket anyway.

src/charm.py
```diff
@@ -74,16 +75,29 @@ def sha256(hashable) -> str:
     return hashlib.sha256(hashable).hexdigest()
 
 
+def pull(container, path, *, default: Optional[str] = None) -> Optional[str]:
+    """Pull file from container (without raising FileNotFoundError).
```
rbarry82, line 79: Other than `default` never being used (and its usage probably breaking the logic of the other places this is called anyway), this has essentially the same question as "why was `push` moved to the top level"? 

In the worst case, this could be a class-level method named `try_pull` or `get_file_contents` with something like:

```python
def get_file_contents(self, path) -> Optional[str]:
  try:
    return self.container.pull(path).read()
  except pebble.PathError:
    logger.debug(...)
    return None
```

Or use `contextlib.suppress`. But Pebble should be returning a `PathError` if it doesn't exist (there's [a test for that, too](https://github.com/canonical/operator/blob/7faa5a2451e4531492d4513feaea0617512f6e9d/test/test_pebble.py#L1893), but the `client` is a Mock, and if Pebble doesn't really do that at runtime, we should raise it with Ben). A `FileNotFoundError` means it's getting [here](https://github.com/canonical/operator/blob/main/ops/pebble.py#L2449), and Pebble [should be raising before it gets there](https://github.com/canonical/operator/blob/7faa5a2451e4531492d4513feaea0617512f6e9d/ops/pebble.py#L1824).

All of that aside, extracting these methods makes the flow of the code much harder for no clear benefit

sed-i: Yes, I'm not too excited about those free functions myself.
Can easily change but here's my thinking.

1. Good point about default. Removed.
2. Thanks for the PathError tip, https://github.com/canonical/operator/issues/896.
3. Currently we only pull hash files. It will always be `utf-8` and I think we always want None back. Repeating a contextlib is fine but like this it's less distractions while reading the main code.
4. I don't think the "jumping around" argument is valid - we jump around all the time everywhere. Also in our charming context it's pretty obvious what "pull" and "push" do.
5. I do not see justification to make it a method (nor a staticmethod) (we have lots of such "methods" already but it's not a reason to continue doing so). What's wrong with good old free functions?
6. `try_pull` or `pull_or_none` are fine. Just `pull` also seems fine.

src/charm.py
sed-i: Unlike with `pull`, the arguments here are indeed weaker:

- Whenever we push here, we always use `make_dirs=True`.
- For symmetry with the `pull`.

src/charm.py
sed-i: Converted pull, push to methods ([diff](https://github.com/canonical/prometheus-k8s-operator/pull/431/commits/2b5da290969e5e297dad9d0933e99a5538f4933c)).

src/charm.py
```diff
@@ -63,6 +62,8 @@
 
 PROMETHEUS_CONFIG = "/etc/prometheus/prometheus.yml"
 RULES_DIR = "/etc/prometheus/rules"
+CONFIG_HASH_PATH = "/etc/prometheus/config.hash"
+ALERTS_HASH_PATH = "/etc/prometheus/alerts.hash"
```
simskij, line 66: File extension should likely be `sha256` to conform to common practices.

sed-i: Being specific here with "sha256" is decoupled from the choice of hashing function below, so we would need to remember to update one with the other.
Changed in next commit.

src/charm.py
rbarry82: I don't think we should use `pull` either...

src/charm.py
```diff
@@ -772,6 +743,22 @@ def _prometheus_version(self) -> Optional[str]:
...
+        try:
+            return self.container.pull(path, encoding="utf-8").read()
+        except (FileNotFoundError, PebbleError):
+            # Drop FileNotFoundError https://github.com/canonical/operator/issues/896
+            return None
+
+    def push(self, path, contents):
+        """Push file to container, creating subdirs as necessary."""
+        self.container.push(path, contents, make_dirs=True, encoding="utf-8")
+
```
Abuelodelanada, line 761: These two methods seems private, I would name them: `_push` and `_pull`

Abuelodelanada (approved):

src/charm.py
rbarry82: Hopefully the OF bug will make it nicer. It looks like it's missing some change in teh return from the Pebble API.

Part of the thing here is around scoping and least responsibility. It's not always easy to get used to a strange/new codebase, and despite the fact that we do "jump around" (during larger patches/refactors, I do at least try to shuffle similar methods near each other, so all of the "alert" stuff is near the "alert" stuff), but somewhat the problem is that it is _not_ obvious what "push" and "pull" do. 

For any other author used to reading/writing charm code, or looking at the docs, `push` and `pull` here do not behave like `ops.model.Container.[push|pull]`. The behavior is different, even if only slightly, rather than having a method which is a tightly scoped as possible (for the case of "what if a hash value is empty?") and named to reflect that. 

_Currently_ we only pull hash files, but that may change in the future, and when/if it does, we must contend with either a "`push|pull` implementation which is specific to this file and not like the rest of the ecosystem", using this `push|pull` for hash files and the "normal" `push|pull` elsewhere, the author who writes the patch to "pull" whatever files are needed then refactoring this out (since `ops` already does it), and so on.

The argument to make it a method is essentially twofold: that the "domain" which owns a container is the instance of the charm, and that the only real "consumer" of that method is also the charm. This somewhat applies to making things `@staticmethod` also. We are now at a point where external authors are reading our code, and using it, and Hyrum's law applies -- every casual change is potentially one which we must live with until we come to a point in time where we _can_ break the API.

There's nothing wrong with top-level functions or `@staticmethod` if they're actually intended to be used by external users, or don't require peering inside objects, or are otherwise visible. Pulling out a method which is only used by one class to the top level (or making a method on a class which isn't generally reusable a `@staticmethod` because it doesn't actually mutate state and it's possible) doesn't end up being good encapsulation, and providing "free" methods which don't require an instance of the object is somewhat asking for them to be used.

Put differently, [this method](https://github.com/canonical/prometheus-k8s-operator/blob/c21baa67cbf7841860793441c602af9c59855650/lib/charms/prometheus_k8s/v0/prometheus_scrape.py#L455) is only called from a single place (in our code), from a private method. Whatever notion of idempotency or state machine we have inside our charms is often in controlling the "flow" of data and transformations from event->handler->read relation data->transform->ingest by workload. 

It's ok to have this where it is. It is idempotent, in that the "worst thing" that happens is pretty much `job_name` getting more strings tacked onto it. It's nice and clean. But it's also true that now we can't ever change it, because we don't know who may be using it, and if we have a need to do some other kind of manipulation, we'll need to put it as a private instance method we control anyway.  

There will, of course, be times to do this either way, and times where that tradeoff is acceptable, but I don't think that we should do it just because we can, because we _cannot_ take it back in those circumstances. Similarly, if we want to assume that every time we're looking for _hashes_ it should have some set of options (`make_dirs=True, encoding="utf8"`), we could make a `functools.partial` and apply whatever options, or narrow it to that scope. 

`push` and `pull` should not be restrictive enough that it's _always_ UTF-8 text, or `None` (and that the author won't want an exception to treat it like "normal" `ops.model.Container.pull`). When secret support lands, there _are_ some secret/authentication methods which are binary `.p12` by default, and this will just block them out. Or use "regular" `push()`, and now we have both paradigms in one charm with the same name, or teach this one to accept `*args, **kwargs` and forward them, or ...

This particular logic around "what to do if a file is not found, what particular options to push a hash with" could be cleanly isolated to this functionality.

I'm not trying to be too harsh or argumentative on the review here, but we need to be careful about making very large changes without a good reason

sed-i: The scope / "blast radius" of these functions is just the charm.
We have many methods that make assumptions that make them not as general as they could be.
And we modify those methods of ours per need.
These push/pull are no exception.
Taking Jose's recommendation to make them "private" with `_`.

src/charm.py
sed-i: Done.

src/charm.py
rbarry82: I'm not going to hold this patch up, but even if the scope/"blast radius" of these function is just the charm, it is still "now we have two almost identically-named ways to push/pull in the same charm codebase" (if someone writes a later patch which uses the "normal" push/pull) or "we have a push/pull which doesn't behave the same way as the 'standard' and requires special handling" rather than keeping it isolated to the one use case it's actually needed for and re-evaluating later.

Again, once something is public (top-level, `PrometheusConfig....`), we're kind of "locked in" to a public API, for better or worse, and we need to be far more careful with it.

I am not convinced that "we already have methods which make assumptions" is a good argument for adding more of them.

What is the downside to keeping these methods scoped to/named for their solitary function (hash push/pulling), whether public instance methods or private/

rbarry82: No, I don't think that `_pull` is public and binding at all (I meant that more in reference to the `PrometheusConfig...` methods). I'm more worried, for this change that if we get to needing [binaries from secrets](https://github.com/canonical/operator/issues/869) that there will be a confusing blend of `_pull` (with its inherent assumptions) and `self.charm.container.pull` (where we need/want binary).

We can defer that as a future case -- not blocking this patch, but it's something worth considering. Even if we don't have a use case right now, it's entirely plausible that we actually will have a use case or need for the "normal" `push|pull` interface in the future, and we may as well leave the door open for that (in all of our charms).

rbarry82 (approved):

src/charm.py
sed-i: Thanks for the clarification!
Agreed about secrets.

PR #433, opened 2023-01-21
Break the Rules monolith. Add support for recording rules.

### LINKED ISSUE #428: Add explicit support for recording rules
## Issue

Currently, Prometheus [recording rules](https://prometheus.io/docs/prometheus/latest/configuration/recording_rules/) are implicitly supported; if they are put in the `prometheus_alert_folder`, the library will pick them up and import them correctly.

However, following our discussion, there should be an explicit support for recording rules in a separate folder.

## Solution

I will sum up the points made in the discussion to serve as guidelines for the PR.

Ideally, we would refactor all mentions of *alert rules* to the more general *rules*, so that all the functions that currently handle *alert rules* would be suited to handle *recording rules* as well. This includes quite some documentation changes to reflect the new code.

The library should take two configurable paths, one for *alert rules* (i.e., `prometheus_alert_folder`) and one for *recording rules* (i.e., `prometheus_record_folder`). There's no need for explicit checking on the rule type; the goal is to encourage the splitting of different rule types in different folder, not to enforce their separation. However, we could emit a warning when a *recording rule* is found in the *alert rule* folder, and vice versa.

The documentation should only describe the **suggested behavior** of using the two different folders.

As per the `AlertRule` class itself, it was suggested to implement the following structure:
```python
class Rules(...):  # the old AlertRules class, after refactoring

class AlertRules(Rules):
  """Some docstring, but no unique methods needed yet."""

class RecordingRules(Rules):
  """Some docstring, but no unique methods needed yet."""
```
This class separation would allow us to implement specific behaviors for different rule types in the future, if we ever need to do so.

The library should then import rules from both folders and process them as it does now.

rbarry82:
## Issue
`prometheus_scrape` doesn't currently support recording rules via any intentional mechanism -- they can be accidentally loaded via the `AlertRules` pathway, but none of the log messages or databags will be meaningful. We should support this use case, separated.

## Solution
The actual handling of the rules is mostly identical, outside of keys to check for if we're validating single rule formats, and a string to use in log messages/alerts.

Make `AlertRules` into `Rules(ABC)` with an `@abstractmethod`+`@property` which returns a `Literal` `_RulesType`, which child classes implement to indicate the string to interpolate where needed. Similarly, extract the handling for `alerts()` into a more generalized method which takes the type as an argument in addition and maps it to a databag. All of the handling there is also identical (for now).

Remove every reference I could find to `alert` in docstrings, docs, and variable names where it should not be present, substituting the interpolated value or appending/amending docstrings to speak about both types. Rename `CosTool.validate_alert_rules()` to `CosTool.validate_rules()` (it isn't specific). Rename exceptions to more generic names.

Add constructor args and events for recording rules.

In this way, if/when we ever _do_ need different handling (such as if we do not want to support "single rule file" formats for recording rules, or Loki, or other) or the relation data needs/APIs for either type diverge, they are already cleanly separated without breaking the current API.

In addition, this restructuring makes it possible to, once pydeps for charmcraft is implemented, move the `[T]Rules` machinery elsewhere, rather than trying to keep `prometheus_scrape`, `prometheus_remote_write`, `loki_push_api`, and others in sync, with independent patches needed any time a bug or feature comes up in any of them (with a little handling to smash `metadata`  into `scrape_metadata`, even as just an optional `metadata_key` arg in the constructor for `Rules()`) 

Closes #428 

## Context
Also, `mock.create_autospec()` exists, and it's great. `_dedupe_...` was moved back into the class it belongs to (the only place that uses it), and `test_functions` uses `create_autospec` to test that function in an isolated way. We don't need to extract methods to the top level to make writing unit tests easier, and, as a common refrain, doing so makes new readers/authors or anyone writing a patch juggle more mental state while trying to keep track of whether it's _really_ doing something generalized, used elsewhere, etc.

The problem domain of `_dedupe_...` is in the only class which ever calls it, and `create_autospec()` makes unit testing it in an isolated fashion just as easy as importing it directly from the top-level. 

## Testing Instructions
Deploy. Make sure there aren't regressions.

## Release Notes
Add support for recording rules.


lib/charms/prometheus_k8s/v0/prometheus_remote_write.py
```diff
@@ -222,47 +234,47 @@ def _from_file(self, root_path: Path, file_path: Path) -> List[dict]:
...
+                logger.error("Failed to read rules from %s: %s", file_path.name, e)
                 return []
 
-            if _is_official_alert_rule_format(rule_file):
-                alert_groups = rule_file["groups"]
-            elif _is_single_alert_rule_format(rule_file):
-                # convert to list of alert groups
+            if _is_official_rule_format(rule_file):
+                groups = rule_file["groups"]
+            elif _is_single_rule_format(rule_file, self.rule_type):
```
sed-i, line 242: I may have missed a discussion on this: should we add a deprecation warning for the custom format?

This really feels like it should be an abstract method isn't it?

lib/charms/prometheus_k8s/v0/prometheus_remote_write.py
```diff
@@ -222,47 +234,47 @@ def _from_file(self, root_path: Path, file_path: Path) -> List[dict]:
...
-                            if label not in alert_rule["labels"]:
-                                alert_rule["labels"][label] = val
-                        # insert juju topology filters into a prometheus alert rule
-                        alert_rule["expr"] = self.tool.inject_label_matchers(
-                            re.sub(r"%%juju_topology%%,?", "", alert_rule["expr"]),
+                            if label not in rule["labels"]:
+                                rule["labels"][label] = val
+                        # insert juju topology filters into a prometheus rule
+                        rule["expr"] = self.tool.inject_label_matchers(
+                            re.sub(r"%%juju_topology%%,?", "", rule["expr"]),
```
sed-i, line 273: Should we emit a deprecation warning if we encounter the %% stub?

lib/charms/prometheus_k8s/v0/prometheus_remote_write.py
```diff
@@ -285,13 +297,13 @@ def _group_name(self, root_path: str, file_path: str, group_name: str) -> str:
         #  - name, from juju topology
         #  - suffix, from the relative path of the rule file;
         group_name_parts = [self.topology.identifier] if self.topology else []
-        group_name_parts.extend([rel_path, group_name, "alerts"])
+        group_name_parts.extend([rel_path, group_name, f"{self.rule_type}s"])
```
sed-i, line 300: If we use the official terms "alerting" and "recording" then could drop the plural `s` suffix.

lib/charms/prometheus_k8s/v0/prometheus_remote_write.py
```diff
@@ -285,13 +297,13 @@ def _group_name(self, root_path: str, file_path: str, group_name: str) -> str:
...
         # filter to remove empty strings
         return "_".join(filter(None, group_name_parts))
 
     def _is_already_modified(self, name: str) -> bool:
         """Detect whether a group name has already been modified with juju topology."""
-        modified_matcher = re.compile(r"^.*?_[\da-f]{8}_.*?alerts$")
+        modified_matcher = re.compile(r"^.*?_[\da-f]{8}_.*?(records|alerts)$")
```
sed-i, line 306: What if we get an alert group originally called "too_many_records"?

lib/charms/prometheus_k8s/v0/prometheus_remote_write.py
```diff
@@ -354,23 +366,57 @@ def add_path(self, path: str, *, recursive: bool = False) -> None:
...
+    The official Prometheus format is a YAML file conforming to the Prometheus documentation
+    (https://prometheus.io/docs/prometheus/latest/configuration/alerting_rules/).
+    The custom single rule format is a subsection of the official YAML, having a single alert
+    rule, effectively "one alert per file".
+    """
+
+    _rule_type = "alert"  # type: _RuleType
+
+    @property
+    def rule_type(self) -> _RuleType:
```
sed-i, line 400: Probably wouldn't need a property if we use ClassVar?

lib/charms/prometheus_k8s/v0/prometheus_remote_write.py
```diff
@@ -354,23 +366,57 @@ def add_path(self, path: str, *, recursive: bool = False) -> None:
...
+        return {"groups": self.groups} if self.groups else {}
+
+
+class AlertRules(Rules):
+    """Utility class for amalgamating prometheus alert files and injecting juju topology.
+
+    The official Prometheus format is a YAML file conforming to the Prometheus documentation
+    (https://prometheus.io/docs/prometheus/latest/configuration/alerting_rules/).
+    The custom single rule format is a subsection of the official YAML, having a single alert
+    rule, effectively "one alert per file".
```
sed-i, line 394:
```suggestion
    rule, effectively "one alerting rule per file".
```

lib/charms/prometheus_k8s/v0/prometheus_remote_write.py
```diff
@@ -354,23 +366,57 @@ def add_path(self, path: str, *, recursive: bool = False) -> None:
...
+        return self._rule_type
+
+
+class RecordingRules(Rules):
+    """Utility class for amalgamating prometheus recording files and injecting juju topology.
+
+    The official Prometheus format is a YAML file conforming to the Prometheus documentation
+    (https://prometheus.io/docs/prometheus/latest/configuration/recording_rules/).
+    The custom single rule format is a subsection of the official YAML, having a single recording
+    rule, effectively "one record per file".
```
sed-i, line 411:
```suggestion
    rule, effectively "one recording rule per file".
```

lib/charms/prometheus_k8s/v0/prometheus_remote_write.py
```diff
@@ -668,28 +732,39 @@ def _handle_endpoints_changed(self, event: RelationEvent) -> None:
...
         alert_rules.add_path(self._alert_rules_path, recursive=False)
-
         alert_rules_as_dict = alert_rules.as_dict()
 
+        recording_rules = RecordingRules(self.topology)
+        recording_rules.add_path(self._recording_rules_path, recursive=False)
+        recording_rules_as_dict = recording_rules.as_dict()
+
         if alert_rules_as_dict:
             relation.data[self._charm.app]["alert_rules"] = json.dumps(alert_rules_as_dict)
```
sed-i, line 755:
```suggestion
            # TODO in v1, rename rel data key to "alerting_rules"
            relation.data[self._charm.app]["alert_rules"] = json.dumps(alert_rules_as_dict)
```

lib/charms/prometheus_k8s/v0/prometheus_remote_write.py
```diff
@@ -222,47 +234,47 @@ def _from_file(self, root_path: Path, file_path: Path) -> List[dict]:
...
             except Exception as e:
-                logger.error("Failed to read alert rules from %s: %s", file_path.name, e)
+                logger.error("Failed to read rules from %s: %s", file_path.name, e)
                 return []
 
-            if _is_official_alert_rule_format(rule_file):
-                alert_groups = rule_file["groups"]
-            elif _is_single_alert_rule_format(rule_file):
-                # convert to list of alert groups
+            if _is_official_rule_format(rule_file):
```
sed-i, line 240: This would evaluate as True for both alerting and recording groups. And also for mixed groups.
For single rule format though, it would ignore the other type and emit an error.
We need some spec around this. Add some utests?

E.g.

```text
# GIVEN a folder of alert rules
# WHEN parsed as alert rules
# THEN all alert rules are present in the `parsed`

# WHEN parsed as recording rules
# THEN a warning is emitted for every file
# AND `parsed` is empty

# GIVEN a mixed folder
# WHEN parsed as alert rules
# THEN all alert rules are present in the `parsed`
# AND all recording rules are omitted
```

sed-i:
The plot thickens?
[Federated rule groups](https://grafana.com/docs/mimir/latest/operators-guide/architecture/components/ruler/#federated-rule-groups):

```yaml
name: MyGroupName
source_tenants: ["tenant-a", "tenant-b"]
rules:
  - record: sum:metric
    expr: sum(metric)
```

rbarry82:
> Federated rule groups

Probably only if we ever get to the point of multi-tenancy, but at least `cos-tool` already supports this, and it's nominally easier to add (excluding Prometheus itself) if we get there.

lib/charms/prometheus_k8s/v0/prometheus_remote_write.py
```diff
@@ -36,6 +37,7 @@
 )
 from ops.framework import EventBase, EventSource, Object, ObjectEvents
 from ops.model import Relation
+from typing_extensions import Literal
```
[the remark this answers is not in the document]
rbarry82: Probably. I'll re-check...

lib/charms/prometheus_k8s/v0/prometheus_remote_write.py
```diff
@@ -56,6 +58,9 @@
 RELATION_INTERFACE_NAME = "prometheus_remote_write"
 
 DEFAULT_ALERT_RULES_RELATIVE_PATH = "./src/prometheus_alert_rules"
```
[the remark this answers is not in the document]
rbarry82: Granted, it's just very late in the game to change that now.

lib/charms/prometheus_k8s/v0/prometheus_remote_write.py
```diff
@@ -56,6 +58,9 @@
...
+DEFAULT_RECORDING_RULES_RELATIVE_PATH = "./src/prometheus_recording_rules"
+
+_RuleType = Literal["alert", "record"]
```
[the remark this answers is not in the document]
rbarry82: Well, sort-of only internally. It's a `Literal` so the log messages and semantics of the rest of the class stay as similar as possible. "Limiting blast radius". If we want `StrEnum`, it's a small refactor after this is merged, but I'm trying to get a sense of "does everything think this is an ok strategy" before it starts accumulating riders and extra stuff.

lib/charms/prometheus_k8s/v0/prometheus_remote_write.py
```diff
@@ -354,23 +366,57 @@ def add_path(self, path: str, *, recursive: bool = False) -> None:
...
+class AlertRules(Rules):
+    """Utility class for amalgamating prometheus alert files and injecting juju topology.
+
+    The official Prometheus format is a YAML file conforming to the Prometheus documentation
+    (https://prometheus.io/docs/prometheus/latest/configuration/alerting_rules/).
+    The custom single rule format is a subsection of the official YAML, having a single alert
+    rule, effectively "one alert per file".
+    """
+
+    _rule_type = "alert"  # type: _RuleType
```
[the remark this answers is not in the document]
rbarry82: I mean, sure, but not really semantically that different, unless we're expecting anyone to override a private class variable.

lib/charms/prometheus_k8s/v0/prometheus_remote_write.py
rbarry82: Well, it would, because there's no similar ClassVar on the parent, so `self.rule_type` won't resolve on `AlertRules(ABC)` without at least an `@abstractmethod`, hence the property. Parents can't use the attributes of their children unless they're defined, and the ABC parent has no reason to have a `_rule_type`

lib/charms/prometheus_k8s/v0/prometheus_remote_write.py
rbarry82: What if we get an alert group called `too_many_deadbeef_alerts` right now? Same breakage.

lib/charms/prometheus_k8s/v0/prometheus_remote_write.py
rbarry82: Yes, but then we change the API/contract. We can make the argument that "it's only our code", but it will still potentially meaningfully change the runtime behavior of charms.

That would need to wait for v1

lib/charms/prometheus_k8s/v0/prometheus_remote_write.py
rbarry82: Not if there's any hope of extracting this into a "unified" set of operations via this kind of abstraction, rather than scrape, remote-write, loki, and everything else having their own 98% compatible implementations.

Loki needs this (not replaced with `""`, but there's nothing we can deprecate it for at the moment)

lib/charms/prometheus_k8s/v0/prometheus_remote_write.py
rbarry82: Are we planning to deprecate it? AFAIK, no. At least nothing firm. @simskij likes it, anyway.

Since there's no _real_ difference in parsing alerts/recording rules (other than "no annotations" and what's in the YAML, neither of which we're enforcing right now), there's no reason to make this an abstract method where both concrete implementations would be identical.

The point is rather to avoid that, but if we feel like there's some reason to in the future, that we _can_, from a single, unified place.

This patch takes a middle ground between the discussion of "mix-n-match rules" (which will be hard or impossible to untangle in relation data, charm filesystems, and so on later) and separation, but saying "the actual _handling_ of the rules is more or less identical now, so let's make the code variadic enough to give the right messages back and not change filenames, but otherwise keep the same logic"). 

From that point of view, it's not clear at all where concrete implementations of abstract methods would differ, or why.

lib/charms/prometheus_k8s/v0/prometheus_remote_write.py
rbarry82: This makes the presumption that we are _not_ going to support "single rule per file recording rules", which is not decided upon. 
 
 No unit tests are needed here unless we come to a consensus that "no, we only want to support the official recording rules format". Otherwise, just a couple of unit tests for "yep, recording rules also pass."
 
 The entire original discussion around "just let users put them in the same path" is orthogonal to differing formats, and "document different paths, different default paths in the constructor, but if some user _really wants_ to jam them all in one place, we'll try to give them best effort". 
 
 It would be a muddled relation data key unless we _actually_ parsed them, and that would involve either some `MixedAlertRules` class, the return sifting through nested keys, or other -- which it's kind of "best effort, but you shouldn't".
 
 Failing that, there doesn't seem to be any real agreement about dropping single-rule-per-file formats, that we should actually validate any of them (that recording rules do not have annotations, that alert rules/recording rules have the right type, and so on, outside of the very small assertion present in `_is_single_rule_format`, which wouldn't have worked with recording rules all in the same bag _anyway_.
 
 In short, this PR/draft was sort-of "put your money where your mouth is" based on what I chatted about for separation.
 
 If we want to have a longer conversation about whether or not we should support single rules or not as a spec document, deprecating old versions, changing databag keys (API break), changing formatting (potential API break), we can, but this PR isn't the right place for that conversation. PR comments just aren't suited for it. 
 
 It will also, inevitably, delay the implementation of Mimir self-monitoring, which seems to rely on some of the recording rules. That's not a blocker, but it's a consideration.
 
 A general sense of "yes, this pattern is fine, let's address any glaring issues and have a follow-up to waffle about what `typing.Whatever` some annotation is", or "no, I think we actually need to have a spec discussion", or "what about this entirely different option?" is what I was hoping for.

dstathis (approved):
This looks good to me. What do people think about making a rules lib once pydeps is a thing?

lib/charms/prometheus_k8s/v0/prometheus_remote_write.py
sed-i: I'm a bit confused by your response. Afaict,
1. The current code has a bug: different combos have inconsistent outcome
2. We need utests around this

lucabello (approved):
Looks good; make sure the CI passes before merging! :)

lib/charms/prometheus_k8s/v0/prometheus_remote_write.py
rbarry82: The current code does not have a bug. Yet. 

Let me try again:

When we discussed the need for recording rules for Mimir monitoring, there was a discussion about whether we should just go with the solution of mixing them in the same folder/databag, because prom/Mimir allow this. 

I thought we should separate them in the API and databags. After some discussion on mattermost, we proposed different folders in the docs, examples, and constructor, but backend code mostly shared (with a very small child class for each) to give us flexibility in case we need to change, but that we would not explicitly block users who wanted to mix them and set the same path for both, or mixed rules in a folder. 

This is that PR. The inconsistent outcome is consistent with the "best effort support user override" compromise, and the suggested unit tests would forbid/block that. 

This PR could be a concrete/final implementation (after unit tests for recording rules themselves are added), but it was created at least in part for discussion about "this is how my proposed solution would look in real code."

Before raising inconsistencies, we first need to agree on whether this is actually a compromise we can all live with. Then whether or not we actually DO want to forbid mixing, and add logic. It's undefined behavior right now more than a bug, because we have no defined it. 

What are your thoughts on it _overall_? Yes? No? Change some part of the proposed compromise? Agree to formalize some set of behavior?

rbarry82:
Closing to move this to `cosl` instead, with unification between all the variants (mostly the same as this patch, just small tweaks to meet in the middle for Prom+Loki)

PR #440, opened 2023-01-26
Send events from `PrometheusRemoteWriteProvider`

rbarry82:
## Issue
Currently, `PrometheusRemoteWriteProvider`, unlike the rest of our charms, does not send any events at all when the status changes.

It seems that the Prometheus charm listens/listened _directly_ to events from that interface, and called a "common exit hook" which masked the issue by always re-reading alerts on every event.

## Solution
Send events which can be watched, so charms and libraries keep their appropriate scope of responsibility. Otherwise, this library behaves differently from all of our others, and it's easy to "lose" or "drop" data.

## Context
See the comments for Mimir [here](https://github.com/canonical/mimir-k8s-operator/pull/15).

## Testing Instructions
Deploy, listen to events, make sure things fire. Unit tests and integration tests were not changed, as "remove the raw listeners and they pass with the other event" is a reasonable assurance that this change did not break code/functionality. 

Authors/users who were previously directly monitoring the relation still can, but now they don't have to.

## Release Notes
Send events from `PrometheusRemoteWriteProvider`


lib/charms/prometheus_k8s/v0/prometheus_remote_write.py
```diff
@@ -810,15 +839,25 @@ def __init__(
         on_relation = self._charm.on[self._relation_name]
         self.framework.observe(
             on_relation.relation_created,
-            self._on_relation_change,
+            self._on_consumers_changed,
```
sed-i, line 842: I do not recall why we need to observe both `created` and `joined`. Shouldn't `joined` be enough?

lib/charms/prometheus_k8s/v0/prometheus_remote_write.py
```diff
@@ -810,15 +839,25 @@ def __init__(
...
         )
+        self.framework.observe(
+            on_relation.relation_changed,
+            self._on_relation_changed,
+        )
+
+    def _on_consumers_changed(self, event: RelationEvent) -> None:
+        if not isinstance(event, RelationBrokenEvent):
+            self.update_endpoint(event.relation)
+        self.on.consumers_changed.emit()
```
sed-i, line 856: Why no relation id here?

lib/charms/prometheus_k8s/v0/prometheus_remote_write.py
```diff
@@ -721,6 +721,33 @@ def endpoints(self) -> List[Dict[str, str]]:
...
+    def snapshot(self):
+        """Save Prometheus remote_write information."""
+        return {"relation_id": self.relation_id}
+
+    def restore(self, snapshot):
+        """Restore Prometheus remote_write information."""
+        self.relation_id = snapshot["relation_id"]
+
+
+class PrometheusRemoteWriteProviderConsumersChangedEvent(EventBase):
```
sed-i, line 740: I'm not sure this is needed for lib users.
Internally we can still do `update_endpoint` but emit `rules_changed`, because for a lib user the implication is probably the same?
I.e. less custom events to observe the better.

lib/charms/prometheus_k8s/v0/prometheus_remote_write.py
```diff
@@ -721,6 +721,33 @@ def endpoints(self) -> List[Dict[str, str]]:
         return endpoints
 
 
+class PrometheusRemoteWriteAlertsChangedEvent(EventBase):
```
sed-i, line 724: Probably should update the very long module docstring of the lib?

lib/charms/prometheus_k8s/v0/prometheus_remote_write.py
rbarry82: Largely that Prometheus itself [needs to know in order to enable remote write](https://github.com/canonical/prometheus-k8s-operator/blob/main/src/charm.py#L500). I would also think `joined` would be enough, but... we don't know what we don't know, and I'm happier leaving it exactly as it was for an "emergency" fix which we can evaluate/remove later.

lib/charms/prometheus_k8s/v0/prometheus_remote_write.py
rbarry82: The relation id honestly isn't used anywhere else either. In similar API changes, it probably could/should be, so Prometheus/Mimir can be more granular about what alerts they keep/delete, but first we'd need either of them to actually start tracking which alerts correlate to which relations, and have some `alerts(self, relation_id: Optional[int] = None)` call to suss it out.

For just the binary "should remote write be enabled in prom? y/n?" the relation id would be meaningless anyway.

lib/charms/prometheus_k8s/v0/prometheus_remote_write.py
rbarry82: This is needed for the same reason as above. Prometheus (via `_configure`) must know to actually check whether it has any relations to enable remote write or not.

Without this, Prometheus would not know (and would not enable it, or disable it) until some other event kicked `_configure`

lib/charms/prometheus_k8s/v0/prometheus_remote_write.py
```diff
@@ -721,6 +721,33 @@ def endpoints(self) -> List[Dict[str, str]]:
...
+
+
+class PrometheusRemoteWriteProviderConsumersChangedEvent(EventBase):
+    """Event emitted when Prometheus remote_write alerts change."""
+
+
+class PrometheusRemoteWriteProviderEvents(ObjectEvents):
+    """Event descriptor for events raised by `PrometheusRemoteWriteProvider`."""
+
+    alert_rules_changed = EventSource(PrometheusRemoteWriteAlertsChangedEvent)
```
sed-i, line 747: Should these be renamed to
```diff
-alert_rules_changed
+rules_changed
```
?

PR #441, opened 2023-01-27
validate incoming scrape jobs

### LINKED ISSUE #314: Validate scrape configs/targets from incoming relations
The addition of a config validation action provides a good way to validate a complete Prometheus configuration, but it is only useful as a all-or-nothing check, where a potentially invalid configuration must either not be pushed to the workload, which may also miss other changes in relations, or to apply it and block Prometheus entirely, forcing the use of the action.

In addition, on an upgrade, pod restart, or other scenario whether the workload container is not running, it is not possible to pre-preemptively validate the configuration, since `promtool` will not be accessible.

Similar to `cos-tool` validating alerts, we could extract the portions of `promtool` used to evaluate rules/targets and validate them in an isolated fashion, sending events back as appropriate (again, similar to the alert rule validation).

dstathis:
## Issue
Fixes #314

## Solution
Create a minimal Prometheus config and check it with cos-tool.

## Testing Instructions
Provide an invalid scrape config and check that it does not get written to disk.

## Release Notes
Prometheus-k8s now validates incoming scrape jobs


lib/charms/prometheus_k8s/v0/prometheus_scrape.py
```diff
@@ -2377,6 +2380,19 @@ def validate_alert_rules(self, rules: dict) -> Tuple[bool, str]:
...
+        """Validate scrape jobs using cos-tool."""
+        conf = {"scrape_configs": jobs}
+        with tempfile.NamedTemporaryFile() as tmpfile:
+            with open(tmpfile.name, "w") as f:
+                f.write(yaml.safe_dump(conf))
+            try:
+                self._exec([str(self.path), "validate-config", tmpfile.name])
+            except subprocess.CalledProcessError as e:
+                logger.error("Validating scrape jobs failed: {}".format(e.output))
+                return False
```
sed-i, line 2393: 1. I wonder if this should raise instead
2. (Haven't looked at the context) what if cos-tool isn't present? Do we lose all scrape jobs?

lib/charms/prometheus_k8s/v0/prometheus_scrape.py
```diff
@@ -1123,6 +1123,9 @@ def jobs(self) -> list:
 
         scrape_jobs = _dedupe_job_names(scrape_jobs)
 
+        if not self._tool.validate_scrape_jobs(scrape_jobs):
+            return []
```
sed-i, line 1127: Not sure we want it to silently fail. I would add a `WARNING`.
This is another "juju model" vs "best effort" question.
I think we should emit a custom event the charm observes and blocks.

(Having status per relation would be handy here.)

dstathis: We log an error message (from within `validate_scrape_jobs`).

dstathis: Not sure why a custom event would be useful here. What would we do with it?

lib/charms/prometheus_k8s/v0/prometheus_scrape.py
dstathis: Yes we would lose all scrape jobs. Fixing now.

dstathis: fixed

lib/charms/prometheus_k8s/v0/prometheus_scrape.py
rbarry82: I don't even know if this is "juju model" vs "best effort". We can do both.

If this gets bumped up inside the loop, we could do almost exactly the kind of validation we [already do with alerts](https://github.com/canonical/prometheus-k8s-operator/blob/main/lib/charms/prometheus_k8s/v0/prometheus_scrape.py#L1199) (and Grafana dashboards, and Loki alerts). 

```python
if static_scrape_jobs:
    if not self._tool.validate_scrape_jobs(static_scrape_jobs):
      # set a message back in relation data so the Provider can see it, and it's isolated per relation
```

This keeps the information in the model (at least the validity/invalidity), the Provider can catch it in [`_on_relation_changed`](https://github.com/canonical/prometheus-k8s-operator/blob/main/lib/charms/prometheus_k8s/v0/prometheus_scrape.py#L1598) (arguably, this code won't even have to change), and block from the Provider side only.

Then we exclude it from the total list (don't extend scrape jobs with bad ones)

lib/charms/prometheus_k8s/v0/prometheus_scrape.py
rbarry82: We don't raise in other places that we [do validation either](https://github.com/canonical/prometheus-k8s-operator/blob/main/lib/charms/prometheus_k8s/v0/prometheus_scrape.py#L2381). Whether or not we should/could raise is a question, but raising an exception back to be caught... wherever is simply re-shuffling the logic in the caller from `if/else` to `try/except` and moving the `logger.error` here to there, for an exception which nothing can really meaningfully be done about it anyway.

What advantages would raising have?

lib/charms/prometheus_k8s/v0/prometheus_scrape.py
dstathis: I am aware that we do this for alert rules and I'm not sure it was actually the best design. The reason I am skeptical of this approach is at the moment, as far as I can tell, *none* of our charms observe `alert_rule_status_changed`. I am sort of struggling to see what the real life use case for this would be.

lib/charms/prometheus_k8s/v0/prometheus_scrape.py
dstathis: I don't think raising makes sense here. We certainly don't want to enter error state. Avoiding that is the whole point of this change. Raising and catching it elsewhere, as Ryan said, more or less doesn't do anything.

lib/charms/prometheus_k8s/v0/prometheus_scrape.py
rbarry82: None of _our_ charms do (`cos-config` really should), but the real life use case is reasonably simple. In a large deployment, or one with gitops, or `cos-proxy`, or other, specifying with some level of granularity to a particular provider that "you gave me bad alert rules/dashboards/scrape targets" is far more specific than blocking Prom/Loki/Grafana and hoping admins can sift through a giant bucket of stuff to figure out what's wrong and why, doesn't block the consumer, doesn't leave it with empty rules, etc.

Even if they don't actively listen to the event, `juju show-unit|application` will tell admins why a rule/dashboard/target from `foo/0` isn't visible.

Conversely, in the current patch, pointing `cos-config` to a git repo with bad scrape targets would cause Prometheus to drop _all_ scrape jobs. Just because none of our charms currently listen to this does not mean that the design a problem.

If you are bulk loading data into a database, going through a multi-select in an application/website, or anything else, being as specific as possible about which data is invalid and why is good UX.

tests/integration/test_check_config.py
```diff
@@ -68,30 +69,21 @@ async def test_good_config_validates_successfully(
...
-    await ops_test.model.wait_for_idle(apps=[scrape_shim, bad_scrape_tester])
-
-    await asyncio.gather(
-        ops_test.model.add_relation(
-            f"{bad_scrape_tester}:metrics-endpoint", f"{scrape_shim}:configurable-scrape-jobs"
-        ),
-        ops_test.model.add_relation(
-            f"{prometheus_app_name}:metrics-endpoint", f"{scrape_shim}:metrics-endpoint"
-        ),
+    await ops_test.model.wait_for_idle(apps=[prometheus_app_name])
```
PietroPasotti, line 72: is it being deployed in a different test (file)?
Are we not doing 'one model per test file/module' anymore? We should.

tests/unit/test_endpoint_consumer.py
```diff
@@ -503,6 +503,29 @@ def test_consumer_accepts_rules_with_no_identifier(self):
...
+
+        rel_id = self.harness.add_relation(RELATION_NAME, "consumer")
+        self.harness.add_relation_unit(rel_id, "consumer/0")
+        self.harness.update_relation_data(rel_id, "consumer", app_data)
+        jobs = self.harness.charm.prometheus_consumer.jobs()
+        self.assertTrue(len(jobs) == 0)
+        app_data = json.loads(
+            self.harness.get_relation_data(rel_id, self.harness.charm.app.name).get("event", "{}")
+        )
+        self.assertIn("scrape_job_errors", app_data)
```
PietroPasotti, line 527: ugh, who invented python's `unittest`

PietroPasotti (approved):
Just a generic clarification, looks good!

tests/integration/test_check_config.py
dstathis: It's being deployed as a fixture I believe. Still one model per file.

lib/charms/prometheus_k8s/v0/prometheus_scrape.py
```diff
@@ -121,7 +121,7 @@ def __init__(self, *args):
             {
```
sed-i, line 121: Bump LIBPATCH?

lib/charms/prometheus_k8s/v0/prometheus_scrape.py
```diff
@@ -1119,7 +1136,18 @@ def jobs(self) -> list:
         for relation in self._charm.model.relations[self._relation_name]:
             static_scrape_jobs = self._static_scrape_config(relation)
             if static_scrape_jobs:
-                scrape_jobs.extend(static_scrape_jobs)
+                # Duplicate job names will cause validate_scrape_jobs to fail.
+                # Therefore we need to dedupe here and after all jobs are collected/
```
sed-i, line 1140: Stray `/` at the end?

lib/charms/prometheus_k8s/v0/prometheus_scrape.py
```diff
@@ -2377,6 +2412,22 @@ def validate_alert_rules(self, rules: dict) -> Tuple[bool, str]:
...
+            return True
+        conf = {"scrape_configs": jobs}
+        with tempfile.NamedTemporaryFile() as tmpfile:
+            with open(tmpfile.name, "w") as f:
+                f.write(yaml.safe_dump(conf))
+            try:
+                self._exec([str(self.path), "validate-config", tmpfile.name])
+            except subprocess.CalledProcessError as e:
+                logger.error("Validating scrape jobs failed: {}".format(e.output))
+                raise
```
sed-i, line 2428: How about:
```suggestion
                msg = "Validating scrape jobs failed: {}".format(e.output)
                logger.error(msg)
                raise ValueError(msg) from e
```

lib/charms/prometheus_k8s/v0/prometheus_scrape.py
dstathis: Done.

lib/charms/prometheus_k8s/v0/prometheus_scrape.py
dstathis: Done.

lib/charms/prometheus_k8s/v0/prometheus_scrape.py
dstathis: Is there a particular reason for this? Just to change the exception type?

sed-i: To display that same error message as part of the traceback.
Not sure if the original exceptions coming out of _exec would print out the contents of `e.output`.

rbarry82: They don't, but they're also not used anywhere other than immediately above this. The exception here is used for flow control, and `CalledProcessError` is explicitly caught (and not logged) as, essentially, the "else" statement.

It is not useful here to raise a different error or even surface the traceback

PR #446, opened 2023-02-01
Use a ROCK image

rbarry82:
## Issue
Use a ROCK image from GHCR instead of Launchpad

### DIFF
--- a/README.md
+++ b/README.md
@@ -202,4 +202,4 @@ juju relate \
 ## OCI Images
 
 This charm by default uses the latest version of the
-[ubuntu/prometheus](https://hub.docker.com/r/ubuntu/prometheus) image.
+[canonical/prometheus](https://ghcr.io/canonical/prometheus) image.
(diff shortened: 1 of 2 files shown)


lucabello (approved):

README.md
```diff
@@ -202,4 +202,4 @@ juju relate \
 ## OCI Images
 
 This charm by default uses the latest version of the
-[ubuntu/prometheus](https://hub.docker.com/r/ubuntu/prometheus) image.
+[canonical/prometheus](https://ghcr.io/canonical/prometheus) image.
```
sed-i, line 205: Is it "ok" to just say `canonical/prometheus` when it's not on dockerhub?

metadata.yaml
```diff
@@ -61,4 +61,4 @@ resources:
   prometheus-image:
     type: oci-image
     description: Container image for Prometheus
-    upstream-source: docker.io/ubuntu/prometheus:2.33-22.04_beta
+    upstream-source: ghcr.io/canonical/prometheus:latest
```
sed-i, line 64: The rock repo CI is separate from charm dev.
I think we want version tags here and there.

PR #448, opened 2023-02-02
check version bump on library modifications

lucabello:
## Issue
Forgetting to bump LIBPATCH on a library update causes that update to not be detected at any level, as the CI won't release a new library, but it can be detected by the PR label, leading to confusion.

## Solution
Add a simple check to the `static-lib` tox environment to make sure that *if* a library has been modified, its LIBPATCH has been bumped. Once this is approved and merged, I will replicate it to all of our other repos.   
I'm open to have this somewhere else if you think it's better, such as its own tox environment (which would require modfiying the *_quality_checks* CI to also run that step), or anywhere else.

## Context
Note that this is intentionally a loose check (it only checks if the LIBPATCH line was modified); this allows for *some* flexibility in not bumping the version by just adding a comment to that line.  
However, if there's a need for multiple PRs on the same library, and we don't want to release the library every time, the correct approach is probably to create another branch out of main (i.e., `library-change`); then merge the different PRs to `library-change`, and finally open a PR from that branch to main.

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -74,6 +74,10 @@ deps =
 commands =
     charm: mypy {[vars]src_path} {posargs}
     lib: mypy --python-version 3.5 {[vars]lib_path} {posargs}
+      /usr/bin/sh -c 'for m in $(git diff --name-only {[vars]lib_path}); do\
+      if ! git diff $m | grep -q "+LIBPATCH"; then\
+        echo "You forgot to bump LIBPATCH on $m!"; exit 1;\
+      fi; done'
     unit: mypy {[vars]tst_path}/unit {posargs}
     integration: mypy {[vars]tst_path}/integration {posargs}
 
--- a/tox.ini
+++ b/tox.ini
@@ -74,12 +74,13 @@ deps =
 commands =
     charm: mypy {[vars]src_path} {posargs}
     lib: mypy --python-version 3.5 {[vars]lib_path} {posargs}
-      /usr/bin/sh -c 'for m in $(git diff --name-only {[vars]lib_path}); do\
+      /usr/bin/env sh -c 'for m in $(git diff --name-only {[vars]lib_path}); do\
       if ! git diff $m | grep -q "+LIBPATCH"; then\
         echo "You forgot to bump LIBPATCH on $m!"; exit 1;\
       fi; done'
     unit: mypy {[vars]tst_path}/unit {posargs}
     integration: mypy {[vars]tst_path}/integration {posargs}
+allowlist_externals = /usr/bin/env
 
 [testenv:unit]
 description = Run unit tests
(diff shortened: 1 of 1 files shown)


tox.ini
```diff
@@ -74,8 +74,13 @@ deps =
 commands =
     charm: mypy {[vars]src_path} {posargs}
     lib: mypy --python-version 3.5 {[vars]lib_path} {posargs}
+      /usr/bin/env sh -c 'for m in $(git diff --name-only {[vars]lib_path}); do \
+      if ! git diff $m | grep -q "+LIBPATCH"; then \
+        echo "You forgot to bump LIBPATCH on $m!"; exit 1; \
+      fi; done'
```
sed-i, line 80: - Indentation seems off
- Would it make sense to move `allowlist_externals` directly above `commands`?
- We need a composite of "LIBAPI.LIBPATCH", hot just LIBPATCH
- Is there a way we could be [more sophisticated](https://stackoverflow.com/questions/1337320/how-to-grep-git-commit-diffs-or-contents-for-a-certain-word) with git usage to avoid the looping etc?

tox.ini
```diff
@@ -74,8 +74,13 @@ deps =
 commands =
     charm: mypy {[vars]src_path} {posargs}
     lib: mypy --python-version 3.5 {[vars]lib_path} {posargs}
```
sed-i, line 76:
```suggestion
    lib: mypy --python-version 3.8 {[vars]lib_path} {posargs}
```

tox.ini
lucabello: * Indenting a tab more in next commit! ✅ 
* `allowlist_externals` is at the very end in all our other environments, I'm just following the style of the file!
* Checking `+LIBPATCH\|+LIBAPI` now ✅ 
* That example would search for **one** `LIBPATCH` being updated across all the changes; if the charm owns two libraries and you only bump one, you wouldn't know through `git log`

dstathis (approved):

tox.ini
```diff
@@ -73,9 +73,14 @@ deps =
     integration: pytest-operator==1.0.0b1
 commands =
     charm: mypy {[vars]src_path} {posargs}
-    lib: mypy --python-version 3.5 {[vars]lib_path} {posargs}
+    lib: mypy --python-version 3.8 {[vars]lib_path} {posargs}
+      /usr/bin/env sh -c 'for m in $(git diff --name-only {[vars]lib_path}); do \
```
sed-i, line 77: 1. This command should have the `lib:` prefix so it only runs when running `tox -e static-lib` and not every testenv.
2. This check fails as soon as the modified lib file is staged (or committed). Which means CI won't ever fail. The easiest fix is to _add_ a diff to the main branch `git diff HEAD..main`.

lucabello: True, I'll fix this in the PRs I'm opening for the other repos and I'll open a fix for this one too!

PR #463, opened 2023-03-08
Remove problematic 'removal' tests

rbarry82:
## Issue
Get rid of the surfeit of tests which remove things. They aren't reliable. Neither is the regression test (which is doomed to fail right now)


tests/integration/test_charm.py
```diff
@@ -106,20 +106,3 @@ async def test_alert_rule_path_can_be_changed(ops_test, prometheus_tester_charm)
     rules_with_relation = await get_prometheus_rules(ops_test, prometheus_app_name, 0)
     tester_rules = get_rules_for(tester_app_name, rules_with_relation)
     assert len(tester_rules) == 2
-
-
-async def test_configuration_resets_after_relation_is_removed(ops_test):
```
sed-i, line 111: This probably needed an xfail, not deletion.

rbarry82: As in the daily, Pietro noticed a while ago that removing applications in CI is unreliable and problematic.

As much as I don't like removing test coverage either, if it's unreliable for him (and hangs here), there's not another option. `xfail` included. It did not fail. It just deadlocked the CI suite; not even output.

sed-i: In that case I'd place a `@skip`.
It's not only coverage, it's also effort.

rbarry82: Which is still in the git history once/if we get to a point that we can recover it.

I agree with this in principle, but it's a "if a tree falls..." argument about "if a test runs...". If it's `@skip`, it also doesn't run locally, and it doesn't run in CI, and may never work (which we can't meaningfully test, because it just hangs the runner and doesn't even have the good grace to fail), so it just becomes dead code.

tests/integration/test_rerelate.py
```diff
@@ -1,111 +0,0 @@
...
-async def test_remove_relation(ops_test: OpsTest):
-    await asyncio.gather(
-        ops_test.model.applications[app_name].remove_relation("metrics-endpoint", tester_app_name),
-        ops_test.model.applications[app_name].remove_relation("alertmanager", "alertmanager"),
-        ops_test.model.applications[app_name].remove_relation("grafana-source", "grafana"),
-        ops_test.model.applications[app_name].remove_relation(
-            "receive-remote-write", "grafana-agent"
-        ),
-    )
-    await ops_test.model.wait_for_idle(apps=[app_name], status="active", timeout=600)
```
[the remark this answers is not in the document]
rbarry82: The problem right now is that `.remove()` hangs indefinitely, so the test never proceeds anywhere, including to `wait_for_idle`. No updates will be posted about it waiting. Just a dead stop.

PR #465, opened 2023-03-10
Sync up the `alerts` method in the libraries

rbarry82:
## Issue
Make `prometheus_scrape` and `prometheus_remote_write` use the same logic for alerts -- one which prioritizes the labels on alerts (if present) over metadata, since providing charms who use `AlertRules` will have them injected, and we shouldn't clobber them.

`prometheus_remote_write` already does this, but `prometheus_scrape` (and `loki_push_api`) would otherwise overwrite forwarded rules with possibly misleading/meaningless data.

Make all of the UUIDs in the tests valid.

Fix a couple of tests with the trimmed JujuTopology-style identifier, as it's always used now.

### DIFF
--- a/lib/charms/prometheus_k8s/v0/prometheus_remote_write.py
+++ b/lib/charms/prometheus_k8s/v0/prometheus_remote_write.py
@@ -45,7 +45,7 @@ LIBAPI = 0
 
 # Increment this PATCH version before using `charmcraft publish-lib` or reset
 # to 0 if you are raising the major API version
-LIBPATCH = 13
+LIBPATCH = 14
 
 
 logger = logging.getLogger(__name__)
@@ -833,7 +833,7 @@ class PrometheusRemoteWriteProvider(Object):
 
         super().__init__(charm, relation_name)
         self._charm = charm
-        self.tool = CosTool(self._charm)
+        self._tool = CosTool(self._charm)
         self._relation_name = relation_name
         self._endpoint_schema = endpoint_schema
         self._endpoint_address = endpoint_address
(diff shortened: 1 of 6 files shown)


lib/charms/prometheus_k8s/v0/prometheus_remote_write.py
```diff
@@ -940,67 +941,85 @@ def alerts(self) -> dict:
...
+                    scrape_metadata = json.loads(relation.data[relation.app]["scrape_metadata"])
+                    identifier = JujuTopology.from_dict(scrape_metadata).identifier
+                    alerts[identifier] = self._tool.apply_label_matchers(alert_rules)  # type: ignore
+
+                except KeyError as e:
+                    logger.debug(
+                        "Relation %s has no 'scrape_metadata': %s",
+                        relation.id,
+                        e,
+                    )
```
sed-i, line 959: Is there a `continue` missing here?

sed-i (commented):
Is it an incremental change before replacing with `cosl`?

lib/charms/prometheus_k8s/v0/prometheus_remote_write.py
rbarry82: No, there isn't.

The `continue` is effectively the next line, where `if not identifier:` evalues to "if `self._get_identifier...` didn't return a `topology`, it may have returned a bare string as `identifier`. If there's no `topology`, and we can't build an `identifier` from `JujuTopology.from_dict(...)`", the bare string is the fallthrough.

At that point, `if not identifier` is checking the failure case of both the `try:` block and `_get_identifier_by_...`

PR #473, opened 2023-04-05
Scrape endpoints ourselves so we can label automatic series

rbarry82:
## Issue
Since automatic series cannot be relabeled with values from metrics. Read the endpoints to get them ourselves

## Release Notes
Scrape endpoints ourselves so we can label automatic series

### DIFF
--- a/lib/charms/prometheus_k8s/v0/prometheus_scrape.py
+++ b/lib/charms/prometheus_k8s/v0/prometheus_scrape.py
@@ -347,7 +347,6 @@ from urllib.error import HTTPError, URLError
 from urllib.parse import urlparse
 from urllib.request import urlopen
 
-
 import yaml
 from charms.observability_libs.v0.juju_topology import JujuTopology
 from ops.charm import CharmBase, RelationRole
--- a/lib/charms/prometheus_k8s/v0/prometheus_scrape.py
+++ b/lib/charms/prometheus_k8s/v0/prometheus_scrape.py
@@ -343,7 +343,10 @@ import tempfile
 from collections import defaultdict
 from pathlib import Path
 from typing import Any, Callable, Dict, List, Optional, Tuple, Union
+from urllib.error import HTTPError, URLError
 from urllib.parse import urlparse
+from urllib.request import urlopen
+
 
 import yaml
 from charms.observability_libs.v0.juju_topology import JujuTopology
@@ -368,7 +371,7 @@ LIBAPI = 0
 
 # Increment this PATCH version before using `charmcraft publish-lib` or reset
 # to 0 if you are raising the major API version
-LIBPATCH = 33
+LIBPATCH = 34
 
 logger = logging.getLogger(__name__)
 
@@ -2222,6 +2225,8 @@ class MetricsEndpointAggregator(Object):
                         "juju_application": application_name,
                         "juju_unit": unit_name,
                         "host": target["hostname"],
+                        # Expanding this will merge the dicts and replace the
+                        # topology labels if any were present/found
                         **self._static_config_extra_labels(target),
                     },
                 }
@@ -2244,7 +2249,16 @@ class MetricsEndpointAggregator(Object):
                 logger.debug("Could not perform DNS lookup for %s", target["hostname"])
                 dns_name = target["hostname"]
             extra_info["dns_name"] = dns_name
+        label_re = re.compile(r'(?P<label>[A-Za-z0-9_]+)\s?=\s?"(?P<value>.*?)",?')
 
+        try:
+            with urlopen(f'http://{target["hostname"]}:{target["port"]}/metrics') as resp:
+                data = resp.read().decode("utf-8").splitlines()
+                for metric in data:
+                    for match in label_re.finditer(metric):
+                        extra_info[match.group("label")] = match.group("value")
+        except (HTTPError, URLError, Exception) as e:
+            logger.debug("Could not scrape target: ", e)
         return extra_info
 
     @property


dstathis (commented):
I'm not sure I understand this change. Are we adding labels that are already there? If so, why?

rbarry82:
> I'm not sure I understand this change. Are we adding labels that are already there? If so, why?

Yes, sort-of. 

Specifically, this is because some reactive charms may have metrics labels we use/expect (such as `juju_unit`, though the formatting isn't the same and it's not our code). If a subordinate Telegraf is added as a target, those labels will not be part of automatic timeseries (such as `up` for dashboards variables), which would have the topology of the target (the subordinate) and not the principal

simskij:
I think we need to discuss this change in our next café. One of the earliest decisions we made was that if someone provides a label, we'll honor it. If we're moving away or making exceptions from that, it warrants more discussion than just a passing PR.

rbarry82:
> I think we need to discuss this change in our next café. One of the earliest decisions we made was that if someone provides a label, we'll honor it. If we're moving away or making exceptions from that, it warrants more discussion than just a passing PR.

No, we don't. 

We have already discussed this. This is the aggregator+telegraf, which we have already discussed, and looked for a workaround for. When related, we attach topology labels from the relation, which is the subordinate. 

That means that automatic timeseries (such as "up") have different topology than the actual metrics (if they have labels added on the scrape endpoint itself, which they do). And trying to filter by a variable in Grafana will result in empty dashboards, because `telegraf/0` is not `redis/0`, which the target itself adds. 

Currently, we *do not* honor provided labels. This is a fix for that, and there's some urgency to it. 

Enough, at least, that pushing it to the next cafe isn't reasonable considering that the root cause, explored options (relabeling, etc), side effects in Grafana, and so on have already been covered in stand-ups and sprint planning.

rbarry82:
For illustration, I'll give the background of this here also.

When telegraf is deployed and related to a subordinate, if you check the endpoint yourself, you will find labeled metrics:
```bash
❯ curl -s http://10.237.127.215:9103/metrics | grep -v '^#'                                                                                                                                                                                         
conntrack_ip_conntrack_count{host="bs:ubuntu-3",juju_application="ubuntu",juju_model="bs",juju_model_uuid="0d95c3d4-7e14-415f-8bba-d330fb816f9f",juju_unit="ubuntu-3"} 0                                                                            
conntrack_ip_conntrack_max{host="bs:ubuntu-3",juju_application="ubuntu",juju_model="bs",juju_model_uuid="0d95c3d4-7e14-415f-8bba-d330fb816f9f",juju_unit="ubuntu-3"} 1.048576e+06                                                                   
exec{host="bs:ubuntu-3",juju_application="ubuntu",juju_model="bs",juju_model_uuid="0d95c3d4-7e14-415f-8bba-d330fb816f9f",juju_unit="ubuntu-3"} 8480  
```

Their formatting of Juju topology doesn't quite match ours (`juju_unit` as `{application}-{num}` rather than `{application}/{num}`, for example). But when telegraf is related to cos-proxy, _we_ set labels based on the topology. Of the relation.

relabeling operates in a pseudo foreach loop (that is, you cannot search all possible label values in all metrics from an endpoint to find an interesting one), and a replacement for series without such a metric (such as `up`) won't match anything. You can _add_ fixed-text labels, but that doesn't help us here. The topology we set is that of the relation, and the relation is from the subordinate.

That means that in Prometheus itself, we see:

```
up{dns_name="juju-816f9f-7.lxd", host="10.237.127.215", instance="bs_0d95c3d4-7e14-415f-8bba-d330fb816f9f_telegraf_telegraf/2", job="juju_bs_0d95c3d_telegraf_prometheus_scrape", juju_application="telegraf", juju_model="bs", juju_model_uuid="0d95c3d4-7e14-415f-8bba-d330fb816f9f", juju_unit="telegraf/2"}
```

For one thing, this _probably_ qualifies as not honoring the labels they intended to provide. Sure, they're there. But it also means that common selectors for Grafana variables, such as:

```
"juju_application":  'label_values(up{juju_model=~"$juju_model",juju_model_uuid=~"$juju_model_uuid"},juju_application)'
```

Won't find it. There is _not_ an `up` metric for `ubuntu`, so it cannot be picked, and it is not possible to drill into the metrics themselves, which do, in fact, keep the labels in the target themselves:

```
net_bytes_recv{dns_name="juju-816f9f-7.lxd", host="bs:ubuntu-3", instance="bs_0d95c3d4-7e14-415f-8bba-d330fb816f9f_telegraf_telegraf/2", interface="eth0", job="juju_bs_0d95c3d_telegraf_prometheus_scrape", juju_application="ubuntu", juju_model="bs", juju_model_uuid="0d95c3d4-7e14-415f-8bba-d330fb816f9f", juju_unit="ubuntu-3"}
```

This patch has a single effect: _other_ metrics (such as `up`) have values which match the actual metrics:

```
up{dns_name="juju-816f9f-7.lxd", host="bs:ubuntu-3", instance="bs_0d95c3d4-7e14-415f-8bba-d330fb816f9f_ubuntu_ubuntu-3",job="juju_bs_0d95c3d_telegraf_prometheus_scrape", juju_application="ubuntu", juju_model="bs", juju_model_uuid="0d95c3d4-7e14-415f-8bba-d330fb816f9f", juju_unit="ubuntu-3"}
--

This causes the automatic timeseries to also abide by the "spirit" of labels authors have set themselves (by bubbling them up and overriding the default topology labels we infer from the relation; mostly important for subordinate/principal relations), and they are now able to be drilled into via Grafana/etc, where `up{juju_application="some_principal_charm"}` will be accurate, rather than `up{juju_unit="telegraf/9999"}` and mapping that back to whatever its principal is by users.

simskij (approved):

PR #476, opened 2023-04-14
Catch more exception types, only capture 'juju_' labels to make execution faster

rbarry82:
## Issue
`requests` may throw a `ConnectionResetError` when attempting to read the response inside a contextmanager, and we should also catch that. 

Ensure that all args are converted when logging. And trim the regexp down to only capture topology labels (for Grafana dashboards)

## Release Notes
Catch more exception types, only capture 'juju_' labels to make execution faster

### DIFF
--- a/lib/charms/prometheus_k8s/v0/prometheus_scrape.py
+++ b/lib/charms/prometheus_k8s/v0/prometheus_scrape.py
@@ -370,7 +370,7 @@ LIBAPI = 0
 
 # Increment this PATCH version before using `charmcraft publish-lib` or reset
 # to 0 if you are raising the major API version
-LIBPATCH = 34
+LIBPATCH = 35
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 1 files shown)


lib/charms/prometheus_k8s/v0/prometheus_scrape.py
```diff
@@ -2248,16 +2248,16 @@ def _static_config_extra_labels(self, target: Dict[str, str]) -> Dict[str, str]:
...
         try:
             with urlopen(f'http://{target["hostname"]}:{target["port"]}/metrics') as resp:
                 data = resp.read().decode("utf-8").splitlines()
                 for metric in data:
                     for match in label_re.finditer(metric):
                         extra_info[match.group("label")] = match.group("value")
-        except (HTTPError, URLError, Exception) as e:
-            logger.debug("Could not scrape target: ", e)
+        except (HTTPError, URLError, OSError, ConnectionResetError, Exception) as e:
+            logger.debug("Could not scrape target: %s", e)
```
Abuelodelanada, line 2260: Shouldn't this log be `error` instead of `debug`?

Abuelodelanada (approved):

simskij (approved):

PR #478, opened 2023-04-27
use ruff instead of flake8

lucabello:
The **flake8** plugins we use are re-implemented in **ruff**, which is why they have been removed from the dependencies.

**isort** is enabled (and used when formatting) through the `I001` rule.

The "exclude list" has been updated to omit folders that **ruff** excludes by default.

Everything else is the same, the only difference being two new rules being ignored (which weren't checked before anyway): 
* `C901: Function is too complex`, which I excluded as we were not checking it before either (an eventual refactor should have its own PR);
* `RET504: Unnecessary variable assignment before return statement`, which we do everywhere as it increases readability.


dstathis (approved):

lucabello:
@simskij you might need to review the branch protection rules on this repo, as it looks like it's stuck even though all checks passed :)

PR #488, opened 2023-06-06
feat: add TLS communication

lucabello:
## Issue
This PR tackles the lack of secure communication between cos-lite components.

## Solution
The communications are encrypted by making use of TLS; this PR adds support for certificates and inter-charm exchanges by making use of the `tls_certificates` library.

### DIFF
--- a/charmcraft.yaml
+++ b/charmcraft.yaml
@@ -12,6 +12,10 @@ parts:
   charm:
     build-packages:
       - git
+      - libffi-dev
+      - libssl-dev
+      - rustc
+      - cargo
     charm-python-packages:
       - wheel==0.37.1
       - setuptools==45.2.0
--- a/src/charm.py
+++ b/src/charm.py
@@ -9,7 +9,7 @@ import logging
 import re
 import socket
 from pathlib import Path
-from typing import Dict, Optional, cast
+from typing import Dict, Optional, Union, cast
 from urllib.parse import urlparse
 
 import yaml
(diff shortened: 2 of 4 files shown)


src/charm.py
```diff
@@ -164,18 +173,61 @@ def __init__(self, *args):
...
+            self._on_certificate_invalidated,
+        )
+        self.framework.observe(
+            self.certificates.on.all_certificates_invalidated,  # pyright: ignore
+            self._on_all_certificates_invalidated,
+        )
+
+        self.framework.observe(
+            self.on.prometheus_pebble_ready, self._on_pebble_ready  # pyright: ignore
+        )
```
simskij, line 203: This probably should not be here?

src/charm.py
```diff
@@ -792,6 +844,119 @@ def _push(self, path, contents):
...
+            return
+        if event.reason == "revoked":
+            self._certificate_revoked(event)
+        if event.reason == "expired":
+            self._on_certificate_expiring(event)
+
+    def _on_all_certificates_invalidated(self, event: AllCertificatesInvalidatedEvent) -> None:
+        # Do what you want with this information, probably remove all certificates.
+        pass
+
```
simskij, line 959: Suggest extracting this into it's own component, or even better - a lib.

simskij (changes requested):

charmcraft.yaml
```diff
@@ -12,6 +12,10 @@ parts:
   charm:
     build-packages:
       - git
+      - libffi-dev
+      - libssl-dev
+      - rustc
+      - cargo
```
simskij, line 18: What is the impact on build times by adding these as build packages?

PR #490, opened 2023-07-04
do not add default metrics if lookaside jobs not empty

neoaggelos:
## Issue

`MetricsEndpointProvider` supports a callable function to generate the jobs dynamically. If no static jobs are set, the default scrape job is always added to the scrape jobs.

```
self._jobs = self._jobs if self._jobs else [DEFAULT_JOB]
```

## Solution

Only return the default job if no static jobs and no lookaside jobs are defined

## Context

https://github.com/canonical/charm-microk8s/pull/88

I have assumed that the `[DEFAULT_JOB]` is just a default fallback, so I do not expect that it should be added in this case. Setting the jobs to an empty list is still adding the default job.

## Release Notes

Do not add a default scrape job if there are lookaside jobs

### DIFF
--- a/lib/charms/prometheus_k8s/v0/prometheus_scrape.py
+++ b/lib/charms/prometheus_k8s/v0/prometheus_scrape.py
@@ -1793,10 +1793,10 @@ class MetricsEndpointProvider(Object):
            A list of dictionaries, where each dictionary specifies a
            single scrape job for Prometheus.
         """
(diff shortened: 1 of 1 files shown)


simskij (approved):

sed-i (approved):

sed-i:
Itests [passed](https://github.com/canonical/prometheus-k8s-operator/actions/runs/5458264716), but gha failed to report back. Merging.

PR #492, opened 2023-07-06
add charm tracing integration

lucabello:
This PR adds the charm tracing integration to Prometheus.


lib/charms/tempo_k8s/v0/tempo_scrape.py
PietroPasotti, line 1: I renamed this library early on, it should be 'tracing', not 'tempo_scrape' (it's not a scrape but a push model, I found out pretty early) :)
so delete this one, fetch-lib with 'tracing' instead, and rename the imports. The rest should work out of the box.

tox.ini
```diff
@@ -78,6 +78,8 @@ deps =
     coverage[toml]
     responses==0.20.0
     -r{toxinidir}/requirements.txt
+    # OpenTelemetry is in PYDEPS for tempo instrumentation
+    opentelemetry-exporter-otlp-proto-grpc
```
PietroPasotti, line 82: in traefik, I added an optional_dependencies field to pyproject.toml called `pydeps`.
then in tox, in all envs that need the charm libs, you can write `.[pydeps]` and it will load them from pyproject.toml. 
I think it's nicer to have a single place where we need to keep them up to date.

But it's just a nice to have.

PietroPasotti (approved):

lib/charms/tempo_k8s/v0/tempo_scrape.py
lucabello: Done!

pyproject.toml
```diff
@@ -57,5 +57,10 @@ follow_imports = "silent"
 module = ["ops.*"]
 follow_imports = "skip"
 
+[tool.pyright]
```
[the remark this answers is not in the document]
lucabello: This repo still doesn't use pyright for linting though, but I'm onto that :)

Issue #493, opened 2023-07-06
jsonschema update has indirect build dependency on rust compiler


lengau:
### Bug Description

The prometheus k8s operator fails to build from a fresh clone

### To Reproduce

1. `git clone git@github.com:canonical/prometheus-k8s-operator.git`
2. `cd prometheus-k8s-operator/`
3. `charmcraft pack --verbose`

### Environment

Using Charmcraft 2.3.0.post12+gitd244e61 (latest/edge),

### Relevant log output

```shell
::    :: Collecting rpds-py>=0.7.1                                                                                                                             
::    ::   Downloading rpds_py-0.8.8.tar.gz (16 kB)                                                                                                            
::    ::   Installing build dependencies: started                                                                                                              
::    ::   Installing build dependencies: finished with status 'error'                                                                                         
::    ::   ERROR: Command errored out with exit status 1:                                                                                                      
::    ::    command: /root/parts/charm/build/staging-venv/bin/python3 /root/parts/charm/build/staging-venv/lib/python3.8/site-packages/pip install --ignore-installed --no-user --prefix /tmp/pip-build-env-_k8xehix/overlay --no-warn-script-location --no-binary :all: --only-binary :none: -i https://pypi.org/simple -- 'maturin>=1.0,<2.0'                                                                                                                                                
::    ::        cwd: None                                                                                                                                      
::    ::   Complete output (65 lines):                                                                                                                         
::    ::   Collecting maturin<2.0,>=1.0                                                                                                                        
::    ::     Downloading maturin-1.1.0.tar.gz (181 kB)                                                                                                         
::    ::     Installing build dependencies: started                                                                                                            
::    ::     Installing build dependencies: finished with status 'done'                                                                                        
::    ::     Getting requirements to build wheel: started                                                                                                      
::    ::     Getting requirements to build wheel: finished with status 'done'                                                                                  
::    ::       Preparing wheel metadata: started                                                                                                               
::    ::       Preparing wheel metadata: finished with status 'done'                                                                                           
::    ::   Collecting tomli>=1.1.0; python_version < "3.11"                                                                                                    
::    ::     Using cached tomli-2.0.1.tar.gz (15 kB)                                                                                                           
::    ::     Installing build dependencies: started                                                                                                            
::    ::     Installing build dependencies: finished with status 'done'                                                                                        
::    ::     Getting requirements to build wheel: started                                                                                                      
[...]

PR #494, opened 2023-07-07
deps: add rust build dependencies to charm

lengau:
## Issue

#493

## Solution

Adds `rustc` and `cargo` as build packages in the 

## Context

The jsonschema update broke [chramcraft's spread tests](https://github.com/canonical/charmcraft/actions/runs/5479458604/jobs/9982991680)

## Testing Instructions

Build the charm :-)

## Release Notes

Add build packages so newer jsonschema versions work.


charmcraft.yaml
```diff
@@ -12,6 +12,8 @@ parts:
   charm:
     build-packages:
       - git
+      - rustc
+      - cargo
```
sed-i, line 16: At this point I think I'd rather get the binary. Otherwise it's 20-30 min packing.
Wdyt?

```suggestion
    charm-binary-python-packages:
      - jsonschema
```

sed-i (commented):
Thanks @lengau.
Change is good but I wonder if we could get the binary instead.

charmcraft.yaml
simskij: Until @sergiusens patch that allows caching pip build artifacts land in charmcraft (maybe it already has?), that seems like a reasonable trade-off to not make it too tedious to iterate on the charm. We could set up our own fork to build the binary with more control.

lengau:
That's a good question @sed-i . I sent this because of the general preference for building everything from source for production charms, but getting the binary instead is indeed much easier 😀

charmcraft.yaml
lengau: Charmcraft does already cache built wheels across runs (with `charmcraft clean` eliminating that cache), but it will of course slow down CI and first build after a clean.

sed-i:
[CI passed](https://github.com/canonical/prometheus-k8s-operator/actions/runs/5539188048/jobs/10109892913), but it is not reflected on the PR page here. Merging.

PR #509, opened 2023-08-16
fix: only add TLS config if cert file has been written to disk

### LINKED ISSUE #506: TLS config is written and workload restarted, before the cert is written to disk
### Bug Description

Replan fails because `/etc/prometheus/server.cert` does not yet exist.

### To Reproduce

Quickly relate self-signed-certificates --- prometheus --- traefik.

### Environment

```
Model        Controller  Cloud/Region        Version  SLA          Timestamp
welcome-k8s  microk8s    microk8s/localhost  3.2.0    unsupported  14:50:08-04:00

App                   Version  Status  Scale  Charm                     Channel  Rev  Address         Exposed  Message
ca                             active      1  self-signed-certificates  edge      17  10.152.183.104  no       
prom                  2.46.0   active      2  prometheus-k8s            edge     136  10.152.183.78   no       
trfk                  2.9.6    active      1  traefik-k8s               edge     141  10.43.8.149     no       
```

### Relevant log output

```shell
unit-prom-0: 14:43:25.205 ERROR unit.prom/0.juju-log Failed to replan; pebble layer: {'summary': 'Prometheus layer', 'description': 'Pebble layer configuration for Prometheus', 'services': {'prometheus': {'summary': 'prometheus daemon', 'startup': 'enabled', 'override': 'replace', 'command': '/bin/prometheus --config.file=/etc/prometheus/prometheus.yml --storage.tsdb.path=/var/lib/prometheus --web.enable-lifecycle --web.console.templates=/usr/share/prometheus/consoles --web.console.libraries=/usr/share/prometheus/console_libraries --web.config.file=/etc/prometheus/prometheus-web-config.yml --web.external-url=https://10.43.8.149/welcome-k8s-prom-0 --log.level=info --storage.tsdb.retention.time=15d --storage.tsdb.retention.size=0.8GB'}}}; cannot perform the following tasks:
- Start service "prometheus" (cannot start service: exited quickly with code 1)
----- Logs from task 0 -----
2023-08-08T18:43:25Z INFO Most recent service output:
    ts=2023-08-08T18:43:25.084Z caller=main.go:585 level=info msg="Starting Prometheus Server" mode=server version="(version=2.46.0, branch=HEAD, revision=cbb69e51423565ec40f46e74f4ff2dbb3b7fb4f0)"
    ts=2023-08-08T18:43:25.084Z caller=main.go:590 level=info build_context="(go=go1.19.11, platform=linux/amd64, user=root@rockcraft-prometheus-322591, date=20230730-12:08:01, tags=netgo,builtinassets,stringlabels)"
    ts=2023-08-08T18:43:25.084Z caller=main.go:591 level=info host_details="(Linux 5.15.0-78-generic #85-Ubuntu SMP Fri Jul 7 15:25:09 UTC 2023 x86_64 prom-0 (none))"
    ts=2023-08-08T18:43:25.084Z caller=main.go:592 level=info fd_limits="(soft=65536, hard=65536)"
    ts=2023-08-08T18:43:25.084Z caller=main.go:593 level=info vm_limits="(soft=unlimited, hard=unlimited)"
    ts=2023-08-08T18:43:25.085Z caller=web.go:563 level=info component=web msg="Start listening for connections" address=0.0.0.0:9090
    ts=2023-08-08T18:43:25.085Z caller=main.go:846 level=error msg="Unable to validate web configuration file" err="failed to load X509KeyPair: open /etc/prometheus/server.cert: no such file or directory"
2023-08-08T18:43:25Z ERROR cannot start service: exited quickly with code 1
-----
```

### Additional context

_No response_

lucabello:
## Issue
Fixes #506.

This likely happens when `_configure()` is called by something other than `_on_server_cert_changed` (e.g., a perfectly timed `update-status`), but when the certificate is already in relation data (or Prometheus has already been related through the `certificates` interface but the certificate is not ready yet).

## Solution
Only add the `tls_server_config` for Prometheus if the certificate exists on the file sytstem.

The `_configure()` method is also called by `_on_server_cert_changed()`, which (sequentially) writes the cert to disk before calling `_configure()`. This means we don't risk skip adding the TLS configuration with this extra check.

## Testing Instructions
It's pretty much a race condition, so it's hard to add a test for it. I guess trying to replicate the issue is something, though of course it doesn't guarantee it.

## Release Notes
Only add the `tls_server_config` for Prometheus if the certificate exists on the file sytstem.

### DIFF
--- a/src/charm.py
+++ b/src/charm.py
@@ -339,7 +339,7 @@ class PrometheusCharm(CharmBase):
         return self.internal_url
 
     def _is_tls_enabled(self):
-        return bool(self.cert_handler.cert)
+        return self.cert_handler.enabled
 
     @property
     def _prometheus_layer(self) -> Layer:
@@ -363,6 +363,10 @@ class PrometheusCharm(CharmBase):
         }
 
         return Layer(layer_config)  # pyright: ignore
+
+    def stop(self) -> None:
+        """Stop Prometheus."""
+        self.container.stop("prometheus")
 
     def _resource_reqs_from_config(self):
         limits = {
@@ -789,13 +793,16 @@ class PrometheusCharm(CharmBase):
 
         Ref: https://prometheus.io/docs/prometheus/latest/configuration/https/
         """
-        if self._is_tls_enabled() and self.container.exists(CERT_PATH):
-            return {
-                "tls_server_config": {
-                    "cert_file": CERT_PATH,
-                    "key_file": KEY_PATH,
+        if self._is_tls_enabled():
+            if self.container.exists(CERT_PATH):
+                return {
+                    "tls_server_config": {
+                        "cert_file": CERT_PATH,
+                        "key_file": KEY_PATH,
+                    }
                 }
-            }
+            else:
+                self.stop()
         return None
 
     def _alerting_config(self) -> dict:
--- a/src/charm.py
+++ b/src/charm.py
@@ -411,9 +411,11 @@ class PrometheusCharm(CharmBase):
 
         self.grafana_source_provider.update_source(self.external_url)
         self._configure(_)
-        if isinstance(self.unit.status, WaitingStatus) and self.unit.status.message == "Waiting for TLS certificates to be written to file":
+        if (
+            isinstance(self.unit.status, WaitingStatus)
+            and self.unit.status.message == "Waiting for TLS certificates to be written to file"
+        ):
             self.unit.status = ActiveStatus()
-
 
     def _configure(self, _):
         """Reconfigure and either reload or restart Prometheus.
@@ -904,7 +906,9 @@ class PrometheusCharm(CharmBase):
             # expected to happen as soon as the the related CA replies with a cert.
             self.stop()
             if isinstance(self.unit.status, ActiveStatus):
-                self.unit.status = WaitingStatus("Waiting for TLS certificates to be written to file")
+                self.unit.status = WaitingStatus(
+                    "Waiting for TLS certificates to be written to file"
+                )
         else:
             if web_config := self._web_config():
                 self._push(WEB_CONFIG_PATH, yaml.safe_dump(web_config))
--- a/src/charm.py
+++ b/src/charm.py
@@ -789,7 +789,7 @@ class PrometheusCharm(CharmBase):
 
         Ref: https://prometheus.io/docs/prometheus/latest/configuration/https/
         """
-        if self._is_tls_enabled():
+        if self._is_tls_enabled() and self.container.exists(CERT_PATH):
             return {
                 "tls_server_config": {
                     "cert_file": CERT_PATH,
--- a/src/charm.py
+++ b/src/charm.py
@@ -411,6 +411,9 @@ class PrometheusCharm(CharmBase):
 
         self.grafana_source_provider.update_source(self.external_url)
         self._configure(_)
+        if isinstance(self.unit.status, WaitingStatus) and self.unit.status.message == "Waiting for TLS certificates to be written to file":
+            self.unit.status = ActiveStatus()
+
 
     def _configure(self, _):
         """Reconfigure and either reload or restart Prometheus.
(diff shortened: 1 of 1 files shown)


dstathis (dismissed):
This change does not fix the issue in a safe way. If tls is enabled, we should not be allowing http traffic at all, but with this solution, we allow http traffic if we can not find the cert.

lucabello:
I don't think it's necessarily in the scope of this issue, which is the failed replan; this behavior when TLS is enabled has been implemented in the TLS PR :)

I'll add it anyway!

lucabello:
After noticing the itests were passing locally, rerunning them fixed things :D

src/charm.py
```diff
@@ -785,12 +789,15 @@ def _web_config(self) -> Optional[dict]:
         Ref: https://prometheus.io/docs/prometheus/latest/configuration/https/
         """
         if self._is_tls_enabled():
-            return {
-                "tls_server_config": {
-                    "cert_file": CERT_PATH,
-                    "key_file": KEY_PATH,
+            if not self.container.exists(CERT_PATH):
+                self.stop()
```
sed-i, line 793: Seems odd to issue a stop from a config getter func.
Is there anywhere else this could fit better?

src/charm.py
```diff
@@ -333,7 +333,7 @@ def sans(self) -> List[str]:
         return sans
 
     def _is_tls_enabled(self):
-        return bool(self.cert_handler.cert)
+        return self.cert_handler.enabled
```
sed-i, line 336: So `.enabled` only means the relation is in place. Is that the semantics we're aiming for now with the `stop()` approach?

lucabello: Yes!

src/charm.py
lucabello: I don't think so; this is also the approach that's already used in grafana-agent :)

sed-i: At the very least we should add a "side effect" section to the docstring.

But this change seems to deepen the spaghetti: we have a `stop` here but not a `start`.
If this is the best we can do then maybe add a comment here?
```
# After a `stop`, the service will autostart on next call to `_configure`, which is expected
# to happen as soon as the the related CA replies with a cert.
```

Iiuc, there's a complication here with relation-broken: on relation-broken I imagine `cert_handler.enabled` would still return True, but what happens with CERT_PATH? Is it deleted before or after? And maybe for this reason we should move the `stop` elsewhere?

sed-i (commented):
Looks like you pinpointed the issue. Need to figure out relation-broken.
Also add a utests? Something like:
```
# GIVEN an init'ed charm
# THEN the prometheus service is running
# WHEN tls-certificates relation joins
# THEN the prometheus service stops
# AND WHEN the remote charm sends the cert 
# THEN the prometheus service is running
```

src/charm.py
lucabello: The reason why `stop` is there is because that's the point in the code where we decide if we have TLS or not (and thus create the web config and pass it via the Pebble plan).  
The fact that the service is reloaded-restarted in `_configure`, is already documented in configure; sure I'll add that extra comment, but it's not new behavior (nor particularly quirky; lots of our charms do this). We don't need `start` because Pebble `replan()` already does this.  

`cert_handler.enabled` is currently implemented as:
```python
    @property
    def enabled(self) -> bool:
        """Boolean indicating whether the charm has a tls_certificates relation."""
        # We need to check for units as a temporary workaround because of https://bugs.launchpad.net/juju/+bug/2024583
        # This could in theory not work correctly on scale down to 0 but it is necessary for the moment.
        return (
            len(self.charm.model.relations[self.certificates_relation_name]) > 0
            and len(self.charm.model.get_relation(self.certificates_relation_name).units) > 0  # type: ignore
        )
```

About `CERT_PATH` being deleted before or after relation-broken, it doesn't matter; `cert_handler.enabled` will return false; the charm will still exclude the TLS configuration because the relation is not there. The relation-broken issues would be better addressed in [cert_handler:_on_server_cert_changed()](https://github.com/canonical/prometheus-k8s-operator/blob/main/src/charm.py#L386), where we could check if we are in a relation-broken event and eventually just remove the certs.    

**Even if** `cert_handler.enabled` returned true, the issue concerns `cert_handler` and not the individual charm, and as such it's not part of the scope of this issue.

I'm fully open to reconsider if methods like `_configure` should have side effects like stopping the charm, however I want to reiterate [we already do this in grafana-agent](https://github.com/canonical/grafana-agent-k8s-operator/blob/main/src/grafana_agent.py#L479); I'd leave further design discussions out of this issue.

dstathis: We decided that we should move these checks to the outer `_configure ` method.

dstathis: Also, we should change to `WaitingStatus` while Prometheus is stopped.

Abuelodelanada (approved):

PR #510, opened 2023-08-21
Cosl alerts

dstathis:
## Issue
Partial progress on https://github.com/canonical/grafana-agent-k8s-operator/issues/190

## Solution
Use the JujuTopology and AlertRules classes from cosl.

## Testing Instructions
There should be no changes to the behavior of the libraries so I am not sure what testing should be done. Unit tests are passing.

## Release Notes
Libraries now use upstream JujuTopology and AlertRules classes from cosl


lucabello (approved):
LGTM, provided itests are fixed; if the tracing relation is what is causing the issues, I'd remove it for the time being (we don't really use it at the moment anyway), as that's only a couple lines of code anyway and it will be extremely easy to re-add.

PR #513, opened 2023-08-24
Alert dict

dstathis:
## Issue
Partial progress was made on converting  prometheus to use the upstream cosl JujuTopology but I missed a few things.

## Solution
Fully use JujuTopology from cosl

## Testing Instructions
Behavior should not have changed so just general testing.

## Release Notes
Now fully uses JujuTopology from cosl


simskij (approved):

simskij:
@dstathis I'm fine merging once the integration test is fixed

Abuelodelanada (approved):

PR #515, opened 2023-08-29
Insecure skip verify

Abuelodelanada:
## Issue

This PR removes `insecure_skip_verify` from Prometheus config


src/charm.py
```diff
@@ -919,6 +880,47 @@ def _generate_prometheus_config(self) -> bool:
...
+        if (tls_config := job.get("tls_config", {})) or job.get("scheme") == "https":
+            # Certs are transferred over relation data and need to be written to files on disk.
+            # CA certificate to validate the server certificate with.
+
+            # If the scrape job has a TLS section but no "ca_file", then use ours, assuming
+            # prometheus and all scrape jobs are signed by the same CA.
+            if ca_file := tls_config.get("ca_file") or self.cert_handler.ca:
+                # TODO we shouldn't be passing CA certs over relation data, because that
+                #  reduces to self-signed certs. Both parties need to separately trust the CA
+                #  instead.
```
dstathis, line 894: I agree with this TODO comment. Why are we configuring CA here at all. Shouldn't we just be using our system certs?

dstathis (approved):

src/charm.py
Abuelodelanada: This is not new, I just extract the code to a new method and keep the comments.
We have an issue to use secrets.

lib/charms/prometheus_k8s/v0/prometheus_scrape.py
```diff
@@ -607,9 +607,6 @@ def render_alertmanager_static_configs(alertmanagers: List[str]):
                     "scheme": scheme,
```
sed-i, line 607: Should we document this?
```suggestion
                    # For https we still do not render a `tls_config` section because
                    # certs are expected to be made available by the charm via the
                    # `update-ca-certificates` mechanism.
                    "scheme": scheme,
```

lib/charms/prometheus_k8s/v0/prometheus_scrape.py
```diff
@@ -1176,16 +1173,6 @@ def _static_scrape_config(self, relation) -> list:
             scrape_configs, hosts, topology
         )
 
-        # If scheme is https but no ca section present, then auto add "insecure_skip_verify",
-        # otherwise scraping errors out with "x509: certificate signed by unknown authority".
-        # https://prometheus.io/docs/prometheus/latest/configuration/configuration/#tls_config
```
sed-i, line 1181: Should we document this?

> For https scrape targets we still do not render a `tls_config` section because
> certs are expected to be made available by the charm via the `update-ca-certificates` mechanism.

src/charm.py
```diff
@@ -919,6 +880,47 @@ def _generate_prometheus_config(self) -> bool:
         logger.info("Pushed new configuration")
         return True
 
+    def _process_tls_config(self, job) -> Dict:
+        certs = {}
```
sed-i, line 884:
```suggestion
        certs: Dict[str, str] = {}  # Mapping form cert filename to cert content.
```

src/charm.py
```diff
@@ -919,6 +880,47 @@ def _generate_prometheus_config(self) -> bool:
...
+                filename = f"{PROMETHEUS_DIR}/{job['job_name']}-client.key"
+                certs[filename] = key_file
+                job["tls_config"]["key_file"] = filename
+            elif "cert_file" in tls_config or "key_file" in tls_config:
+                raise ConfigError(
+                    'tls_config requires both "cert_file" and "key_file" if client '
+                    "authentication is to be used"
+                )
+
+        return {"job": job, "certs": certs}
```
sed-i, line 922: Should we simply `return job, certs` here, so that we could `processed_job, extracted_certs = self._process_tls_config(job)` above?

sed-i (approved):

src/charm.py
Abuelodelanada: Yes, it's a little bit golanger, but I can live with that 😄

PR #519, opened 2023-09-06
feat: remote_write_receiver option

Guillaumebeuzeboc:
## Issue
the juju cos setup cannot receive data from the grafana-agent if the grafana-agent is not running inside juju.

## Solution
Add a charm option to enable the remote `remote_write_receiver` feature even when no relation is established.

## Context
IoT devices cannot run juju due to resources constraint, so they must run grafana-agent snap standalone.
This way the COS stack can be used on servers and snaps can be used on devices to export data.

## Notes

This is a draft PR as a suggestion as I am exploring the usage of COS for robots.
The PR is based on my recent learning of Juju/Charm and COS.

### DIFF
--- a/config.yaml
+++ b/config.yaml
@@ -72,8 +72,3 @@ options:
       automatically deduced from it).
       See https://kubernetes.io/docs/concepts/configuration/manage-resources-containers/
     type: string
-  remote_write_receiver:
-    description: |
-      enforce the remote-write-receiver feature
-    type: boolean
-    default: false
--- a/config.yaml
+++ b/config.yaml
@@ -72,3 +72,8 @@ options:
       automatically deduced from it).
       See https://kubernetes.io/docs/concepts/configuration/manage-resources-containers/
     type: string
+  remote_write_receiver:
+    description: |
+      enforce the remote-write-receiver feature
+    type: boolean
+    default: false
--- a/src/charm.py
+++ b/src/charm.py
@@ -667,10 +667,7 @@ class PrometheusCharm(CharmBase):
         args.append(f"--web.external-url={external_url}")
         args.append("--web.route-prefix=/")
 
-        if self.model.relations[DEFAULT_REMOTE_WRITE_RELATION_NAME] or config.get(
-            "remote_write_receiver"
-        ):
-            args.append("--web.enable-remote-write-receiver")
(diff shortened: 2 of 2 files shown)


dstathis:
This seems like a good idea. What do you think @simskij?

simskij (approved):

config.yaml
```diff
@@ -72,3 +72,8 @@ options:
       automatically deduced from it).
       See https://kubernetes.io/docs/concepts/configuration/manage-resources-containers/
     type: string
+  remote_write_receiver:
```
sed-i, line 75: I think we shouldn't have this config option. Instead, always enable remote write:
Currently, when we enable remote-write, it's enabled for everyone; afaict, there isn't greater risk in enabling it to begin with.

sed-i (changes requested):
Thanks @Guillaumebeuzeboc!
Ideally we'd want to keep config options minimal or scalable.
Since juju config options cannot be nested, we try to stick to minimal.

src/charm.py
```diff
@@ -667,7 +667,9 @@ def _generate_command(self) -> str:
         args.append(f"--web.external-url={external_url}")
         args.append("--web.route-prefix=/")
 
-        if self.model.relations[DEFAULT_REMOTE_WRITE_RELATION_NAME]:
+        if self.model.relations[DEFAULT_REMOTE_WRITE_RELATION_NAME] or config.get(
+            "remote_write_receiver"
+        ):
             args.append("--web.enable-remote-write-receiver")
```
simskij, line 673:
```suggestion
        args.append("--web.enable-remote-write-receiver")
```

config.yaml
```diff
@@ -72,3 +72,8 @@ options:
...
+    description: |
+      enforce the remote-write-receiver feature
+    type: boolean
+    default: false
```
simskij, line 79:
```suggestion
```

simskij (approved):

simskij:
@Guillaumebeuzeboc would you mind squashing your commits into a signed commit to allow for merging? Thanks.

Guillaumebeuzeboc:
> @Guillaumebeuzeboc would you mind squashing your commits into a signed commit to allow for merging? Thanks.

Done, let me know if it's all good.
Thanks!

Issue #521, opened 2023-09-07
prometheus fails to restart after loki upgrade


PietroPasotti:
### Bug Description

prometheus locked in error state after refreshing loki

### To Reproduce

juju deploy cos-lite --trust
juju refresh loki --channel edge

### Environment

microk8s
edge prometheus (rev 143)

### Relevant log output

```shell
unit-prometheus-0: 17:27:29 INFO unit.prometheus/0.juju-log HTTP Request: GET https://10.152.183.1/api/v1/namespaces/bar/pods/prometheus-0 "HTTP/1.1 200 OK"
unit-prometheus-0: 17:27:30 INFO unit.prometheus/0.juju-log reqs=ResourceRequirements(claims=None, limits={}, requests={'cpu': '0.25', 'memory': '200Mi'}), templated=ResourceRequirements(claims=None, limits=None, requests={'cpu': '250m', 'memory': '200Mi'}), actual=ResourceRequirements(claims=None, limits=None, requests={'cpu': '250m', 'memory': '200Mi'})
unit-prometheus-0: 17:27:30 INFO unit.prometheus/0.juju-log HTTP Request: GET https://10.152.183.1/apis/apps/v1/namespaces/bar/statefulsets/prometheus "HTTP/1.1 200 OK"
unit-prometheus-0: 17:27:30 INFO unit.prometheus/0.juju-log HTTP Request: GET https://10.152.183.1/api/v1/namespaces/bar/pods/prometheus-0 "HTTP/1.1 200 OK"
unit-prometheus-0: 17:27:30 INFO unit.prometheus/0.juju-log HTTP Request: GET https://10.152.183.1/api/v1/namespaces/bar/pods/prometheus-0 "HTTP/1.1 200 OK"
unit-prometheus-0: 17:27:31 INFO unit.prometheus/0.juju-log HTTP Request: GET https://10.152.183.1/api/v1/namespaces/bar/persistentvolumeclaims/prometheus-database-88da8e7c-prometheus-0 "HTTP/1.1 200 OK"
unit-prometheus-0: 17:27:31 INFO unit.prometheus/0.juju-log HTTP Request: GET https://10.152.183.1/api/v1/namespaces/bar/pods/prometheus-0 "HTTP/1.1 200 OK"
[...]

PR #524, opened 2023-09-08
Fix handling of tls config

sed-i:
## Issue
TLS config is written and workload restarted, before the cert is written to disk.

## Solution
- [x] Use rock from https://github.com/canonical/prometheus-rock/pull/30
  - [x] Use the `update-ca-certificates` machinery
- [x] Instead of `stop()`, have `_scheme` return `http` and `_is_tls_enabled` return `False` until the cert is in place.

Depends on:
- https://github.com/canonical/prometheus-rock/pull/30 (https://github.com/canonical/oci-factory/pull/58)

Fixes https://github.com/canonical/prometheus-k8s-operator/issues/521

## Testing Instructions
- Run itests in https://github.com/canonical/cos-lite-bundle/pull/80
- Manually test with the following bundles

```yaml
bundle: kubernetes
applications:
  ca:
    charm: self-signed-certificates
    channel: edge
    scale: 1
  prom:
    charm: ./prometheus-k8s_ubuntu-20.04-amd64.charm
    series: focal
    scale: 1
    trust: true
    resources:
        prometheus-image: ghcr.io/canonical/prometheus:dev
relations:
- - ca:certificates
  - prom:certificates
```

```yaml
bundle: kubernetes
applications:
  ca:
    charm: self-signed-certificates
    channel: edge
    scale: 1
  cat:
    charm: catalogue-k8s
    channel: edge
    series: focal
    scale: 1
  prom:
    charm: ./prometheus-k8s_ubuntu-20.04-amd64.charm
    series: focal
    scale: 1
    trust: true
    resources:
        prometheus-image: ghcr.io/canonical/prometheus:dev
  trfk:
    charm: traefik-k8s
    channel: edge
    series: focal
    scale: 1
relations:
- - cat:catalogue
  - prom:catalogue
- - ca:certificates
  - prom:certificates
- - prom:ingress
  - trfk:ingress-per-unit
- - trfk:certificates
  - ca:certificates
```

## Release Notes
Fix handling of tls config.

### DIFF
--- a/lib/charms/tempo_k8s/v0/charm_tracing.py
+++ b/lib/charms/tempo_k8s/v0/charm_tracing.py
@@ -313,7 +313,7 @@ def trace_charm(
     method calls on instances of this class.
 
     Usage:
-    >>> from charms.tempo_k8s.v0.charm_tracing import trace_charm
+    >>> from charms.tempo_k8s.v0.charm_instrumentation import trace_charm
     >>> from charms.tempo_k8s.v0.tracing import TracingEndpointProvider
     >>> from ops import CharmBase
     >>>
@@ -370,7 +370,7 @@ def _autoinstrument(
 
     Usage:
 
-    >>> from charms.tempo_k8s.v0.charm_tracing import _autoinstrument
+    >>> from charms.tempo_k8s.v0.charm_instrumentation import _autoinstrument
     >>> from ops.main import main
     >>> _autoinstrument(
     >>>         MyCharm,
--- a/src/charm.py
+++ b/src/charm.py
@@ -401,7 +401,6 @@ class PrometheusCharm(CharmBase):
         )
 
     def _is_tls_ready(self) -> bool:
-        # self._update_cert()  # This setter has side effects to avoid code ordering issues.
         return (
             self.container.can_connect()
             and self.container.exists(CERT_PATH)
--- a/src/charm.py
+++ b/src/charm.py
@@ -143,8 +143,6 @@ class PrometheusCharm(CharmBase):
             relation_name="alertmanager",
         )
 
-        urlparse(self.external_url)
-
         self._scraping = MetricsEndpointProvider(
             self,
             relation_name="self-metrics-endpoint",
--- a/src/charm.py
+++ b/src/charm.py
@@ -133,7 +133,7 @@ class PrometheusCharm(CharmBase):
             port=self._port,
             strip_prefix=True,
             redirect_https=True,
-            scheme=lambda: "https" if self._is_cert_available() else "http",
+            scheme=lambda: "https" if self._is_tls_ready() else "http",
         )
 
         self._topology = JujuTopology.from_charm(self)
@@ -307,7 +307,7 @@ class PrometheusCharm(CharmBase):
             ],
         }
 
-        if self._is_cert_available():
+        if self._is_tls_ready():
             config.update(
                 {
                     "scheme": "https",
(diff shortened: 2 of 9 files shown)


sed-i:
@PietroPasotti, check-libs complains:
```
Library charms.tempo_k8s.v0.charm_tracing has local changes, cannot be updated.
```
Haven't checked but might be a LIBPATCH bump issue.

PietroPasotti:
> @PietroPasotti, check-libs complains:
> 
> ```
> Library charms.tempo_k8s.v0.charm_tracing has local changes, cannot be updated.
> ```
> 
> Haven't checked but might be a LIBPATCH bump issue.

see:
https://github.com/canonical/prometheus-k8s-operator/pull/523

src/charm.py
```diff
@@ -144,7 +143,7 @@ def __init__(self, *args):
             relation_name="alertmanager",
         )
 
-        external_url = urlparse(self.external_url)
+        urlparse(self.external_url)
```
PietroPasotti, line 146: this line is odd, we should add a comment to explain what's going on

src/charm.py
```diff
@@ -158,7 +157,7 @@ def __init__(self, *args):
                 self.cert_handler.on.cert_changed,
             ],
         )
-        self._prometheus_client = Prometheus(f"{external_url.scheme}://localhost:9090")
+        self._prometheus_client = Prometheus(self.internal_url)
```
PietroPasotti, line 160: ++

src/charm.py
```diff
@@ -539,7 +554,11 @@ def _configure(self, _):
 
         self.remote_write_provider.update_endpoint()
         self.grafana_source_provider.update_source(self.external_url)
-        self.unit.status = ActiveStatus()
+
+        if self._is_tls_enabled() and not self.container.exists(CERT_PATH):
```
PietroPasotti, line 558: I think it would be more clear if we abstracted this to a `self._is_tls_ready()` method. 
Enabled but not ready -> push cert to container, then restart. Regardless of how we got there, this should be the reasonable course of action. Or what am I missing?

src/charm.py
sed-i: Removed in next commit.

src/charm.py
sed-i: I'm thinking of writing the cert only on `cert-changed` event, and:

|                                  | Relation present | Cert file on disk                  | Hypothetical `is_tls_ready() -> bool` |
|----------------------------------|------------------|------------------------------------|-----------------------------|
| Isolated charm                   | No               | No                                 | No                          |
| tls-cert rel just joined         | Yes              | No                                 | No                          |
| on-cert-changed (after rel join) | Yes              | Yes                                | Yes                         |
| cert revoked/expired             | Yes              | Yes (keep on disk even if expired) | Yes                         |
| relation departed / broken       | Yes              | ?                                  | No (how? code ordering...)  |
| any event after rel-broken       | No              | No                                  | No |

Wdyt?

lib/charms/tempo_k8s/v0/charm_tracing.py
sed-i, line 1: This was a local change that I reverted:
```
Library charms.tempo_k8s.v0.charm_tracing has local changes, cannot be updated.
```

src/charm.py
PietroPasotti: seems reasonable. I think departed/broken should immediately remove the cert from disk and from that moment onwards is_tls_ready should return false (assuming it checks the file presence synchronously). 
Or we refactor to use manifests, but I'm not convinced that will make code ordering less of a headache

Abuelodelanada (approved):

PR #561, opened 2024-01-17
Set ActiveStatus when retention size config option is valid

### LINKED ISSUE #560: Stuck on BlockedStatus even after `maximum_retention_size` config option changed to valid
If we provide an invalid string to `maximum_retention_size`, the charm goes into BlockedStatus. Then, when it is corrected, there is no code in place to revert the status back to active.

https://github.com/canonical/prometheus-k8s-operator/blob/c03298b7736f2aa3623446c86014263a94bdd8c4/src/charm.py#L678-L683

sed-i:
## Issue
BlockedStatus from an invalid retention size config option was never cleared.

## Solution
- Add a TypedDict for per-component status setting.
- Use the per-component status dict to set status centrally in `_configure`.

Fixes #560.

### DIFF
--- a/src/charm.py
+++ b/src/charm.py
@@ -10,7 +10,7 @@ import re
 import socket
 import subprocess
 from pathlib import Path
-from typing import Dict, List, Optional, TypedDict, cast
+from typing import Dict, Optional, TypedDict, cast
 from urllib.parse import urlparse
 
 import yaml
@@ -46,6 +46,7 @@ from cosl import JujuTopology
 from lightkube.core.client import Client
 from lightkube.core.exceptions import ApiError as LightkubeApiError
 from lightkube.resources.core_v1 import PersistentVolumeClaim, Pod
+from ops import CollectStatusEvent, StoredState
 from ops.charm import ActionEvent, CharmBase
 from ops.main import main
 from ops.model import (
@@ -96,7 +97,6 @@ class CompositeStatus(TypedDict):
     """Per-component status holder."""
 
     retention_size: StatusBase
-    timespec: StatusBase
     k8s_patch: StatusBase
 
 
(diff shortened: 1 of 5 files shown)


src/charm.py
```diff
@@ -563,13 +575,9 @@ def _configure(self, _):
...
-        if (
-            isinstance(self.unit.status, BlockedStatus)
-            and self.unit.status not in early_return_statuses.values()
-        ):
-            return
-
-        self.unit.status = ActiveStatus()
+        self.unit.status = StatusBase._get_highest_priority(
+            cast(List[StatusBase], list(self.status.values()))
+        )
```
sed-i, line 580: @benhoyt, just drawing your attention to this status-related (ab)use, if you have a moment to take a look.

Currently there's a mix between centralized and decentralized status setting, and collect-status can't be used without major refactoring.

An alternative could be using the collect-status machinery to commit all values inside `self.status`.

PietroPasotti (approved):
lgtm

src/charm.py
benhoyt: Yeah, that sounds like a good direction to me too.

Additionally, I'd prefer charms didn't use private (underscore-prefixed) Ops methods like `StatusBase._get_highest_priority`. That said, I think that one's been useful enough for people doing various things with status that maybe we should make it public, either as `StatusBase.get_highest_priority` or another name. If you're interested in that, please open an issue for us to consider.

src/charm.py
```diff
@@ -91,6 +93,26 @@ class ConfigError(Exception):
...
+    """Per-component status holder."""
+
+    # These are going to go into stored state, so we must use marshallable objects.
+    # They are passed to StatusBase.from_name().
+    retention_size: Tuple[str, str]
+    k8s_patch: Tuple[str, str]
+    config: Tuple[str, str]
+
+
+def to_tuple(status: StatusBase) -> Tuple[str, str]:
```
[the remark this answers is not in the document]
sed-i: Could really appreciate rust traits here :)

I agree in general but here I think a shorter name is a better tradeoff: used only in this file, intention is immediately clear from either the definition or the usage, static checker would complain if misused.

PR #566, opened 2024-01-25
Fix topology for self-scraping

simskij:
## Issue
Currently, we populate the unit name label with the charm name, and discard the unit name.

## Solution

- Make `juju_unit` the unit name. 
- Add `juju_charm` and store the charm name there.

## Context

N/A

## Testing Instructions

Check the labels for Prometheus `up`1 and make sure it has the right values.

## Release Notes

Apply the correct Juju topology while self-scraping.


PietroPasotti (approved):
huh, seems wise.

dstathis (approved):

PR #574, opened 2024-02-08
Add timeout to container.exec

amandahla:
## Issue

For some reason, alertmanager and prometheus pods are getting constantly restarted and we couldnt figured out why. 

Versions:
Juju 3.1.6
Alertmanager latest/edge 98
Prometheus latest/edge 170

## Solution

I was not able to put my finger on it but this PR just adds timeout to the container.exec calls to make sure that this is not leaking and making pebble unresponsive.

## Context

Pebble

## Testing Instructions

No instructions. It happens randomly.

### DIFF
--- a/src/charm.py
+++ b/src/charm.py
@@ -775,7 +775,9 @@ class PrometheusCharm(CharmBase):
         Returns:
             A 2-tuple, (stdout, stderr).
         """
-        proc = self.container.exec(["/usr/bin/promtool", "check", "config", PROMETHEUS_CONFIG], timeout=3)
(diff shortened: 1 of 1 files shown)


sed-i:
Thanks @amandahla and sorry for the long delay.
- Do you still find the timeout useful?
- Could 3 sec be too short?
- Iiuc, a timeout would put the charm in error state. Not sure which approach is best.

simskij:
This has been stale for more than six months, so I will go ahead and close this. If you still think this is relevant, please open a new PR and answer the question stated in the last comment above.

PR #587, opened 2024-04-11
fix: tell type checkers that the config options are strings

tonyandrewmeyer:
## Issue

ops<=2.12 wrongly has `self.config[x]` typed as `str`, when actually it could be an int, float, bool, or str, depending on the config type. We're fixing this in [ops:#1183](https://github.com/canonical/operator/pull/1183), but that will break static checking that currently assumes that the config is a str (because ops doesn't validate the schema, so all options will be bool|int|float|str).

## Solution

Six `typing.cast` calls where the config values are loaded and used where the type should be str.

## Context

Covered above.

## Testing Instructions

Run `tox -e static-charm` with/without the PR with ops 2.12 and with either the PR branch linked above or once that's merged ops main/2.13.

## Upgrade Notes

N/A

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,7 +1,7 @@
 cosl
 cryptography
 jsonschema
-ops
+ops @ git+https://github.com/tonyandrewmeyer/operator@fix-config-types-1182
 pyaml
 requests
 lightkube >= 0.11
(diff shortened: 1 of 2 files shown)


tonyandrewmeyer:
Let me know if there's anything else I need to do here - if we can get this merged then we can enable the CI tests in ops against this charm, which would be great :). Thanks!

lucabello (approved):

Issue #578, opened 2024-02-14
Set external URL as a Grafana Data Source


Abuelodelanada:
### Enhancement Proposal

Unlike Loki, if the relation between Prometheus and Traefik is established, Prometheus set fqdn address instead of the external one as a Grafana Data Source.

![image](https://github.com/canonical/prometheus-k8s-operator/assets/939888/84181b78-cacd-4871-9db6-ccfddca83029)

PietroPasotti:
picking this up as it breaks TLS integration

PietroPasotti:
I've taken a look at the context in https://github.com/canonical/operator/issues/970 and IIUC we don't need to check for CMR because even for model-internal grafana-source relations, we still need ingress URL as FQDN will not work with TLS because the cert is based on traefik's external hostname.

So it seems that  the deployment we're after is:

```mermaid
graph LR

subgraph COS model
  prometheus ---|ingress URL| grafana

end
external-prometheus ---|ingress URL| grafana

```

PietroPasotti:
if not we can use something like

```python


# cfr: https://github.com/canonical/operator/pull/1096/files
# cfr: https://github.com/canonical/operator/issues/970
def _is_cmr(relation: Relation):
    """Temporary workaround for checking if the given relation is a cross-model relation.
    Note that this property will return a false positive if a local related app happens to be
    named in the same pattern juju names CMRs, e.g. "remote-abc".
    """
    # Units in cross model relations have names such as
    # "remote-c87d7acb413449cd8097b523af7ff830/0".
    if not relation.units:
        raise ModelError("No units in relation, so cannot determine if cross-model.")

    return any(re.match(r"remote-[a-f0-9]+/", unit.name) for unit in relation.units)

```

PietroPasotti:
if we do this, TLS stops working.

PR #600, opened 2024-05-02
internal url in datasource https

PietroPasotti:
Doing the opposite of #578 
- sets internal url in datasource, plus https scheme if we have TLS.

To test:
deploy with cos-lite (with and without self-signed-certs) and cross-relate.
Verify that the datasources work from grafana.


src/charm.py
sed-i, line 1: Is there anything missing? There are no functional changes here atm.

PR #632, opened 2024-07-16
feat: use external_url in grafana source

### LINKED ISSUE #578: Set external URL as a Grafana Data Source

lucabello:
## Issue
Closes #578.

## Solution
Use `external_url` in Grafana Source.


mmkay (approved):
The changeset seems reasonable, however I'm wondering if it would be good to add a bit of testing around which URL is used in the source. What do you think?

PR #634, opened 2024-07-19
Compress rules, jobs for cos-proxy

sed-i:
## Issue
Some productions deployments have several megs of relation data just for alert rules, scrape jobs.
Contributing factors to the high volume of reldata:
- Scrape jobs are repeated [per unit per parameter](https://github.com/canonical/cos-proxy-operator/blob/a09087f9b72e46a79f1c023c877fe81488cafc3f/lib/charms/nrpe_exporter/v0/nrpe_exporter.py#L543).
- Alert rules are replicated per unit.

As a result, every event cos-proxy gets from nrpe relation-changed, results in reading/writing several megs of data, taking several seconds to complete. When many units are at play, model settling takes a long time.

In addition, relation data limit is 16M, and in common envs we already approach 25% of that.

## Solution
- LZMA-compress rules, jobs before storing in reldata, but only in MetricsEndpointAggregator.
- Decode in MetricsEndpointConsumer.
- Use encode/decode methods [from grafana](https://github.com/canonical/grafana-k8s-operator/blob/a2688ade7af5d0dc1607c1ac560faf065966f063/lib/charms/grafana_k8s/v0/grafana_dashboard.py#L547).

TODO:
- Since we're bumping LIBAPI here, could consider addressing some of the TODO/FIXMEs sprinkled in the code.

## Context
- https://github.com/canonical/cos-proxy-operator/issues/56
- https://github.com/canonical/cos-proxy-operator/pull/139


sed-i:
As of writing, the diff between v0 and v1 is:
image

sed-i:
Superseded by #864.

PR #635, opened 2024-08-14
fix: use external_url in alert source

### LINKED ISSUE #581: Internal url is rendered as the "source" in alerts
### Bug Description

> If i go to alertmanager, check an alert and click on the Source button, it will want to take me to `http://prometheus-0.cos.svc.kubernetes.cluster:9090` (or something along those lines), but i have a traefik loadbalancer in front of both alertmanager and prometheus, so i would expect the Source button to take me to `http://$LOADBALANCER_IP/cos-prometheus-0` instead.
>
> The ingress url is there alright, but the command used to launch prometheus points to the internal DNS name
> ```bash
> # kubectl -n cos exec prometheus-0 -c prometheus -- ps uax | grep -oE -- '--web.external-url=[a-z0-9:/\.\-]*'
> --web.external-url=http://prometheus-0.prometheus-endpoints.cos.svc.cluster.local:9090/cos-prometheus-0
> ```
> This comes from charm code:
> ```python
>         route_prefix = urlparse(self.external_url).path.strip("/")
>         external_url = f"{self.internal_url.rstrip('/')}/{route_prefix}".rstrip("/")
> ```
> -- @peterctl

### To Reproduce

traefik - prometheus - alertmanager

### Environment

TBD

### Relevant log output

```shell
N/A
```

### Additional context

We may be able to address this with label rewrite.

lucabello:
## Issue
Closes #581.  
Closes https://github.com/canonical/alertmanager-k8s-operator/issues/206.

Not sure why we were using the `route_prefix` from the `self.external_url` (from ingress), but adding it to the `self.internal_url` (which is always the FQDN).

## Solution

Use the external URL instead of the internal one.

### Screenshots

Before:
Screenshot-2024-08-14_10-23

After:
Screenshot-2024-08-14_10-39

## Testing Instructions

```bash
juju deploy alertmanager-k8s --channel=latest/edge alertmanager
juju deploy prometheus-k8s --channel=latest/edge prometheus
juju deploy traefik-k8s --channel=latest/edge traefik

juju relate prometheus:alertmanager alertmanager
juju relate prometheus:metrics-endpoint alertmanager
# observe the watchdog alert's source link (internal url)
juju relate prometheus:ingress traefik
# wait for things to settle
# observe the same alert's source link again and try to click it!
```

## Upgrade Notes

Anything that refreshes the Pebble plan will activate this change, so the upgrade should be seamless.

### DIFF
--- a/charmcraft.yaml
+++ b/charmcraft.yaml
@@ -13,6 +13,7 @@ parts:
     build-packages:
       - git
     charm-binary-python-packages:
+      - pip>=24
       - jsonschema
       - cryptography
       - pyyaml
@@ -20,7 +21,7 @@ parts:
       - ops
       - wheel==0.37.1
       - setuptools==45.2.0
-      - pydantic>=2
+      - pydantic
       - pydantic-core
   cos-tool:
     plugin: dump
--- a/charmcraft.yaml
+++ b/charmcraft.yaml
@@ -12,6 +12,8 @@ parts:
   charm:
     build-packages:
       - git
+      - rustc
+      - cargo
     charm-binary-python-packages:
       - pip>=24
       - jsonschema
--- a/charmcraft.yaml
+++ b/charmcraft.yaml
@@ -21,6 +21,7 @@ parts:
       - wheel==0.37.1
       - setuptools==45.2.0
       - pydantic>=2
+      - pydantic-core
   cos-tool:
     plugin: dump
     source: .
--- a/tests/integration/test_prometheus_scrape_multiunit.py
+++ b/tests/integration/test_prometheus_scrape_multiunit.py
@@ -48,13 +48,14 @@ num_units = 2  # Using the same number of units for both prometheus and the test
 idle_period = 90
 
 
+@pytest.mark.skip(reason="xfail")
 async def test_setup_env(ops_test: OpsTest):
     await ops_test.model.set_config(
         {"logging-config": "<root>=WARNING; unit=DEBUG", "update-status-hook-interval": "60m"}
     )
 
 
-@pytest.mark.xfail
+@pytest.mark.skip(reason="xfail")
 async def test_prometheus_scrape_relation_with_prometheus_tester(
     ops_test: OpsTest, prometheus_charm, prometheus_tester_charm
 ):
@@ -179,7 +180,7 @@ async def test_prometheus_scrape_relation_with_prometheus_tester(
         )
 
 
-@pytest.mark.xfail
+@pytest.mark.skip(reason="xfail")
 async def test_upgrade_prometheus(ops_test: OpsTest, prometheus_charm):
     """Upgrade prometheus and confirm all is still green (see also test_upgrade_charm.py)."""
     # GIVEN an existing "up" timeseries
@@ -221,7 +222,7 @@ async def test_upgrade_prometheus(ops_test: OpsTest, prometheus_charm):
     assert all(up_before[i] <= up_after[i] for i in range(num_units))
 
 
-@pytest.mark.xfail
+@pytest.mark.skip(reason="xfail")
 async def test_rescale_prometheus(ops_test: OpsTest):
     # GitHub runner doesn't have enough resources to deploy 3 unit with the default "requests", and
     # the unit fails to schedule. Setting a low limit, so it is able to schedule.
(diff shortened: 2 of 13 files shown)


PietroPasotti (changes requested):
*cough* tests *cough*

lucabello:
A unit test has been fixed :)

I also skipped the tests we were XFail-ing, because they'll just waste time in integration tests without contributing :)

tests/integration/test_prometheus_scrape_multiunit.py
```diff
@@ -48,13 +48,14 @@
 idle_period = 90
 
 
+@pytest.mark.skip(reason="xfail")
```
PietroPasotti, line 51: this feels bad. 'xfail' is not a reason for skipping a test.
1) why are we skipping it instead of removing it altogether?
2) should we be xfailing it instead? 

Also applies below

tox.ini
```diff
@@ -68,6 +68,7 @@ deps =
     cosl
     fs
     pytest
+    pytest-asyncio
```
PietroPasotti, line 71: let's add a comment about why we need this

PietroPasotti (approved):
In principle the change is OK and given it's high prio I don't want to block it, but we're skipping a huge amount of tests for unclear reasons. 
Can you explain what's going on and remove any tests that are no longer necessary?

tests/integration/test_prometheus_scrape_multiunit.py
lucabello: I can remove the change from this PR if you want, to keep it cleaner, but imho `xfail` is a reason for skipping a test; if we don't care about the test passing, isn't it running it a waste of time?

I don't want to remove it because if we rewrite tests, it could be useful to have them there, but the `skip(reason="we were xfailing this")` sounds like a good reason to me

However if you disagree, I'll remove it and we'll talk about it in more detail

tox.ini
lucabello: I don't remember which warning it was clearing :)

charmcraft.yaml
```diff
@@ -12,15 +12,19 @@ parts:
...
       - jsonschema
       - cryptography
       - pyyaml
       - requests
       - ops
       - wheel==0.37.1
       - setuptools==45.2.0
-      - pydantic>=2
+      - pydantic
+      - pydantic-core
```
sed-i, line 27: Should be bring back the `>=2` pin?

tests/integration/test_prometheus_scrape_multiunit.py
```diff
@@ -48,13 +48,14 @@
...
 async def test_setup_env(ops_test: OpsTest):
     await ops_test.model.set_config(
         {"logging-config": "<root>=WARNING; unit=DEBUG", "update-status-hook-interval": "60m"}
     )
 
 
-@pytest.mark.xfail
+@pytest.mark.skip(reason="xfail")
```
sed-i, line 58: What's the difference between `@pytest.mark.xfail` and `@pytest.mark.skip(reason="xfail")`?
And why do we suddenly have so many skips?

sed-i (approved):

charmcraft.yaml
lucabello: Those binary packages are not doing much anyway, for some reason `charcmraft` still builds them from scratch; my current guess is that `PYDEPS` are always rebuilt

tests/integration/test_prometheus_scrape_multiunit.py
lucabello: `@pytest.mark.xfail` executes the tests, but ignores the end result.
`@pytest.mark.skip` skips the tests, just like commenting it out.

The purpose of `xfail` is, generally, to mark tests that **should**, but don't pass. Ideally, you would temporarily mark something with `xfail`, understand what the problem is, and fix it.

If we had a test suddenly failing only in CI, we'd probably `xfail` it until we fix it.

However, what we do with `xfail` is use it to essentially "not care about the result of a test" without the guilt of removing it. Our oldest `xfail` of this set is from November 2022, and to my knowledge we never went to revisit an `xfail`ing test, to the point I'm sure no one would be able to tell you if they're passing or not.

Instead of deleting them though, we can just `skip` until we decide what to do with them; it's hard to see (due to `charmcraft pack` taking 10 minutes extra), but we improved the itests execution time by about 10 minutes.

PR #639, opened 2024-09-11
tests: remove ops.testing patch that is no longer required

tonyandrewmeyer:
## Issue

The tests currently patch a private class in ops.testing, which will no longer work in an upcoming version of ops (no compatibility guarantees are provided for private objects).

## Solution

It seems like the patching is no longer required with the way that `network_get` is handled with Harness now, so it's removed. In a small number of cases a specific network is added to the Harness instead.

## Context

ops?

## Testing Instructions

Run the unit tests.

## Upgrade Notes

N/A


dimaqq:
poke @lucabello

dimaqq:
poke @simskij

PR #641, opened 2024-10-09
Migrate tracing libs

PietroPasotti:
migrate tracing libs to tempo-coordinator


mmkay:
@lucabello not for the scope of this PR, but it's worth looking at changes from #639 that would hopefully unblock ops 2.16. For ops 2.17 I think the full scenario 7 migration will be needed similar to https://github.com/canonical/traefik-k8s-operator/pull/405

mmkay (approved):

PR #647, opened 2024-11-25
forward grafana source UIDs

PietroPasotti:
## Issue
the `grafana_datasource` interface has been updated to give prometheus back the datasource UID provisioned for its units.
This PR instructs prometheus to forward those UIDs over remote-write, so that applications remote-writing to prometheus, such as Tempo or Loki, can obtain the datasource UIDs that have been created for their datasources.

Tandem PR: https://github.com/canonical/grafana-k8s-operator/pull/359

### DIFF
--- a/lib/charms/grafana_k8s/v0/grafana_source.py
+++ b/lib/charms/grafana_k8s/v0/grafana_source.py
@@ -432,7 +432,8 @@ class GrafanaSourceProvider(Object):
     def get_source_uids(self) -> Dict[str, Dict[str, str]]:
         """Get the datasource UID(s) assigned by the remote end(s) to this datasource.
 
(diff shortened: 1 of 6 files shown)


mmkay:
I get the intention behind this but it seems like this might be much cleaner if we went the direction with a separate interface (`datasource_exchange` that would be a relation between prometheus and tempo / loki). Setting `datasource_uids` in the remote_write relation seems to be adding a field that is coming from an entirely different relation that might even not be a concern for someone that doesn't relate to grafana.

simskij:
I agree with @mmkay on this one, and as such, will close this pr.

PietroPasotti:
this must have been my feverish brain, but yes.

PR #664, opened 2025-01-13
fix: update cos-tool permissions to adhere to cis hardening rules

lucabello:
## Issue
Needed for the following PR: https://github.com/canonical/grafana-agent-operator/pull/231
More info on the CIS hardening rule can be found here[^1].

## Solution
Use `path.chmod(0o775)` in order to comply wit the CIS Hardening rules.

[^1]: https://github.com/canonical/grafana-agent-operator/issues/208

### DIFF
--- a/lib/charms/prometheus_k8s/v0/prometheus_scrape.py
+++ b/lib/charms/prometheus_k8s/v0/prometheus_scrape.py
@@ -362,7 +362,7 @@ LIBAPI = 0
 
 # Increment this PATCH version before using `charmcraft publish-lib` or reset
 # to 0 if you are raising the major API version
-LIBPATCH = 47
+LIBPATCH = 48
 
 PYDEPS = ["cosl"]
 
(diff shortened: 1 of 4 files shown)


lib/charms/prometheus_k8s/v0/prometheus_scrape.py
```diff
@@ -2365,7 +2365,7 @@ def _get_tool_path(self) -> Optional[Path]:
         res = "cos-tool-{}".format(arch)
         try:
             path = Path(res).resolve()
-            path.chmod(0o777)
+            path.chmod(0o775)
```
sed-i, line 2368: Should we do this in charmcraft.yaml?
https://github.com/canonical/prometheus-k8s-operator/blob/62cd2d9b8c3a528910ce12c553209817e1b7a889/charmcraft.yaml#L35

lucabello: We can do it *also* in `charmcraft.yaml` (to address the CIS Hardening rule for this charm), but not exclusively; otherwise, the library will set the permission to `0o775` for every charm that uses it.

sed-i (approved):

PR #673, opened 2025-02-04
feat: add support for disabling alert rules forwarding

lucabello:
## Issue
First step in addressing https://github.com/canonical/grafana-agent-operator/issues/154

## Solution
Allow charm authors to pass `forward_alert_rules=False` to both `MetricsEndpointProvider` (scrape) and `PrometheusRemoteWriteConsumer` (remote write).

When the flag is `True`, the alert rules information is simply not put in relation data.

## Testing Instructions
Pack and deploy both this charm and grafana-agent from https://github.com/canonical/grafana-agent-operator/pull/252 , and test the new config option.

### prometheus_scrape

Screenshot-2025-02-05_10-12

### prometheus_remote_write

Screenshot-2025-02-05_10-33

### DIFF
--- a/lib/charms/prometheus_k8s/v0/prometheus_scrape.py
+++ b/lib/charms/prometheus_k8s/v0/prometheus_scrape.py
@@ -1847,6 +1847,9 @@ class MetricsEndpointAggregator(Object):
         self.framework.observe(alert_rule_events.relation_changed, self._on_alert_rules_changed)
         self.framework.observe(alert_rule_events.relation_departed, self._on_alert_rules_departed)
 
+        # refresh alert rules on config-changed
+        self.framework.observe(self._charm.on.config_changed, self._on_alert_rules_changed)
+
     def _set_prometheus_data(self, event):
         """Ensure every new Prometheus instances is updated.
 
--- a/lib/charms/prometheus_k8s/v0/prometheus_scrape.py
+++ b/lib/charms/prometheus_k8s/v0/prometheus_scrape.py
@@ -1847,8 +1847,39 @@ class MetricsEndpointAggregator(Object):
         self.framework.observe(alert_rule_events.relation_changed, self._on_alert_rules_changed)
         self.framework.observe(alert_rule_events.relation_departed, self._on_alert_rules_departed)
 
-        # refresh alert rules on config-changed
-        self.framework.observe(self._charm.on.config_changed, self._on_alert_rules_changed)
+    def reset_prometheus_alerts(self):
+        """Re-set Prometheus alert rules."""
+        if not self._charm.unit.is_leader():
+            return
+
+        for prometheus_relation in self.model.relations[self._prometheus_relation]:
+            # Gather the alert rules
+            groups = []
+            if self._stored.alert_rules:
+                groups += _type_convert_stored(
+                    self._stored.alert_rules  # pyright: ignore
+                )  # list of alert rule groups
+            for relation in self.model.relations[self._alert_rules_relation]:
+                unit_rules = self._get_alert_rules(relation)
+                if unit_rules and relation.app:
+                    appname = relation.app.name
+                    rules = self._label_alert_rules(unit_rules, appname)
+                    group = {"name": self.group_name(appname), "rules": rules}
+                    groups.append(group)
+            alert_rules = AlertRules(query_type="promql", topology=self.topology)
+            # Add alert rules from file
+            if self.path_to_own_alert_rules:
+                alert_rules.add_path(self.path_to_own_alert_rules, recursive=True)
+            # Add generic alert rules
+            alert_rules.add(
+                generic_alert_groups.application_rules, group_name_prefix=self.topology.identifier
+            )
+            groups.extend(alert_rules.as_dict()["groups"])
+
+            # Set scrape jobs and alert rules in relation data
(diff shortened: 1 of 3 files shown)


lib/charms/prometheus_k8s/v0/prometheus_scrape.py
```diff
@@ -1306,6 +1306,7 @@ def __init__(
         relation_name: str = DEFAULT_RELATION_NAME,
         jobs=None,
         alert_rules_path: str = DEFAULT_ALERT_RULES_RELATIVE_PATH,
+        disable_forwarding_alert_rules: bool = False,
```
sed-i, line 1309: - Update PR description accordingly?
- Should we rename it to `forward_alert_rules`? `enable=True` seems easier than `disable=False`.
- We do not have a [`, *,`](https://peps.python.org/pep-3102/) here, so this new arg must  be last. Should probably add a `*,` before it.

lib/charms/prometheus_k8s/v0/prometheus_scrape.py
```diff
@@ -1530,8 +1534,10 @@ def set_scrape_job_spec(self, _=None):
             return
 
         alert_rules = AlertRules(query_type="promql", topology=self.topology)
-        alert_rules.add_path(self._alert_rules_path, recursive=True)
+        if not self._disable_alerts:
+            alert_rules.add_path(self._alert_rules_path, recursive=True)
         alert_rules_as_dict = alert_rules.as_dict()
+        logger.info(f"+++ alert_rules_as_dict: {alert_rules_as_dict}")  # TODO: remove, debug
```
sed-i, line 1540: Reminder to remove.

lib/charms/prometheus_k8s/v1/prometheus_remote_write.py
```diff
@@ -401,6 +401,7 @@ def __init__(
         charm: CharmBase,
         relation_name: str = DEFAULT_CONSUMER_NAME,
         alert_rules_path: str = DEFAULT_ALERT_RULES_RELATIVE_PATH,
+        disable_forwarding_alert_rules: bool = False,
```
sed-i, line 404: Same comments as above.

lib/charms/prometheus_k8s/v1/prometheus_remote_write.py
```diff
@@ -453,6 +456,9 @@ def __init__(
         self.framework.observe(
             self._charm.on.upgrade_charm, self._push_alerts_to_all_relation_databags
         )
+        self.framework.observe(
+            self._charm.on.config_changed, self._push_alerts_to_all_relation_databags
+        )
```
sed-i, line 461: Do we need to observe config-changed in the scrape lib as well? Or, do we need to observe config_changed here? This information coming from a config option is a charm impl detail we shouldn't leak.

lib/charms/prometheus_k8s/v1/prometheus_remote_write.py
```diff
@@ -484,11 +490,12 @@ def _push_alerts_to_relation_databag(self, relation: Relation) -> None:
             return
 
         alert_rules = AlertRules(query_type="promql", topology=self.topology)
-        alert_rules.add_path(self._alert_rules_path)
+        if not self._disable_alerts:
+            alert_rules.add_path(self._alert_rules_path)
```
sed-i, line 494: Is there any code path you can think of where in the constructor we now flipped from one bool to another, but `_push_alerts_to_relation_databag` isn't called?

In other words
- should the lib update reldata on construction?
- do we need a "recreate the world" function in the lib?

Same question applies to the scrape lib.

lib/charms/prometheus_k8s/v0/prometheus_scrape.py
lucabello: Done

lib/charms/prometheus_k8s/v0/prometheus_scrape.py
lucabello: Done

lib/charms/prometheus_k8s/v1/prometheus_remote_write.py
lucabello: It's only necessary in Prometheus remote-write. For scrape, alerts end up in Prometheus' unit data; for remote-write, it's in the other charm's unit data. We're already observing **leader-elected** and **upgrade-charm**, I don't think this changes much.

Changing to a generic `refresh_event` like we have in other libraries :)

sed-i (approved):

MichaelThamm (changes requested):

lib/charms/prometheus_k8s/v0/prometheus_scrape.py
```diff
@@ -2198,8 +2212,8 @@ def remove_alert_rules(self, group_name: str, unit_name: str) -> None:
                 changed_group["rules"] = rules_kept  # type: ignore
                 groups.append(changed_group)
 
-            relation.data[self._charm.app]["alert_rules"] = (
```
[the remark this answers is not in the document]
lucabello: It's functionally equivalent, and in some other functions we in fact return `{'groups': []}` already. I also manually tested (other than adding scenario tests in cos-proxy) and everything works fine :)

sed-i (approved):

PR #685, opened 2025-03-14
refactor: refactor: change PrometheusCharm.external_url, add most_external_url

ca-scribner:
## Issue
Previously, PrometheusCharm.external_url returned the external url if it was available, otherwise it returned the ingress url. This meant that any consuming code that explicitly wanted the external_url or nothing needed to duplicate checks that exist in external_url, making it easy to later have these go out of sync.

## Solution
This change breaks external_url into two functions:

* external_url, which explicitly returns only an external url or None (if it isn't available)
* most_external_url, which returns the external_url if it is available, otherwise the internal url

This is useful for a planned future PR which will need exactly the external_url

## Context
none

## Testing Instructions
change is covered by existing unit tests

## Upgrade Notes
no effect

### DIFF
--- a/src/charm.py
+++ b/src/charm.py
@@ -192,7 +192,7 @@ class PrometheusCharm(CharmBase):
             self,
             relation_name="self-metrics-endpoint",
             jobs=self.self_scraping_job,
-            external_url=self.external_url,
+            external_url=self.most_external_url,
             refresh_event=[  # needed for ingress
                 self.ingress.on.ready_for_unit,
                 self.ingress.on.revoked_for_unit,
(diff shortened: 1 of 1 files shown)


src/charm.py
```diff
@@ -205,14 +205,14 @@ def __init__(self, *args):
         self.remote_write_provider = PrometheusRemoteWriteProvider(
             charm=self,
             relation_name=DEFAULT_REMOTE_WRITE_RELATION_NAME,
-            server_url_func=lambda: PrometheusCharm.external_url.fget(self),  # type: ignore
+            server_url_func=lambda: PrometheusCharm.most_external_url.fget(self),  # type: ignore
```
MichaelThamm, line 208: 1. naming `external`, `most_external` is bad imo. What's next `mostest_external` (exaggeration for the point).
2. You are adding new functionality. 
> change is covered by existing unit tests
We should add checks (if they don't exist) for:
- `external_url` returns only an external url or None (if it isn't available)
- `most_external_url` returns the external_url if it is available, otherwise the internal url

sed-i: I think what Andrew means here is effectively max(urls), if urls had a comparator function :sweat_smile:

sed-i (approved):

src/charm.py
ca-scribner: exactly!  

the complication here is that, prior to this PR, `external_url` **already means** `most_external`.  It doesn't return an external url, it returns the external url if it exists, otherwise internal.

You're welcome to name it differently (eg: `external_url_if_exists_otherwise_internal_url`, although that one is kinda long haha), but I don't think the old `external_url` is accurately named since it might give you the internal url

PR #687, opened 2025-03-14
feat: add prometheus-api lib, relation

ca-scribner:
## Issue
The Prometheus charm does not provide a way for other charms to consume its prometheus api for querying metrics (apart from the grafana-source interface, which is an opinionated interface specific to grafana consuming prometheus).

## Solution
This PR uses Mimir's prometheus-api library to send Prometheus's API to other applications, enabling it to be consumed by other charms such as Kiali.  

## Context
This implementation is essentially identical to that of:
* [mimir-coordinator](https://github.com/canonical/mimir-coordinator-k8s-operator/pull/115) (except that PR both **implements** and uses the prometheus-api library)

Merge #685 before merging this

## Testing Instructions
Automated tests cover everything changed here.  

## Upgrade Notes
No upgrade considerations

### DIFF
--- a/lib/charms/mimir_coordinator_k8s/v0/prometheus_api.py
+++ b/lib/charms/mimir_coordinator_k8s/v0/prometheus_api.py
@@ -1,0 +1,226 @@
+"""prometheus_api.
+
+This library implements endpoint wrappers for the prometheus-api interface.  The prometheus-api interface is used to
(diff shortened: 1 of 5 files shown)


ca-scribner:
left this as draft just until I've tested it live connected to Kiali.  Its in a reviewable state now, but that test will catch if I typo'd the url or something silly like that

sed-i (approved):

PR #705, opened 2025-06-20
feat: deduplicate remote_write addresses to avoid consumer crashes

lucabello:
## Issue
Supersedes https://github.com/canonical/grafana-agent-k8s-operator/pull/383 (track/1) and https://github.com/canonical/grafana-agent-k8s-operator/pull/384 (main).

Duplication of the remote write endpoints causes consumer charms like grafana-agent to crash.

## Solution
Deduplicate in the library itself instead of Grafana Agent.

## Testing Instructions
I modified the `test_remote_write.py` unit test to also verify this deduplication works.


lib/charms/prometheus_k8s/v1/prometheus_remote_write.py
```diff
@@ -510,10 +510,25 @@ def reload_alerts(self) -> None:
         """Reload alert rules from disk and push to relation data."""
         self._push_alerts_to_all_relation_databags(None)
 
+    def _deduplicate_endpoints(self, endpoints: List[Dict[str, Any]]) -> List[Dict[str, Any]]:
```
PietroPasotti, line 513: I may be wrong, but this is equivalent to `[dict(t) for t in {tuple(d.items()) for d in endpoints}]`

lib/charms/prometheus_k8s/v1/prometheus_remote_write.py
```diff
@@ -537,7 +552,8 @@ def endpoints(self) -> List[Dict[str, str]]:
                         }
                     )
 
-        return endpoints
+        deduplicated_endpoints = self._deduplicate_endpoints(endpoints)
```
PietroPasotti, line 555:
```suggestion
        # deduplicate the endpoints by converting each dict to a tuple of dict.items(), throwing them into a set, and then chucking them back to dicts
        deduplicated_endpoints = [dict(t) for t in {tuple(d.items()) for d in endpoints}]
```

PietroPasotti (approved):
one tip about the implementation, but the rest looks good

lib/charms/prometheus_k8s/v1/prometheus_remote_write.py
lucabello: Good point, done!

PR #719, opened 2025-08-29
feat: block Prometheus charm when available disk space is insufficient

### LINKED ISSUE #715: Better juju status message when `/var/lib/prometheus`  is full
### Enhancement Proposal

Currently when `/var/lib/prometheus` is full the following message is shown in the output of `juju status`.

```
prometheus/0*              blocked   idle       172.16.1.135         Prometheus failed to restart (config valid?); see debug logs
```

`juju debug-log` eventually shows the following:

```
caller=main.go:1270 level=error err="opening storage failed: lock DB directory: open /var/lib/prometheus/lock: no space left on device"
```

### LINKED ISSUE #717: Implement disk.rules equivalent for prometheus itself
### Enhancement Proposal

Currently we have no alert rules for when prometheus's own disk fills up.

In grafana-agent (vm) we have [disk.rules](https://github.com/canonical/grafana-agent-operator/blob/main/src/prometheus_alert_rules/disk.rules), but those do not apply to prometheus because there are no node-exporter metrics coming from the prometheus pod.

Potential solutions (not mutually exclusive):
- Add a node-exporter container, similar to what [we already have in loki](https://github.com/canonical/loki-k8s-operator/pull/373). And add alert rules.
- On update status, change to Blocked if we have less than 1Gi left.
- Rely on k8s integration and [their alerts](https://github.com/canonical/k8s-operator/blob/808ca91103db71ee7597a10f6173ec55af7b5e0d/charms/worker/k8s/src/prometheus_alert_rules/kubernetesControlPlane-prometheusRule.yaml#L697).

sinapah:
## Issue

This PR fixes #717, fixes #715.

PR #720 should be merged into this PR before this is merged into `main`.

## Solution

On the `update_status` hook and `pebble_ready`, it checks the disk space left for Prometheus' database storage. If less than 1 GiB (to be exact, 1024**3) remains, the charm state is set to blocked, along with a message that says `<1 GiB disk space remaining`.

Note that this PR touches some extra files due to the switch to `ruff format` in `tox.ini` as recommended by @lucabello.

## Context

Unless the user uses an overlay when deploying Prometheus, by default they get 1 GiB of storage. This is insufficient in most cases as it tends to run out quickly. As a result of this PR, the charm will always get set to blocked when not deployed with an overlay, until we start setting a higher disk space for this charm in its `charmcraft.yaml`.

## Testing Instructions

You can rely on the unit tests for testing.

### DIFF
--- a/src/charm.py
+++ b/src/charm.py
@@ -590,6 +590,7 @@ class PrometheusCharm(CharmBase):
             database_storage = self.model.storages.get('database', [])
             if database_storage and shutil.disk_usage(database_storage[0].location).free < 1e9: # type: ignore
                 self._stored.status["disk_space"] = to_tuple(BlockedStatus("Less than 1 Gi remaining in storage"))
+
         # If this check is done before storage is attached, we don't want the charm to go error state
         except FileNotFoundError:
             self.unit.status = BlockedStatus("Storage not available")
--- a/src/charm.py
+++ b/src/charm.py
@@ -752,7 +752,7 @@ class PrometheusCharm(CharmBase):
             self._configure(event)
         try:
             database_storage = self.model.storages.get('database', [])
-            if database_storage and shutil.disk_usage(database_storage.location).free < 1e9:
+            if database_storage and shutil.disk_usage(database_storage.location).free < 1e9: # type: ignore
                 self._stored.status["config"] = to_tuple(BlockedStatus("Less than 1 Gi remaining in storage"))
         # If this check is done before storage is attached, we don't want the charm to go error state
         except FileNotFoundError:
--- a/src/charm.py
+++ b/src/charm.py
@@ -129,6 +129,7 @@ def to_status(tpl: Tuple[str, str]) -> StatusBase:
     name, message = tpl
     return StatusBase.from_name(name, message)
 
+
 @dataclass
 class TLSConfig:
     """TLS configuration received by the charm over the `certificates` relation."""
@@ -136,6 +137,7 @@ class TLSConfig:
     server_cert: str
     ca_cert: str
     private_key: str
+
 
 @trace_charm(
     tracing_endpoint="charm_tracing_endpoint",
@@ -269,7 +271,9 @@ class PrometheusCharm(CharmBase):
         self.framework.observe(self.on.update_status, self._update_status)
         self.framework.observe(self.ingress.on.ready_for_unit, self._on_ingress_ready)
         self.framework.observe(self.ingress.on.revoked_for_unit, self._on_ingress_revoked)
-        self.framework.observe(self._cert_requirer.on.certificate_available, self._on_certificate_available)
+        self.framework.observe(
+            self._cert_requirer.on.certificate_available, self._on_certificate_available
+        )
         self.framework.observe(self.remote_write_provider.on.alert_rules_changed, self._configure)
         self.framework.observe(self.remote_write_provider.on.consumers_changed, self._configure)
         self.framework.observe(self.metrics_consumer.on.targets_changed, self._configure)
@@ -573,7 +577,9 @@ class PrometheusCharm(CharmBase):
 
             # Repeat for the charm container. We need it there for prometheus client requests.
             ca_cert_path.parent.mkdir(exist_ok=True, parents=True)
-            ca_cert_path.write_text(tls_config.ca_cert,)  # pyright: ignore
+            ca_cert_path.write_text(
+                tls_config.ca_cert,
+            )  # pyright: ignore
         else:
             self.container.remove_path(CERT_PATH, recursive=True)
             self.container.remove_path(KEY_PATH, recursive=True)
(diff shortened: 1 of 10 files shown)


src/charm.py
```diff
@@ -584,6 +585,16 @@ def _update_cert(self):
         self.container.exec(["update-ca-certificates", "--fresh"]).wait()
         subprocess.run(["update-ca-certificates", "--fresh"])
 
+    def _check_disk_space(self):
+        try:
+            database_storage = self.model.storages.get('database', [])
```
sed-i, line 590: We want this runtime dep to loudly fail (we have unit tests):

```suggestion
            database_storage = self.model.storages["database"]
```

src/charm.py
```diff
@@ -584,6 +585,16 @@ def _update_cert(self):
...
+            if database_storage and shutil.disk_usage(database_storage[0].location).free < 1e9: # type: ignore
```
sed-i, line 591:
```suggestion
            if shutil.disk_usage(database_storage[0].location).free < 1024**3: # type: ignore
```

src/charm.py
```diff
@@ -584,6 +585,16 @@ def _update_cert(self):
...
+                self._stored.status["disk_space"] = to_tuple(BlockedStatus("Less than 1 Gi remaining in storage"))
```
sed-i, line 592:
```suggestion
                self._stored.status["disk_space"] = to_tuple(BlockedStatus("<1 GiB disk space remaining"))
```

src/charm.py
```diff
@@ -584,6 +585,16 @@ def _update_cert(self):
...
+
+        # If this check is done before storage is attached, we don't want the charm to go error state
+        except FileNotFoundError:
+            self.unit.status = BlockedStatus("Storage not available")
```
sed-i, line 596: Should we use the same error namespace as before? And should 

```suggestion
            self._stored.status["disk_space"] = MaintenanceStatus("Storage not available")
```

src/charm.py
sed-i: We need an `else` here for clearing the status to `active`.

tests/unit/test_charm.py
```diff
@@ -249,6 +249,22 @@ def test_configuration_reload_read_timeout(self, trigger_configuration_reload, *
         self.harness.evaluate_status()
         self.assertIsInstance(self.harness.model.unit.status, MaintenanceStatus)
 
+    def test_check_disk_space(self):
+        test_cases = [
+            (1e10, ActiveStatus),
+            (1e8, BlockedStatus),
```
sed-i, line 255: - Let's use `1024**3` and `1024**3 + 1`.
- Consider using `unittest.subTest` in addition to the for loop below.

src/charm.py
```diff
@@ -584,6 +585,18 @@ def _update_cert(self):
         self.container.exec(["update-ca-certificates", "--fresh"]).wait()
         subprocess.run(["update-ca-certificates", "--fresh"])
 
+    def _check_disk_space(self):
+        try:
+            #database_storage = self.model.storages.get('database', []) #self.model.storages["database"]
```
lucabello: Forgot to remove the comment :)

src/charm.py
```diff
@@ -584,6 +585,18 @@ def _update_cert(self):
...
+            database_storage = self.model.storages["database"]
+            if database_storage and shutil.disk_usage(database_storage[0].location).free < 1024**3: # type: ignore
```
lucabello:
```suggestion
            
            if database_storage:
                # NOTE: we measure the disk space from the charm container because it shares the same storage as the workload container
                free_disk_space = shutil.disk_usage(database_storage[0].location).free
                
             if free_disk_space < 1024**3:
```

`type: ignore` is a leftover from a previous approach!

lucabello (approved):

src/charm.py
```diff
@@ -584,6 +591,22 @@ def _update_cert(self):
...
+                # NOTE: we measure the disk space from the charm container because it shares the same storage as the workload container
+                free_disk_space = shutil.disk_usage(database_storage[0].location).free
+                if free_disk_space < 1024**3:
+                    self._stored.status["disk_space"] = to_tuple(
+                        BlockedStatus("<1 GiB disk space remaining")
+                    )
+                else:
+                    self._stored.status["disk_space"] = to_tuple(ActiveStatus())
+        # If this check is done before storage is attached, we don't want the charm to go error state
+        except FileNotFoundError:
```
sed-i, line 607: We intent to catch FileNotFoundError from `shutil.disk_usage`, nothing else, so we could tighten this a bit:

```python
if database_storage := self.model.storages["database"]:
    try:
        # NOTE: we measure the disk space from the charm container because it shares the same storage as the workload container
        free_disk_space = shutil.disk_usage(database_storage[0].location).free
    
    # If this check is done before storage is attached, we don't want the charm to go error state
    except FileNotFoundError:
        self._stored.status["disk_space"] = MaintenanceStatus("Storage not available")
    
    else:
        if free_disk_space < 1024**3:
            self._stored.status["disk_space"] = to_tuple(
                BlockedStatus("<1 GiB disk space remaining")
            )
        else:
            self._stored.status["disk_space"] = to_tuple(ActiveStatus())

else:
    # When would self.model.storages["database"] evaluate to false?
    # Do we need this outer if-block at all?
    self._stored.status["disk_space"] = to_tuple(ActiveStatus())
```

sinapah: On the question regarding whether we need the out if-block:
- No, we don't need it. If there is no `database` storage (because it was renamed in `charmcraft.yaml`), a `KeyError` will be thrown. How about:
```
try:
        database_storage = self.model.storages["database"]
        # NOTE: we measure the disk space from the charm container because it shares the same storage as the workload container
        free_disk_space = shutil.disk_usage(database_storage[0].location).free

        # If this check is done before storage is attached, we don't want the charm to go error state
        # except FileNotFoundError:
        # self._stored.status["disk_space"] = MaintenanceStatus("Storage not available")
        except KeyError:
            self._stored.status["disk_space"] = to_tuple(BlockedStatus("Unable to locate storage"))
        else:
            if free_disk_space < 1024**3:
                self._stored.status["disk_space"] = to_tuple(
                    BlockedStatus("<1 GiB disk space remaining")
                )
            else:
                self._stored.status["disk_space"] = to_tuple(ActiveStatus())
```
Technically, we don't need to catch the KeyError. In the event that that during dev the storage is renamed, (1) the update-status hook will run into an error and (2) the unit tests testing this logic will fail. Nonetheless, including the except KeyError may improve readability and understanding here? I'm open to removing it too.

PR #748, opened 2025-11-13
feat(charm lib)!: duplicate host metrics missing alert for collectors thru RemoteWriteConsumer

sinapah:
## Issue

We need to ensure that when we have a remote writer (which relies on `PrometheusRemoteWriterConsumer`), its HostMetricsMissing alert (whose expression is `absent(up{})` is duplicated per each unit of it. This is so that we get an alert even if one of multiple units of that remote writer is down.

## Solution

Addresses https://github.com/canonical/cos-lib/issues/165.

The remote write consumer leader will now rely on peer relation data to duplicate the HostMetricsMissing rule for each of its units. This rule will have severity warning unless the remote writer is a subordinate charm, in which case it will be critical.

Notes:
1. We want the severity of HostMetricsMissing to be `warning` UNLESS the remote write is a subordinate machine charm. In that case, it should be `critical`.

## For K8s
Grafana Agent K8s currently has a bug where it goes into error state if scaled up to 1. Since testing this PR relies on a scaled up aggregator, let's use Otelcol K8s. Pack the charm from the branch in this PR so there is a peer relation: https://github.com/canonical/opentelemetry-collector-k8s-operator/pull/145.

The steps below outline what needs to be done to test this with Otelcol K8s.
1. in `src/integrations.py`, change line 230 to 
```python
remote_write = PrometheusRemoteWriteConsumer(
        charm,
        alert_rules_path=charm_root.joinpath(METRICS_RULES_DEST_PATH).as_posix(),
        extra_alert_labels=key_value_pair_string_to_dict(
            cast(str, charm.model.config.get("extra_alert_labels", ""))
        ),
        peer_relation_name="peers"
    )
```
3. Bump cosl in the dependencies using `uv lock --upgrade-package cosl`. We need version >= 1.4.0.
4. Pack the charm and deploy it with scale=2.
5. Relate to Prometheus at the `receive-remote-write` endpoint.
6. Go to the alerts page in Prometheus. Since your `otel` has scale 2, you should see the HostMetricsMissing alert twice, one per each unit of `otel`. See the pictures below.
image

image

## For machines
Perform the same steps as above. You can relate to the Ubuntu principal. You should see similar results. Ensure that the severity is critical.
image

image

## Upgrade Notes

This change introduces `peer_relation_name`, which will be a mandatory keyword parameter when instantiating a PrometheusRemoteWrite object. Having the cos-tool binary is also necessary for any remote writer, as it's needed for injecting alert rule expressions.

### DIFF
--- a/lib/charms/prometheus_k8s/v1/prometheus_remote_write.py
+++ b/lib/charms/prometheus_k8s/v1/prometheus_remote_write.py
@@ -510,7 +510,7 @@ class PrometheusRemoteWriteConsumer(Object):
 
         if self._forward_alert_rules:
 
-            agg_rules = self._duplicate_rules_per_unit(generic_alert_groups.aggregator_rules, unit_names, duplicate_keys=[HOST_METRICS_MISSING_RULE_NAME], is_subordinate=self._charm.meta.subordinate)
+            agg_rules = self._duplicate_rules_per_unit(generic_alert_groups.aggregator_rules, unit_names, rule_names_to_duplicate=[HOST_METRICS_MISSING_RULE_NAME], is_subordinate=self._charm.meta.subordinate)
             alert_rules.add(agg_rules, group_name_prefix=self.topology.identifier)
 
             alert_rules.add_path(self._alert_rules_path)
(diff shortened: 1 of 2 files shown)


lucabello:
Minor comment: please either remove the `[DRAFT]` from the title (which we shouldn't use in general, having the PR in draft state is enough) or put the PR pack in draft!

lib/charms/prometheus_k8s/v1/prometheus_remote_write.py
```diff
@@ -499,13 +501,21 @@ def _push_alerts_to_relation_databag(self, relation: Relation) -> None:
         if not self._charm.unit.is_leader():
             return
 
+        peer_relations = self._charm.model.get_relation(self._peer_relation_name)
+        peer_relation_units = peer_relations.units if peer_relations else []
```
sed-i: Would it be cleaner if we prepare the unit names here?


```suggestion
        unit_names = ({unit.name for unit in peer_relations.units} if peer_relations else set()) + {self._charm.unit.name}
```

sed-i (commented):
Looking good.
- Per conventional commits, PR title should have a `!`. Also, under "Upgrade notes" section we need to outline the new mandatory arg and the requirement to have cos-tool binary.
- Needing cos-tool should probably be mentioned in the Consumer class docstring?

lib/charms/prometheus_k8s/v1/prometheus_remote_write.py
```diff
@@ -570,6 +579,51 @@ def endpoints(self) -> List[Dict[str, str]]:
...
+            duplicated per unit. The list is to be assigned to the `groups` attribute of an object of type AlertRules.
+        """
+        updated_alert_rules = copy.deepcopy(alert_rules)
+
+        for group in updated_alert_rules.groups:
+            new_rules = []
+            for rule in group["rules"]:
+
+                # If there is a HostMetricsMissing alert there, we need to duplicate it per unit of this app.
+                if rule.get("alert") == "HostMetricsMissing":
```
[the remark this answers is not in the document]
sinapah: Update: we'll add a filter to the method which accepts a list of `include_only` alerts.

lib/charms/prometheus_k8s/v1/prometheus_remote_write.py
```diff
@@ -570,6 +578,49 @@ def endpoints(self) -> List[Dict[str, str]]:
...
 
+    def _duplicate_rules_per_unit(
+        self, alert_rules: Dict[str, Any], peer_unit_names: Set[str], duplicate_keys: List[str], is_subordinate: bool = False
+    ) -> Dict[str, Any]:
+        """Duplicate alert rule per unit in peer_units list.
+
+        Args:
+            alert_rules: A dictionary where key = "groups" and value is a list of rules.
+            peer_unit_names: A set of unit names (str) representing units of this charm.
+            duplicate_keys: A list of alert rule names to be duplicated.
```
MichaelThamm:
```suggestion
            rules_names_to_duplicate: A list of alert rule names to be duplicated.
```

PR #750, opened 2025-11-17
feat: receive-ca-cert relation endpoint

MichaelThamm:
# TODO
- [x] Backport this into track 2

## Issue

Need a way to convey to Prometheus can trust the certs of the servers it's talking to. This is required in COS, explained in this issue:
- https://github.com/canonical/observability-stack/issues/74

And was resolved for Grafana, and needs to be solved for Prometheus, but it requires a `receive-ca-cert` relation endpoint.

## Solution

The CA cert Prometheus gets from the CA which signed it is located at:
`/usr/local/share/ca-certificates/ca.crt`

while the received CA certs are located at:
`/usr/local/share/ca-certificates/juju_receive-ca-cert`

```shell
jrel ca prom:receive-ca-cert
jrel ca prom:certificates

jssh --container prometheus prom/0 ls /usr/local/share/ca-certificates/juju_receive-ca-cert
> 0.crt
jssh --container prometheus prom/0 ls /usr/local/share/ca-certificates/  
> ca.crt
```

Technically these certs can be combined into 1 dir simplifying the logic. Thoughts @sed-i ?

## Context

> [!NOTE]
> Instead of going with the reconciler approach, like we do in otelcol, I decided to implement it with event handlers.

The reason being, that `update-ca-certificates` would be called on every hook and is expensive. We _could_ add conditional logic with hashing to check if `update-ca-certificates` is necessary, but the current code implementation can easily be migrated to a reconciler pattern because both:
- `self._cert_transfer.on.certificate_set_updated`
- `self._cert_transfer.on.certificates_removed`

execute `self._update_ca_certs()`

References:
- https://github.com/canonical/opentelemetry-collector-k8s-operator/pull/20
- https://github.com/canonical/charm-relation-interfaces/tree/main/interfaces/certificate_transfer/v1

## Testing Instructions

> [!TIP]
> 1. The unit test `tox -e unit -- tests/unit/test_receive_ca_cert.py` tests that CA certs from 2 different CAs arrive on disk. And when one relation is removed, the certs for it are also removed.
> 2. The integreation test `tox -e integration -- tests/integration/test_receive_ca_cert.py` tests that Prometheus can scrape a TLS metrics endpoint with a `receive-ca-cert` relation.

### Manual testing
1. Testing deployment
```shell
jd self-signed-certificates --channel 1/stable external-ca --trust
jd self-signed-certificates --channel 1/stable ca --trust
charmcraft pack
jd ./prometheus-k8s_ubuntu-20.04-amd64.charm --resource prometheus-image=IMAGE_NAME prom --trust 
```
2. Relate CAs and check for certs
```shell
jrel ca prom:receive-ca-cert
jssh --container prometheus prom/0 ls /usr/local/share/ca-certificates/juju_receive-ca-cert
> 0.crt
jrel external-ca prom:receive-ca-cert
jssh --container prometheus prom/0 ls /usr/local/share/ca-certificates/juju_receive-ca-cert
> 0.crt  1.crt
```
3. Remove the CA relations and check certs are removed
```shell
jrmrel external-ca prom
jssh --container prometheus prom/0 ls /usr/local/share/ca-certificates/juju_receive-ca-cert
> 0.crt
```
4. The `certificates` relation also provides us a CA cert
```
jrel external-ca prom:certificates
jssh --container prometheus prom/0 ls /usr/local/share/ca-certificates/  
> ca.crt	juju_receive-ca-cert/0.crt
```

### Manual integration tests
1. Deploy Alertmanager and relate it to `ca:certificates` AND `prom:metrics-endpoint`
2. See that scraping alertmanager fails
	- image
3. Relate Prometheus to `ca:receive-ca-cert`
	- _Alternatively_, Prometheus can also server via TLS i.e. `certificates` relation and if the CA that signed Alertmanager is the same that signed Prometheus, then we do not need a `recieve-ca-cert` relation.
5. Scraping succeeds

### DIFF
--- a/.wokeignore
+++ b/.wokeignore
@@ -1,0 +1,1 @@
+lib/charms/certificate_transfer_interface
--- a/src/charm.py
+++ b/src/charm.py
@@ -1,6 +1,6 @@
 #!/usr/bin/env python3
 
-# Copyright 2020 Canonical Ltd.
+# Copyright 2025 Canonical Ltd.
 # See LICENSE file for licensing details.
 
 """A Juju charm for Prometheus on Kubernetes."""
@@ -549,7 +549,6 @@ class PrometheusCharm(CharmBase):
 
     def _on_receive_ca_certs(self, _):
         self._update_ca_certs()
-        self._configure(_)
 
     # TLS CONFIG
     @property
@@ -618,10 +617,8 @@ class PrometheusCharm(CharmBase):
         for i, cert in enumerate(ca_certs):
             self.container.push(RECV_CA_CERT_FOLDER_PATH + f"/{i}.crt", cert, make_dirs=True)
 
-        # TODO: Hash the contents of the certs dir, ALL CERTS, to reduce restarts
         # Refresh system certs
         self.container.exec(["update-ca-certificates", "--fresh"]).wait()
-        # TODO: add this to an itest
         self.container.restart("prometheus")
 
 
--- a/terraform/outputs.tf
+++ b/terraform/outputs.tf
@@ -11,6 +11,7 @@ output "endpoints" {
     ingress          = "ingress"
     catalogue        = "catalogue"
     certificates     = "certificates"
+    receive_ca_cert  = "receive-ca-cert"
     charm_tracing    = "charm-tracing"
     workload_tracing = "workload-tracing"
 
--- a/tests/unit/conftest.py
+++ b/tests/unit/conftest.py
@@ -1,4 +1,4 @@
-# Copyright 2023 Canonical Ltd.
+# Copyright 2025 Canonical Ltd.
 # See LICENSE file for licensing details.
 
 from unittest.mock import patch


MichaelThamm:
WRT Tracing in Prometheus, I assume there are no changes required with this `receive-ca-cert` feature. I think this because [charm_tracing_config](https://github.com/canonical/prometheus-k8s-operator/blob/867c5ae349dafc93b42dc3cd69c51b38cac4dbf0/src/charm.py#L268) accepts a CA cert for tracing via TLS. We pass the CA cert received from the `certificates` relation (i.e. the CA that signed Prometheus) to the tracing function. If for some reason tracing was using Prometheus' ingress address, then in Prometheus we would need to determine which to pass based on ingress. Is the tracing endpoint via TLS always the in-model k8s DNS name, or is it possible that we ever go through ingress?

ping @canonical/tracing-and-profiling

tests/unit/test_receive_ca_cert.py
```diff
@@ -0,0 +1,66 @@
...
+        Relation(
+            "receive-ca-cert", remote_app_data={"certificates": json.dumps([cert1a, cert1b])}
+        ),
+        Relation(
+            "receive-ca-cert", remote_app_data={"certificates": json.dumps([cert2a, cert2b])}
+        ),
+    ]
+    state = State(leader=True, containers={prometheus_container}, relations=relations)
+
+    # WHEN any event is emitted
```
sed-i: We're not using reconciler, so the WHEN should say "recv-ca-changed".

tests/unit/test_receive_ca_cert.py
```diff
@@ -0,0 +1,66 @@
...
+    certs_dir = fs.joinpath(RECV_CA_CERT_FOLDER_PATH.lstrip("/"))
+    assert certs_dir.exists()
+    certs = {file.read_text() for file in certs_dir.glob("*.crt")}
+    assert certs == {cert1a, cert1b, cert2a, cert2b}
+
+    # AND WHEN one relation is removed
+    relations.pop(0)
+    state = State(leader=True, containers={prometheus_container}, relations=relations)
+    # WHEN any event is emitted
+    out = context.run(context.on.update_status(), out)
```
sed-i: This should also be rel changed I assume?

charmcraft.yaml
```diff
@@ -138,6 +138,14 @@ requires:
...
     description: |
       Enables sending workload traces (internal Prometheus traces) to a distributed tracing backend such as Tempo.
+  receive-ca-cert:
+    interface: certificate_transfer
+    optional: true
+    description: |
+      For Prometheus to create a trusted (TLS) connection with servers
+      it scrapes, it needs to trust the CA that signed the server it
+      talks to. CA certs forwarded over this relation are installed in
+      the root CA store.
```
sed-i:
```suggestion
      the root CA store of the prometheus container.
```

PR #782, opened 2026-03-30
fix: add `juju_unit` label to non-wildcard scrape targets when unit can be identified

Abuelodelanada:
## Issue

Charms that register their self-scrape job using fully-qualified hostnames or IP addresses (non-wildcard targets) were missing the `juju_unit` label in the final Prometheus scrape job. This affected `alertmanager-k8s` and any other charm that builds its scrape targets from peer FQDNs rather than using the `*:port` wildcard notation.

Reported in:
- canonical/alertmanager-k8s-operator#395
  - canonical/alertmanager-k8s-operator#272
- canonical/loki-k8s-operator/issues/570
- canonical/grafana-k8s-operator/issues/514

In tandem with:

- https://github.com/canonical/alertmanager-k8s-operator/pull/397
- https://github.com/canonical/loki-k8s-operator/pull/579
- https://github.com/canonical/grafana-k8s-operator/pull/523
- canonical/observability-stack/pull/245

## Root cause

`MetricsEndpointConsumer` has two code paths in `expand_wildcard_targets_into_individual_jobs()`:

- **Wildcard targets** (`*:9093`): the library expands each target into a per-unit job and explicitly adds `juju_unit` because the unit–address mapping is known at expansion time.
- **Non-wildcard targets** (`10.0.0.5:9093` or a FQDN): the library had no way to map the target back to a Juju unit, so it used `label_matcher_dict` which intentionally excludes `unit`. All non-wildcard targets ended up in a single job without `juju_unit`.

## Solution

The information needed to resolve the mapping was already available: each provider unit publishes its address in unit-level relation data via `prometheus_scrape_unit_address`. The fix has three parts:

### 1. Publish the unit FQDN alongside the pod IP (`MetricsEndpointProvider`)

`_set_unit_ip()` now also writes `prometheus_scrape_unit_fqdn` to the unit relation data bag. When the bind address is a valid IP (the common K8s case), the FQDN is obtained from `socket.getfqdn()`. When the address is already a hostname, `fqdn` is set to the same value to avoid a redundant lookup.

image

### 2. Read the FQDN in `_relation_hosts()` (`MetricsEndpointConsumer`)

`_relation_hosts()` now returns a 3-tuple `(address, path, fqdn)` instead of `(address, path)`. The `fqdn` field is empty string for provider units that have not yet published it (old library version), which preserves full backwards compatibility.

### 3. Reverse-lookup in `expand_wildcard_targets_into_individual_jobs()`

Before processing jobs, a reverse-lookup dict is built:

```python
host_to_unit = {
    identifier: unit_name
    for unit_name, (addr, _, fqdn) in hosts.items()
    for identifier in {addr, fqdn} - {""}
}
```

Both the IP and the FQDN of each known unit are mapped to the unit name.
Non-wildcard targets are then classified:

- **Matched** (target host found in `host_to_unit`): expanded into a per-unit job with `juju_unit` and `topology_relabel_config_wildcard`, identical to how wildcard targets are handled.
- **Unmatched** (external service, not a known Juju unit): kept in a single job without `juju_unit`, preserving the original behaviour.

Additionally, the matched-target expansion logic was factored out into a new private helper `_build_per_unit_job()` to eliminate the duplication between the wildcard and non-wildcard matched paths.

image
image

## Testing Instructions

Four new test cases added to `TestWildcardExpansionWithTopology` in
`tests/unit/test_prometheus_config.py`:

- `test_non_wildcard_target_matching_unit_ip_gets_juju_unit_label` — target IP
  matches a known unit's pod IP
- `test_non_wildcard_target_matching_unit_fqdn_gets_juju_unit_label` — target FQDN
  matches a known unit's FQDN (the alertmanager use case)
- `test_non_wildcard_target_not_matching_any_unit_has_no_juju_unit_label` — external
  target, no match → original behaviour preserved
- `test_mixed_non_wildcard_targets_matched_and_unmatched` — one matched + one
  unmatched target in the same job → each ends up in the correct path

### Backwards compatibility

| Scenario | Result |
|---|---|
| Old provider + new consumer | ✅ `fqdn` is empty string → only IP matching active → no regression |
| New provider + old consumer | ✅ Extra relation data key ignored → identical behaviour |
| Both updated | ✅ Non-wildcard targets now get `juju_unit` where resolvable |

### DIFF
--- a/lib/charms/prometheus_k8s/v0/prometheus_scrape.py
+++ b/lib/charms/prometheus_k8s/v0/prometheus_scrape.py
@@ -1227,9 +1227,9 @@ class MetricsEndpointConsumer(Object):
 
         Returns:
             A dict mapping each unit name to a ``(address, path, fqdn)`` tuple. The
-            ``fqdn`` element may be an empty string when the FQDN is not known; when
-            present, it may either be distinct from or equal to ``address`` (for
-            example, when the unit address itself is already a hostname).
+            ``fqdn`` element may be an empty string when the FQDN is not known. When
+            present, it may either be distinct from, or equal to ``address``. For
+            example, when the unit address itself is already a hostname.
         """
         hosts = {}
         for unit in relation.units:
--- a/lib/charms/prometheus_k8s/v0/prometheus_scrape.py
+++ b/lib/charms/prometheus_k8s/v0/prometheus_scrape.py
@@ -555,7 +555,8 @@ class PrometheusConfig:
                 continue
 
             # Accumulates non-wildcard targets that could not be matched to any known unit.
-            # These are kept in a single job without juju_unit, preserving original behaviour.
+            # These are kept in a single job with topology-only labels (no juju_unit):
+            # fully-qualified targets that predate this feature are unaffected.
             unmatched_static_configs = []
 
             for static_config in static_configs:
@@ -582,7 +583,7 @@ class PrometheusConfig:
 
                 # Non-wildcard targets: try to match each target's host against known unit
                 # addresses. Matched targets get a per-unit job with juju_unit; unmatched
-                # targets fall back to the previous behaviour (no juju_unit label).
+                # targets get topology-only labels with no per-unit expansion.
                 if non_wildcard_targets:
                     unmatched_targets = []
                     matched_by_unit: Dict[str, List[str]] = {}
@@ -598,7 +599,8 @@ class PrometheusConfig:
                         else:
                             unmatched_targets.append(target)
 
-                    # Unmatched targets: no unit mapping found — preserve original behaviour.
+                    # Unmatched targets: no unit mapping found — kept with topology-only
+                    # labels and no per-unit expansion (juju_unit is not added).
                     if unmatched_targets:
                         unmatched_static_config = static_config.copy()
                         unmatched_static_config["targets"] = unmatched_targets
--- a/lib/charms/prometheus_k8s/v0/prometheus_scrape.py
+++ b/lib/charms/prometheus_k8s/v0/prometheus_scrape.py
@@ -513,15 +513,17 @@ class PrometheusConfig:
     ) -> List[dict]:
         """Extract wildcard hosts from the given scrape_configs list into separate jobs.
 
-        For wildcard targets (e.g. "*:9093"), one job per unit is created with the
-        corresponding ``juju_unit`` topology label.
+        For wildcard targets (e.g. "*:9093"), one job per unit is created. When
+        ``topology`` is provided, the ``juju_unit`` label is injected into each
+        per-unit job; without ``topology`` the per-unit jobs are created but no
+        topology labels are added.
 
         For non-wildcard targets (fully qualified hostnames/IPs), the host portion of
         each target is matched against the known unit addresses in ``hosts``. Targets
-        whose address matches a known unit are expanded into a per-unit job with the
-        ``juju_unit`` label, mirroring the wildcard behaviour. Targets with no match
-        (e.g. external services) are kept in a single job without ``juju_unit``,
-        preserving the previous behaviour.
+        whose address matches a known unit are expanded into a per-unit job (with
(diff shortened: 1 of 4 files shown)


lib/charms/prometheus_k8s/v0/prometheus_scrape.py
```diff
@@ -461,32 +461,97 @@ def prefix_job_names(scrape_configs: List[dict], prefix: str) -> List[dict]:
...
+            ]
+        return new_job
+
     @staticmethod
     def expand_wildcard_targets_into_individual_jobs(
         scrape_jobs: List[dict],
-        hosts: Dict[str, Tuple[str, str]],
+        hosts: Dict[str, Tuple[str, str, str]],
         topology: Optional[JujuTopology] = None,
     ) -> List[dict]:
```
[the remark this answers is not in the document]
Abuelodelanada: All callers in this repo already use 3-tuples — always produced by _relation_hosts().

sed-i (commented):
Would you mind adding tandem PRs to cos-proxy and otelcol (the charm that has the tandem itest)?

Abuelodelanada:
> Would you mind adding tandem PRs to cos-proxy and otelcol (the charm that has the tandem itest)?

Look @sed-i for `cos-proxy` seems it is not necessary.

With this deployment:

image

I see this target in Prometheus:

image


And the config in prometheus:

```yaml
  job_name: juju_test_231c328_telegraf_prometheus_scrape
  relabel_configs:
  - regex: (.*)
    separator: _
    source_labels:
    - juju_model
    - juju_model_uuid
    - juju_application
    - juju_unit
    target_label: instance
  static_configs:
  - labels:
      dns_name: juju-030d6d-6.lxd
      host: 10.72.158.38
      juju_application: telegraf
      juju_model: test
      juju_model_uuid: 231c328b-caff-4dee-8a15-e280df030d6d
      juju_unit: telegraf/4
    targets:
    - 10.72.158.38:9103

```

lib/charms/prometheus_k8s/v0/prometheus_scrape.py
```diff
@@ -510,58 +580,59 @@ def expand_wildcard_targets_into_individual_jobs(
...
-                        # there is no reliable way to determine the name (Juju topology unit name)
-                        # for such a target. Therefore labeling with Juju topology, excluding the
-                        # unit name.
-                        non_wildcard_static_config["labels"] = {
-                            **topology.label_matcher_dict,
-                            **non_wildcard_static_config.get("labels", {}),
-                        }
-
-                    non_wildcard_static_configs.append(non_wildcard_static_config)
+                    unmatched_targets = []
```
sed-i: The volume of the diff here is quite large and difficult to reason about.
Would it be possible to make the diff more grokable? E.g. maintain previous order of processing? Extract free functions?

lib/charms/prometheus_k8s/v0/prometheus_scrape.py
```diff
@@ -461,32 +461,102 @@ def prefix_job_names(scrape_configs: List[dict], prefix: str) -> List[dict]:
...
         """
-        # hosts = self._relation_hosts(relation)
+        # Reverse lookup: both unit address and FQDN → unit name, so that non-wildcard
+        # targets specified as either IP or FQDN can be matched to their Juju unit.
+        # {addr, fqdn} deduplicates in the case where address is already a FQDN.
+        host_to_unit: Dict[str, str] = (
+            {
+                identifier: unit_name
+                for unit_name, (addr, _, fqdn) in hosts.items()
+                for identifier in {addr, fqdn} - {""}
```
sed-i: Would this work? Does it seem clearer?

```suggestion
                for identifier in {addr, fqdn} if identifier
```

Abuelodelanada: Yes, I'm trying to convert that way.

Comment polished.

I wouldn't use the `if identifier` because an `" "` or "0" are truthy

sed-i (approved):
Thanks Jose!

PR #785, opened 2026-04-14
feat!: bump to `prometheus` workload version 3.11.1

sinapah:
This pull request bumps the workload version used in the charm from 2.55.3 to 3.11.1. The most significant changes as a result of this change is the addition of a fallback scrape protocol for Prometheus metrics scraping. A `charmcraft fetch-lib` has also been done.

**Prometheus Metrics Scraping Enhancements:**

* Added a `fallback_scrape_protocol` parameter to the `MetricsEndpointConsumer` class in `prometheus_scrape.py`. This allows specifying a protocol (e.g., `"PrometheusText0.0.4"`) to use when the Content-Type header is missing or invalid, improving compatibility with older Prometheus versions and non-standard scrape targets.

**Container Image and Library Updates:**

* Updated the Prometheus container image in `charmcraft.yaml` to a newer version (`3.11-24.04`), ensuring up-to-date features and security fixes.
* Bumped patch versions for the Grafana dashboard, Prometheus scrape, and Traefik ingress libraries to reflect new features and fixes. 

## Upgrade Notes

We're going from workload version 2.55.3 to 3.11.1. Staring in COS and COS Lite track 3, we will use the charm with the 3.11.x rock. If a user is on Prom with COS track 3 and they wish to downgrade to track 2, they **must** downgrade to a Prom rev which uses Prom 2.55, such as rev 289. Otherwise, they will experience data loss.


charmcraft.yaml
```diff
@@ -55,7 +55,7 @@ resources:
   prometheus-image:
     type: oci-image
     description: Container image for Prometheus
-    upstream-source: ubuntu/prometheus@sha256:e14c7d6a46d4c7a8c8e127a5334beeabbe1080b6e4d6f439a59ae564d5bec341  # renovate: oci-image tag: 2.55-24.04
+    upstream-source: ubuntu/prometheus@sha256:3b037e9157370026b1e67e5b3edfb17fb08f22b8db2a22400baeb250963dd78f  # renovate: oci-image tag: 3.11-24.04
```
sinapah, line 58: Specifically, this is 3.11.1.

src/charm.py
```diff
@@ -834,8 +836,6 @@ def _generate_command(self) -> str:
             f"--config.file={PROMETHEUS_CONFIG}",
             "--storage.tsdb.path=/var/lib/prometheus",
             "--web.enable-lifecycle",
-            "--web.console.templates=/usr/share/prometheus/consoles",
-            "--web.console.libraries=/usr/share/prometheus/console_libraries",
```
sinapah, line 838: Removed because they no longer come with the Prom bin.

sinapah:
The tag `charm libraries: out of sync` is because I bumped the Prom scrape lib. Everything else is up to date.

lib/charms/prometheus_k8s/v0/prometheus_scrape.py
```diff
@@ -950,6 +963,11 @@ def __init__(self, charm: CharmBase, relation_name: str = DEFAULT_RELATION_NAME)
                 It is strongly advised not to change the default, so that people
                 deploying your charm will have a consistent experience with all
                 other charms that consume metrics endpoints.
+            fallback_scrape_protocol: an optional fallback protocol to use when the
+                Content-Type header of a scrape response is missing or invalid. Supported
+                values: "PrometheusProto", "OpenMetricsText0.0.1", "OpenMetricsText1.0.0",
+                "PrometheusText0.0.4", "PrometheusText1.0.0". Ref:
+                https://prometheus.io/docs/prometheus/latest/configuration/configuration/#scrape_config.
```
sed-i, line 970: Let's add here a bit more details for our future selves, e.g.

> Explicit protocol is required starting Prometheus v3. For backwards compat, ...

release-notes/prometheus-2.55.1-to-3.11.1-notes.md
sed-i, line 1: We made two separate bumps, one to 2.55 and then to 3.
Would we gain from splitting the release notes accordingly?

src/charm.py
```diff
@@ -215,7 +215,9 @@ def __init__(self, *args):
         self._topology = JujuTopology.from_charm(self)
 
         self.grafana_dashboard_provider = GrafanaDashboardProvider(charm=self)
-        self.metrics_consumer = MetricsEndpointConsumer(self)
+        # Fallback to PrometheusText0.0.4 for older Prometheus versions that don't support OpenMetrics or the newer Prometheus text format OR
+        # for scrape targets that don't set a Content-Type header.
+        self.metrics_consumer = MetricsEndpointConsumer(self, fallback_scrape_protocol="PrometheusText0.0.4")
```
sed-i, line 223: Can you elaborate:
- why 0.0.4 and not 1.0.0 (e.g. "most legacy instrumentations are assumed to be ...")
- that this works for prom v3+, but not for prom v2

release-notes/prometheus-2.55.1-to-3.11.1-notes.md
sinapah: Sure. I've split the one MD file into:
- release-notes/prometheus-2.53.3-to-2.55.1-notes.md
- release-notes/prometheus-2.55.1-to-3.11.1-notes.md

lib/charms/prometheus_k8s/v0/prometheus_scrape.py
sinapah: Updated to:
```py
fallback_scrape_protocol: an optional fallback protocol to use when the
                Content-Type header of a scrape response is missing or invalid. Supported
                values: "PrometheusProto", "OpenMetricsText0.0.1", "OpenMetricsText1.0.0",
                "PrometheusText0.0.4", "PrometheusText1.0.0". Ref:
                https://prometheus.io/docs/prometheus/latest/configuration/configuration/#scrape_config.
                This had to be added after we bumped to Prometheus workload major version 3. Starting in major 3,
                Prometheus no longer defaults to the Prometheus text format (PrometheusText0.0.4)
                when the Content-Type header is missing or invalid, and instead fails the scrape with an error.
                This parameter should only be used by MetricsEndpointConsumers that use Prometheus 3 and above, as setting
                this key in the scrape configs of Prometheus 2 will result in the error:
                "field fallback_scrape_protocol not found in type config.ScrapeConfig".
```

sed-i (approved):

Issue #662, opened 2024-12-20
Charm behaviour differs from grafana-agent for log_level config option


MichaelThamm:
### Enhancement Proposal

The `log_level` config option was introduced into grafana agent machine charm inspired by the functionality from the Prometheus charm.
- https://github.com/canonical/grafana-agent-operator/pull/217

However, the above PR was implemented in a way that if an incorrect log level config was passed then the charm enters `Blocked` status rather than just silenty accepting the config and only creating a warning log message. The user may not realize their log level was not accepted.

Make this behaviour consistent here like in Grafana Agent.

gycsaba96:
Fixed by PR #786.

PR #786, opened 2026-04-14
fix: use blocked status on invalid log_level (#662)

gycsaba96:
## Issue

See #662 

Before this commit, an invalid log_level config value just created a log message. After this commit, it also moves into a blocked state.

## Solution

- move to a blocked state when an invalid log_level is provided instead of just providing a warning
- add the same test scenario as for maximum_retention_size
- fix a small typo

(A similar solution is used for maximum_retention_size.)

### DIFF
--- a/tests/unit/test_charm.py
+++ b/tests/unit/test_charm.py
@@ -68,7 +68,7 @@ class TestCharm(unittest.TestCase):
             expected_logs = {
                 "WARNING:root:Invalid loglevel: bad-level given, "
                 "debug/info/warn/error/fatal allowed. "
-                "defaulting to DEBUG loglevel."
+                "Defaulting to DEBUG loglevel."
             }
             self.assertGreaterEqual(set(logger.output), expected_logs)  # type: ignore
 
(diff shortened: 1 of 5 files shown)


gycsaba96:
It is worth noting that prometheus-k8s defaults to the `debug` level in case of an invalid `log_level` config while grafana-agent uses `info` in the same scenario as mentioned [here](https://github.com/canonical/grafana-agent-operator/pull/217). I left it as `debug` since there was already a unit test in place.

tests/unit/test_charm.py
```diff
@@ -370,6 +370,52 @@ def test_invalid_retention_size_config_option_string(self, *unused):
...
+        self.assertIsInstance(self.harness.model.unit.status, status_class)
+
+    @k8s_resource_multipatch
+    @patch("lightkube.core.client.GenericSyncClient")
+    def test_invalid_log_level_config_option_string(self, *unused):
+        # GIVEN a running charm with default values
+        self.mock_capacity.return_value = "1Gi"
+        self.harness.container_pebble_ready("prometheus")
+        self.harness.handle_exec("prometheus", ["update-ca-certificates"], result=0)
+        self.harness.begin_with_initial_hooks()
```
MichaelThamm:
```suggestion
```

tests/unit/test_charm.py
```diff
@@ -370,6 +370,52 @@ def test_invalid_retention_size_config_option_string(self, *unused):
...
+class TestConfigLogLevel(unittest.TestCase):
+    """Test the charmcraft.yaml option 'log_level'."""
+
+    def setUp(self):
+        self.harness = Harness(PrometheusCharm)
+        self.addCleanup(self.harness.cleanup)
+
+        patcher = patch.object(PrometheusCharm, "_get_pvc_capacity")
+        self.mock_capacity = patcher.start()
+        self.addCleanup(patcher.stop)
```
MichaelThamm, line 386: We should keep the setup calls in the `setUp` method of the TestCase:
```suggestion
        self.addCleanup(patcher.stop)
        self.mock_capacity.return_value = "1Gi"
        self.harness.container_pebble_ready("prometheus")
        self.harness.handle_exec("prometheus", ["update-ca-certificates"], result=0)
        self.harness.begin_with_initial_hooks()
```

tests/unit/test_charm.py
```diff
@@ -370,6 +370,52 @@ def test_invalid_retention_size_config_option_string(self, *unused):
...
+    def _check_state(self,expected_log_level,status_class):
+        plan = self.harness.get_container_pebble_plan("prometheus")
+        self.assertEqual(cli_arg(plan, "--log.level"),expected_log_level)
+        self.harness.evaluate_status()
+        self.assertIsInstance(self.harness.model.unit.status, status_class)
+
+    @k8s_resource_multipatch
+    @patch("lightkube.core.client.GenericSyncClient")
+    def test_invalid_log_level_config_option_string(self, *unused):
+        # GIVEN a running charm with default values
```
MichaelThamm, line 403: You can move this `GIVEN` to the `setUp` method if you like

tests/unit/test_charm.py
```diff
@@ -370,6 +370,52 @@ def test_invalid_retention_size_config_option_string(self, *unused):
...
+        self.harness.container_pebble_ready("prometheus")
+        self.harness.handle_exec("prometheus", ["update-ca-certificates"], result=0)
+        self.harness.begin_with_initial_hooks()
+        self.harness.evaluate_status()
+        self.assertIsInstance(self.harness.model.unit.status, ActiveStatus)
+
+        # WHEN the config option is set to an invalid string
+        self.harness.update_config({"log_level": "bad-level"})
+
+        # THEN cli arg is unspecified and the unit is blocked
```
MichaelThamm: What do you mean by:
> CLI arg is unspecified

your assertion does not reflect that comment.

tests/unit/test_charm.py
```diff
@@ -370,6 +370,52 @@ def test_invalid_retention_size_config_option_string(self, *unused):
...
+        self.harness.update_config({"log_level": "very-bad-level"})
+
+        # THEN cli arg is unspecified and the unit is blocked
+        self._check_state("debug",BlockedStatus)
+
+        # AND WHEN the config option is corrected
+        self.harness.update_config({"log_level": "error"})
+
+        # THEN cli arg is updated and the unit is goes back to active
+        self._check_state("error",ActiveStatus)
```
MichaelThamm: Here is a format suggestion:
```suggestion
        self._check_state("error", ActiveStatus)
```

pro tip: run `tox -e fmt` and it should format your tests and src code.

MichaelThamm (dismissed):
Overall, this PR looks good! I think you picked a bad charm to contribute to because it is very outdated i.e., Harness tests vs. the newer Scenario tests. Thank you for contributing to core <3

PS: This repo has some flaky itests, just re-run them a few times to get them to pass.

sinapah (commented):
Would you also be able to do `charmcraft fetch-lib` 😊?

gycsaba96:
Thank you @sinapah and @MichaelThamm for the review! I reworked the unit test and updated the charm libraries.

sinapah (approved):
LGTM :)

MichaelThamm (approved):

PR #799, opened 2026-04-29
feat: migrate charm-tracing to ops[tracing]

lucabello:
## Summary

Migrate charm tracing from the old `charm_tracing` charm library (`@trace_charm` decorator) to the new `ops[tracing]` approach using `ops_tracing.Tracing`. Workload tracing continues to use `TracingEndpointRequirer`.

### Changes
- Replaced `@trace_charm` decorator with `ops_tracing.Tracing` instantiation
- Updated dependencies to use `ops[tracing]`
- Added `cargo` to build-packages
- Removed old `charm_tracing.py` library (kept `tracing.py` for workload-tracing)
- Updated test fixtures

### Motivation
The `ops[tracing]` approach is the recommended way to add charm tracing as of ops 2.20.0.


src/charm.py
```diff
@@ -263,16 +252,14 @@ def __init__(self, *args):
         )
 
         self.catalogue = CatalogueConsumer(charm=self, item=self._catalogue_item)
-        self.charm_tracing = TracingEndpointRequirer(
-            self, relation_name="charm-tracing", protocols=["otlp_http"]
+        self.tracing = ops_tracing.Tracing(
```
mmkay: Again, I think it's worth mentioning that this is charm tracing, not to cause any misunderstandings.

tests/unit/test_prometheus_rules_provider.py
```diff
@@ -19,6 +19,7 @@ def _make_consumer_charm(alert_rules_path: str):
     class ConsumerCharm(CharmBase):
         metadata_yaml = textwrap.dedent(
             """
+            name: consumer-charm
```
mmkay, line 22: Rules change?

mmkay (approved):

src/charm.py
lucabello: Done!

tests/unit/test_prometheus_rules_provider.py
lucabello: I think so; not sure why else, cause this test fails with `KeyError` otherwise :(

Abuelodelanada (approved):
LGTM

PR #800, opened 2026-05-05
feat: drop alert labels when label value is empty

sed-i:
## Issue
In tiered otelcol setups where different tiers owned by different teams, it could be useful to have some alert labels dropped.

## Solution
Slightly modify the semantics such that empty label values would result with the label being dropped.

Addresses:
- https://github.com/canonical/opentelemetry-collector-k8s-operator/issues/157
- https://github.com/canonical/opentelemetry-collector-operator/issues/150

In tandem with:
- https://github.com/canonical/loki-k8s-operator/pull/606


lucabello:
@sinapah @sed-i is this still relevant?

PR #814, opened 2026-06-04
feat: bump to 26.04

lucabello:


tests/integration/test_resource_limits.py
```diff
@@ -4,117 +4,132 @@
...
-    podspec = get_podspec(ops_test, app_name, "prometheus")
-    assert podspec.resources
-    assert equals_canonically(podspec.resources.limits, default_limits)
-    assert equals_canonically(podspec.resources.requests, default_requests)
-    assert await check_prometheus_is_ready(ops_test, app_name, 0)
+def test_default_resource_limits_applied_after_resetting_config(juju: jubilant.Juju):
+    juju.config(APP_NAME, reset=["cpu", "memory"])
+    juju.wait(lambda status: jubilant.all_active(status, APP_NAME) and jubilant.all_agents_idle(status, APP_NAME), timeout=600, delay=5.0)
+
+    assert juju.model
```
MichaelThamm, line 130: Why do we need this? We are using Jubilant, not pytest-operator

tests/integration/test_upgrade_charm.py
```diff
@@ -17,6 +17,12 @@
     run_promql,
 )
 
+# Cross-base upgrades (e.g. 24.04 -> 26.04) are not supported via juju refresh.
+# The charmhub charm is built for 24.04 (Python 3.12), while the local charm
+# targets 26.04 (Python 3.14). Juju refresh only replaces charm code, not the
+# container image, so the old container's Python cannot load the new venv.
+pytestmark = pytest.mark.skip(reason="Cross-base upgrade from 24.04 to 26.04 not supported")
```
MichaelThamm: Am I missing something, or is this test just simulating an `upgrade` hook by deploying the packed charm and then refreshing to that same packed charm. There is no upgrading from `24.04` to `26.04` happening here. Why xfail?

MichaelThamm (changes requested):

tests/integration/test_resource_limits.py
lucabello: Because it's similarly needed for the static check to pass

tests/integration/test_upgrade_charm.py
lucabello: This seems like a bug in the tests, because what the `upgrade` tests we usually have do is deploy from Charmhub and then refresh; I was an autopilot, will remove the skip and try again

PR #815, opened 2026-06-09
fix: refresh metrics-endpoint unit address on `update_status` and `relation_changed`

### LINKED ISSUE #270: Opentelemetry-collector-k8s doesn't keep track of the IP address of containers with the metrics-endpoint
### Bug Description

When performing a maintenance on a k8s host, such as rebooting the node or stopping the k8s service for a moment, the containers internal IP (usually in 10.1.0.0/16 range) will potentially change.

While most application will use the service IP address, this is not the case with Otelcol's metrics-endpoint which will use directly the pod IP address to collect metrics from various pods.

This creates an issue where after such maintenance or outage, pods may use a different IP address, breaking Otelcol from scraping metrics for Prometheus/Mimir.
An alert is visible on AlertManager to notify about the absence of metrics for the problematic units.

It is expected to have the pod IP address to be automatically updated whenever a container uses a new IP address from k8s.

### To Reproduce

1. Deploy 2 k8s clusters
2. Deploy COS lite or HA on first k8s cluster
3. Deploy Mysql application with 3 units on second cluster
4. Deploy Otelcol-k8s in the same model as Mysql and relate it with Mysql and COS.
5. Do a maintenance on one k8s node hosting MySQL (shutdown, reboot, ...)

6. Mysql container IP may have changed, COS is unable to find metrics and AlertManager will raise an alert

### Environment

Ubuntu 24.04
COS 2/stable , deployed through Terraform
Openstack Sunbeam 2024.1
opentelemetry-collector-k8s 2/stable rev 174

### Relevant log output

```shell
# kubectl get -n openstack pods | grep mysql
NAME      READY   STATUS    RESTARTS      AGE    IP           NODE             NOMINATED NODE   READINESS GATES
mysql-0   2/2     Running   2 (23d ago)   24d    10.1.2.203   k8s-ctrl-2       <none>           <none>
mysql-1   2/2     Running   0             23d    10.1.1.206   k8s-ctrl-3       <none>           <none>
mysql-2   2/2     Running   0             23d    10.1.0.200   k8s-ctrl-1       <none>           <none>


# Current content from metrics-endpoint relation between otelcol-k8s and mysql
    related-units:
      mysql/0:
        in-scope: true
        data:
          egress-subnets: 10.152.183.115/32
          ingress-address: 10.152.183.115
          private-address: 10.152.183.115
          prometheus_scrape_unit_address: 10.1.2.179  <---- this is the previous IP of the pod
          prometheus_scrape_unit_name: mysql/0
      mysql/1:
        in-scope: true
        data:
          egress-subnets: 10.152.183.115/32
          ingress-address: 10.152.183.115
          private-address: 10.152.183.115
          prometheus_scrape_unit_address: 10.1.1.206
          prometheus_scrape_unit_name: mysql/1
      mysql/2:
        in-scope: true
        data:
          egress-subnets: 10.152.183.115/32
          ingress-address: 10.152.183.115
          private-address: 10.152.183.115
          prometheus_scrape_unit_address: 10.1.0.200
          prometheus_scrape_unit_name: mysql/2
```

### Additional context

Removing and then re-adding the relation between otelcol-k8s and mysql will refresh the content and then fix the capability of scrapping the containers.
But this is really not a sustainable workaround for large scale deployments such as Openstack where there are many applications related...

```
    related-units:
      mysql/0:
        in-scope: true
        data:
          egress-subnets: 10.152.183.115/32
          ingress-address: 10.152.183.115
          private-address: 10.152.183.115
          prometheus_scrape_unit_address: 10.1.2.203
          prometheus_scrape_unit_name: mysql/0
      mysql/1:
        in-scope: true
        data:
          egress-subnets: 10.152.183.115/32
          ingress-address: 10.152.183.115
          private-address: 10.152.183.115
          prometheus_scrape_unit_address: 10.1.1.206
          prometheus_scrape_unit_name: mysql/1
      mysql/2:
        in-scope: true
        data:
          egress-subnets: 10.152.183.115/32
          ingress-address: 10.152.183.115
          private-address: 10.152.183.115
          prometheus_scrape_unit_address: 10.1.0.200
          prometheus_scrape_unit_name: mysql/2
```

Abuelodelanada:
## Issue

Fixes https://github.com/canonical/opentelemetry-collector-k8s-operator/issues/270

On Kubernetes, a workload pod can be rescheduled (node reboot/maintenance, eviction, etc.) and come back with a **new pod IP**. The `MetricsEndpointProvider` only republished its address (`prometheus_scrape_unit_address`) on `relation_joined` and the configured `refresh_event` — which for a sidecar/pebble charm defaults to `pebble_ready`. None of those events reliably re-fire on a bare IP change, so the **stale IP lingered in relation data**. Any consumer (Prometheus/Mimir directly, or `opentelemetry-collector-k8s` in between) kept scraping a dead IP. The only workaround was to remove and re-add the relation, which does not scale for large deployments (e.g. OpenStack Sunbeam with many related applications).

## Solution

Make address publishing reconcile-friendly so a changed pod IP self-heals without manual intervention. Two observers are added in `MetricsEndpointProvider`, both calling the existing `_set_unit_ip()`:

1. **`update_status`** — observed unconditionally or any user-provided `refresh_event`. Since `update_status` fires periodically, the published address self-heals within one hook interval after a reschedule.
2. **`relation_changed`** — `_set_unit_ip()` now runs at the start of `_on_relation_changed`, giving a faster reaction than waiting for the next `update_status`. It is a no-op when the address is unchanged.

`LIBPATCH` is bumped `58 → 59`. No public API changes; the fix is purely additive
(extra event observers) and writes the same unit-level relation data key as before.

## Context

- The bug is on the **provider** side of `prometheus_scrape`. The consumer (`MetricsEndpointConsumer`) already reads unit relation data live on every reconcile, so it faithfully reflects whatever address the provider last published — it never caches a stale value itself.
- `_set_unit_ip()` writes to **unit** relation data (`relation.data[self._charm.unit]`), so it is correct and safe for every unit to call it (no leadership gating required); each unit publishes its own address.
- Why the old behaviour missed the IP change: `_set_unit_ip()` only ran via `relation_joined` and the `refresh_event`(s). When a charm does not pass an explicit `refresh_event`, the library picks a default for it — and for a typical single-container K8s charm that default is just `[<container>_pebble_ready]`. A pod rescheduled onto a new IP does not necessarily re-emit `pebble_ready`/`relation_joined`, so nothing rewrote the address. 
- This mirrors the intent already documented in `_set_unit_ip`'s docstring ("Each time a metrics provider charm container is restarted it updates its own host address …") — this PR makes that promise hold for IP changes that are not accompanied by a container restart event.

## Testing Instructions

Unit tests (two new regression cases in
`tests/unit/test_endpoint_provider.py::TestEndpointProvider`):

```bash
tox -e unit -- tests/unit/test_endpoint_provider.py tests/unit/test_endpoint_consumer.py
```

New tests:
- `test_provider_unit_refreshes_address_on_update_status` — publishes an initial pod IP, then simulates a reschedule (new `ops.Network.bind_address`) and emits only `update_status`; asserts `prometheus_scrape_unit_address` is updated to the new IP.
- `test_provider_unit_refreshes_address_on_relation_changed` — same setup, but the refresh is driven by a `relation_changed` (via a remote databag update).

To confirm the tests catch the regression, revert only the library and keep the tests:

```bash
git checkout main -- lib/charms/prometheus_k8s/v0/prometheus_scrape.py

tox -e unit -- tests/unit/test_endpoint_provider.py -k refreshes_address_on
# Expected: both new tests FAIL (address stays at the old IP)

git checkout fix/ip-address -- lib/charms/prometheus_k8s/v0/prometheus_scrape.py
```

## Upgrade Notes

No action required. Once a provider charm re-fetches `prometheus_scrape` (>= LIBPATCH 59)
and is refreshed, its unit address is refreshed automatically on `update_status` and
`relation_changed`; any already-stale address self-heals within one update-status interval.
The fix is backwards compatible:

| Scenario | Result |
|---|---|
| New provider (>=59) + old consumer | OK — same relation data key, just refreshed more often |
| Old provider (<=58) + any consumer | OK — unchanged behaviour; the stale-IP bug persists until the provider is updated |

Note on scope: the fix lives on the **provider** side, so the benefit lands once the
metrics-providing charms adopt the updated library. `prometheus-k8s` is the source of truth
for this library; after `charmcraft publish-lib`, downstream charms re-fetching the library
pick up the fix.

### DIFF
--- a/lib/charms/prometheus_k8s/v0/prometheus_scrape.py
+++ b/lib/charms/prometheus_k8s/v0/prometheus_scrape.py
@@ -361,7 +361,7 @@ LIBAPI = 0
 
 # Increment this PATCH version before using `charmcraft publish-lib` or reset
 # to 0 if you are raising the major API version
-LIBPATCH = 58
+LIBPATCH = 59
 
 # Version 0.0.53 needed for cosl.rules.generic_alert_groups
 PYDEPS = ["cosl>=0.0.53"]
@@ -1622,8 +1622,22 @@ class MetricsEndpointProvider(Object):
         for ev in refresh_event:
             self.framework.observe(ev, self.set_scrape_job_spec)
 
+        # Always re-evaluate the unit address on `update_status`, regardless of the charm type
+        # (sidecar/pebble or podspec) or any user-provided `refresh_event`. On Kubernetes a pod
+        # can be rescheduled (e.g. node reboot/maintenance) and come back with a new IP without
+        # re-emitting `relation_joined`/`pebble_ready`. Without this, the stale address lingers in
+        # relation data and the consumer keeps scraping a dead IP until the relation is recreated.
+        # `update_status` fires periodically, so the address self-heals within one hook interval.
+        # See https://github.com/canonical/opentelemetry-collector-k8s-operator/issues/270
+        self.framework.observe(self._charm.on.update_status, self._set_unit_ip)
+
     def _on_relation_changed(self, event):
         """Check for alert rule messages in the relation data before moving on."""
+        # Refresh the unit address on every `relation_changed`. This reacts faster than waiting for
+        # the next `update_status` when the pod IP changes, and is a no-op when the address is
+        # unchanged. See https://github.com/canonical/opentelemetry-collector-k8s-operator/issues/270
+        self._set_unit_ip()
+
         if self._charm.unit.is_leader():
             ev = json.loads(event.relation.data[event.app].get("event", "{}"))
 
--- a/lib/charms/prometheus_k8s/v0/prometheus_scrape.py
+++ b/lib/charms/prometheus_k8s/v0/prometheus_scrape.py
@@ -1634,8 +1634,7 @@ class MetricsEndpointProvider(Object):
     def _on_relation_changed(self, event):
         """Check for alert rule messages in the relation data before moving on."""
         # Refresh the unit address on every `relation_changed`. This reacts faster than waiting for
-        # the next `update_status` when the pod IP changes, and is a no-op when the address is
-        # unchanged. See https://github.com/canonical/opentelemetry-collector-k8s-operator/issues/270
+        # the next `update_status` when the pod IP changes, and is safe to call repeatedly.
         self._set_unit_ip()
 
         if self._charm.unit.is_leader():
--- a/tests/unit/test_endpoint_provider.py
+++ b/tests/unit/test_endpoint_provider.py
@@ -238,6 +238,55 @@ class TestEndpointProvider(unittest.TestCase):
         self.assertIn("prometheus_scrape_unit_name", data)
         self.assertEqual(data.get("prometheus_scrape_unit_path", ""), "")
 
+    def test_provider_unit_refreshes_address_on_update_status(self):
+        """A rescheduled pod (new IP) must refresh its address on update_status.
+
+        Regression test for the scenario where a Kubernetes node maintenance/reboot
+        reschedules the workload pod onto a new IP without re-emitting
+        relation_joined/pebble_ready. Without refreshing on update_status, the stale
+        address would linger in relation data and the consumer would keep scraping a
+        dead IP. See canonical/opentelemetry-collector-k8s-operator#270.
+        """
+        rel_id = self.harness.add_relation(RELATION_NAME, "provider")
+        self.harness.add_relation_unit(rel_id, "provider/0")
+
+        # GIVEN a provider that already published its (old) pod IP
+        with patch("ops.Network.bind_address", new="10.1.157.116"):
+            self.harness.charm.provider.set_scrape_job_spec()
+            data = self.harness.get_relation_data(rel_id, self.harness.charm.unit.name)
+            self.assertEqual(data["prometheus_scrape_unit_address"], "10.1.157.116")
+
+            # WHEN the pod is rescheduled onto a new IP and only update_status fires
+            with patch("ops.Network.bind_address", new="10.1.200.42"):
+                self.harness.charm.on.update_status.emit()
+
+        # THEN the published address is refreshed to the new pod IP
+        data = self.harness.get_relation_data(rel_id, self.harness.charm.unit.name)
+        self.assertEqual(data["prometheus_scrape_unit_address"], "10.1.200.42")
+
+    def test_provider_unit_refreshes_address_on_relation_changed(self):
+        """A new pod IP must be reflected on relation_changed as a faster reaction.
+
+        Regression test for canonical/opentelemetry-collector-k8s-operator#270.
+        """
+        rel_id = self.harness.add_relation(RELATION_NAME, "provider")
+        self.harness.add_relation_unit(rel_id, "provider/0")
+
+        # GIVEN a provider that already published its (old) pod IP
+        with patch("ops.Network.bind_address", new="10.1.157.116"):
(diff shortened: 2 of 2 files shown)


lib/charms/prometheus_k8s/v0/prometheus_scrape.py
```diff
@@ -1622,8 +1622,22 @@ def __init__(
         for ev in refresh_event:
             self.framework.observe(ev, self.set_scrape_job_spec)
 
+        # Always re-evaluate the unit address on `update_status`, regardless of the charm type
+        # (sidecar/pebble or podspec) or any user-provided `refresh_event`. On Kubernetes a pod
+        # can be rescheduled (e.g. node reboot/maintenance) and come back with a new IP without
```
sed-i, line 1627: Thought (non-blocking): Afaik, "pod reschedule" results in the upgrade sequence. In the past @ca-scribner discovered that `start` is missing iirc, but I imagine "upgrade-charm" should reliably fire. Another option is to call `_set_unit_ip` unconditionally (e.g. in Provider's `__init__`, if that works.

tests/unit/test_endpoint_provider.py
MichaelThamm, line 1: `test_provider_unit_refreshes_address_on_update_status` and `test_provider_unit_refreshes_address_on_relation_changed` are essentially the same test: consider moving them into one with pytest.paramatrize?

Abuelodelanada: I've considered that, but since we are not using pytest here I did not want to pollute the PR with that addition

MichaelThamm:
I assume if a user sets their update-status interval to 1 hour or greater, then this solution is not ideal. However, I think this is at least an improvement.

MichaelThamm (approved):

lib/charms/prometheus_k8s/v0/prometheus_scrape.py
Abuelodelanada: Hey @sed-i thanks!

AFAIK an `upgrade-charm` fires reliably when the rescheduling goes through the charm upgrade sequence. but a bare pod reschedule (node reboot/drain/eviction) is a K8s event, not a Juju charm upgrade, so the lifecycle events Juju re-emits there aren't guaranteed.

Regarding running `_set_unit_ip` unconditionally in the `__init__`: The problem I see is that the `bind_address` may be unavailable. If that is the case we will end up with a fqdn.

I would keep these changes as is, and eventually open another PR.

nobuto-m:
> I assume if a user sets their update-status interval to 1 hour or greater, then this solution is not ideal. However, I think this is at least an improvement.

In the classic charm world, the `start` hook was where charm authors can do anything related to the unit service initialization just after the restart of the unit. My understanding is the issue happens when pod is restarted/recreated on host reboot so I thought the `start` was a suitable place as a guess.
https://documentation.ubuntu.com/juju/3.6/reference/hook/#start
However, I'm not so familiar with K8s charm hooks so I might be wrong.

nobuto-m:
Thinking out loud, but in case `config-changed` hook is not called in such an event as pod restart/recreation, it could be a wider issue as Juju and its ecosystem. Otherwise, each charm has to implemented this kind of workaround. 

https://documentation.ubuntu.com/juju/3.6/reference/hook/#config-changed
> when any IP address of the machine hosting the unit changes.

PR #817, opened 2026-06-10
feat: set blocked status if invalid rules in relation data and filter out invalid rules

### LINKED ISSUE #816: The charm should go into blocked state if there are invalid rules coming on relation data
### Enhancement Proposal

If there are any invalid rules in relation data, the Prometheus Scrape and Prometheus Remote Write libs should write the validation errors to the `.events.errors` field, similar to Loki Push API. Subsequently, the Prometheus charm should set status to blocked.

sinapah:
## Issue

Fixes #816.
Previously, Prometheus wouldn't set its status to blocked if there were invalid rules on disk. 
## Solution

This PR introduces several changes. These include:
1. The Prometheus Remote Write library now correctly filters out invalid rules from the alerts dictionary it returns to the caller which is the charm itself. The Prometheus Scrape library already correctly filters out invalid rules.
1. Note re item above: the Prometheus Remote Write library already correctly filters out invalid rules in _most_ cases. The reason a change in logic was required was to cover the very unlikely corner case where an alert rule does not have labels from which the Juju Topology can be constructed AND it does not have `scrape_metadata`. In all other practical cases, the filtering logic is already as desired.
1. Both Prometheus Remote Write and Prometheus Scrape update relation data and write any validation errors to `.event.errors` when iterating over rules coming from relation data.
1. The charm determines whether it needs to set blocked due to invalid rules by checking relation data. This is related to the item directly above. If there are invalid rules in relation data, the charm sets blocked with an appropriate message in order to get the attention of an admin.
1. The Prometheus Remote Write library now observes `relation_broken` as a hook that should emit a custom `consumers_changed` event. Previously, `relation_departed` would cause a`consumers_changed`, but not `relation_broken`. This is because the evaluation of alert validation happens by looking at relation data. The `consumers_changed` event causes a `_configure` to happen in `charm.py`. However, there is no guarantee that app data has been wiped by the time `relation_departed` has occurred. It is, however, guaranteed that app data is cleared by the time `relation_broken` has fired. So, we should also observe `relation_broken` so that we `_configure`, which means we take another look at app data to determine if there are invalid rules.
1. The Prometheus Scrape library does the same thing as above. `relation_broken` now causes a `targets_changed` emission in this lib, which the charm will then observe, leading it to `_configure()`.

## Context

I think the combination of 5 and 6 under "Solution" will also fix(?) #780. To be confirmed.
Re 5 and 6 above, see [this](https://matrix.to/#/!iO1K5yQ6KvVHgl2XnLHrzN2efmA0MrKZ-ZyijgbzWHc/$PyTpqk0KypGuGdHQx6i-JNZjwiSWJDeueFh_HX0zQTI?via=ubuntu.com&via=matrix.org&via=matrix.debian.social).

## Testing Instructions

Deploy the following bundle:
```yaml
bundle: kubernetes
applications:
  aval:
    charm: avalanche-k8s
    channel: 2/edge
    revision: 59
    resources:
      avalanche-image: 20
    scale: 1
    constraints: arch=amd64
    trust: true
  cos-config:
    charm: cos-configuration-k8s
    channel: dev/edge
    revision: 102
    resources:
      git-sync-image: 36
    scale: 1
    options:
      git_branch: main
      git_repo: https://github.com/sinapah/cos-config
      git_ssh_key_secret: secret://d8k306fmp25c76c0n6ug/key
      prometheus_alert_rules_path: prometheus_alert_rules
    constraints: arch=amd64
    storage:
      content-from-git: kubernetes,1,1024M
    trust: true
  prom:
    charm: prometheus-k8s
    channel: dev/edge
    revision: 296
    resources:
      prometheus-image: 154
    scale: 1
    constraints: arch=amd64
    storage:
      database: kubernetes,1,1024M
    trust: true
relations:
- - aval:metrics-endpoint
  - prom:metrics-endpoint
```

Initially, this bundle deploys the latest revision of Prom which does not contain this change. Prom is related to Avalanche, which contains only valid rules. 
1. Run `juju ssh --container prometheus prom/0 ls /etc/prometheus/rules`.
You should see a result similar to the following because Avalanche has valid rules:
```
juju_test_1fa6431e_aval_metrics-endpoint_27.rules
```
3. Now, relate Prometheus to COS Config:
`juju relate cos-config prom:receive-remote-write`
 
COS Config will send invalid rules to Prometheus over relation data. The content of `juju ssh --container prometheus prom/0 ls /etc/prometheus/rules` shouldn't have changed because `prometheus_remote_write` will correctly filter out the invalid rules.
However, even though there are invalid rules in relation data, the charm's status must still be `active`.

4. Pack the charm from this branch
```bash
charmcraft pack
```
5. Refresh Prometheus
```bash
juju refresh prom --path=./prometheus-k8s_ubuntu@24.04-amd64.charm
```
6. Now, Prometheus should go into blocked.
```
prom/0*        blocked   idle   10.1.0.91          Invalid alert rules. See debug-log
```
7. If you remove the relation between Prometheus and COS Config:
```bash
juju remove-relation cos-config prom
```
8. Then Prometheus should become `active` again.


lucabello:
I have two points to raise about this PR.

1. The reason why currently Prometheus doesn't go into blocked state on some invalid alert rules is because we didn't want a remote charm being able to essentially "block" Prometheus, making it look like that's the problem from a Juju status perspective. I'm not sure changing this is a good idea, but if we don't have other options, I'm okay with it.
2. A few of our libraries have this RPC-style mechanisms like the "errors" field, with the idea of "Prometheus will pilot the status of other charms via relation data by telling them 'hey, there is an error'". While this mechanism is present, it's currently not used, and I would rather keep it that way. We've seen over the years that this is not a good mechanism to handle status, so I would rather go towards a direction of us eventually removing "errors" entirely, rather than start using it. I believe we shouldn't use relation data RPC-style mechanisms. Can we remove changes to that without impacting the issue you want to fix?

lucabello (changes requested):
Let a comment on the approach, especially on the `errors` field, which I think we shouldn't utilize

sinapah:
Hi @lucabello,

A couple of comments regarding your points:
- > The reason why currently Prometheus doesn't go into blocked state on some invalid alert rules is because we didn't want a remote charm being able to essentially "block" Prometheus, making it look like that's the problem from a Juju status perspective.

I don't disagree here. I initially stumbled upon a related issue to this in Loki K8s, where the intention was that we'd set blocked in the case of invalid rules. I wanted to reverse that behaviour (and had even put up https://github.com/canonical/loki-k8s-operator/pull/622) before realizing that setting `blocked` on invalid rules/dashboards was actually a core epic. It seems like this is the behaviour that users expect
- I came to actually prefer this approach, as even in the past week, on the support channels I received one or two messages asking why rules coming from COS Config were not showing up on the Grafana alerts page. It looks like without a visual cue that something WRT alert rules and validation is amiss, the first instinct of an admin is that something related to reading rules from relation data or syncing them through RGW is the problem.
- Yes, ideally, the charm providing the alert rules should set blocked if it has invalid rules bundled. In the case of COS Config, it is _ideally_ the charm that should set blocked. However, performing that validation on the side of the charm that provides the alerts requires all of them to have `cos-tool` as a part in the charm. For something like COS Config that can be done since it's ours. For all the other charms that have bundled rules, it's probably not doable.
- The `errors` field is actually used in Loki, as you can see [here](https://github.com/canonical/loki-k8s-operator/pull/622). It has been in the code for a few years now. It's actually a pretty clean and straightforward way for the provider side of the library to communicate the validation errors to both the charm at the other end and the charm on "this" side of the relation. From there it's pretty straightforward for the charm to read the validation errors and surface it in status.

lucabello (approved):

PR #834, opened 2026-06-25
fix(prometheus_remote_write): sort peer names for alerts duplication

michaeldmitry:
Contributes to fixing https://github.com/canonical/opentelemetry-collector-operator/issues/330

## Issue

When opentelemetry-collector (or any charm using `PrometheusRemoteWriteConsumer`) is scaled to multiple units, the send-remote-write relation databag churns continuously, causing an infinite cascade of `relation-changed` events on the remote-write receiver (e.g. Mimir).

`_duplicate_rules_per_unit` iterates over peers (a `set`). On every charm invocation the HostMetricsMissing rules are appended in a different order for each peer unit. 

## Solution
Sort `peer_unit_names` before iterating


sed-i (approved):
Thank you!

PR #846, opened 2026-07-10
feat: source `cosTool` from cosl

sinapah:
## Issue

Similar to https://github.com/canonical/loki-k8s-operator/pull/607/.
Before https://github.com/canonical/cos-lib/pull/202 goes in, we should source CosTool in these two libs from `cosl` to pull all the performance efficiency changes easily.

## Solution

Replaces the local CosTool clone so it can be sourced from `cosl`.

## Testing Instructions

Once merged, if we decide we want the perf improvements of https://github.com/canonical/cos-lib/pull/202 in tracks 2 and 3, then we should backport the changes of this PR using `charmcraft fetch-lib` in those two tracks.

### DIFF
--- a/lib/charms/prometheus_k8s/v0/prometheus_scrape.py
+++ b/lib/charms/prometheus_k8s/v0/prometheus_scrape.py
@@ -335,12 +335,13 @@ import subprocess
 import tempfile
 from collections import defaultdict
 from pathlib import Path
-from typing import Any, Callable, Dict, List, Literal, Optional, Tuple, Union
+from typing import Any, Callable, Dict, List, Literal, Optional, Tuple, Union, cast
 from urllib.parse import urlparse
 
 import yaml
-from cosl import JujuTopology
+from cosl import CosTool, JujuTopology
 from cosl.rules import AlertRules, generic_alert_groups
+from cosl.types import OfficialRuleFileFormat
 from ops.charm import CharmBase, RelationRole
 from ops.framework import (
     BoundEvent,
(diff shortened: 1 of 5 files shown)


lib/charms/prometheus_k8s/v0/prometheus_scrape.py
```diff
@@ -1356,6 +1357,43 @@ def _target_parts(self, target) -> list:
         return parts
 
 
+def _validate_scrape_jobs(jobs: list) -> bool:
```
sinapah, line 1362: `validate_scrape_jobs` was in this local fork, but it never made it to the `cosTool` from `cosl`. I added it as a free-standing func here, but should it go to `cosl`? AFAICT, the scrape lib is the only lib that relies on it, so perhaps not.

lib/charms/prometheus_k8s/v0/prometheus_scrape.py
```diff
@@ -1143,7 +1144,7 @@ def alerts(self) -> dict:
 
             alerts[identifier] = alert_rules
 
-            _, errmsg = self._tool.validate_alert_rules(alert_rules)
+            _, errmsg = self._tool.validate_alert_rules(cast(OfficialRuleFileFormat, alert_rules))
```
MichaelThamm: This casting is annoying, and type checker avoidance. Could we do something like this:
```python
self._tool.validate_alert_rules(OfficialRuleFileFormat(groups=alert_rules["groups"]))
```

or have [_inject_alert_expr_labels](https://github.com/canonical/prometheus-k8s-operator/blob/0ad0e0c498f74506e1a331dcda1c00e416d52e94/lib/charms/prometheus_k8s/v0/prometheus_scrape.py#L1212) return this type so we don't need to cast.

lib/charms/prometheus_k8s/v0/prometheus_scrape.py
MichaelThamm: We can leave it here for now

MichaelThamm (approved):

MichaelThamm (approved):

PR #849, opened 2026-08-05
feat: blocked on invalid scrape jobs

### LINKED ISSUE #633: Error in scrape job definition pass silently
### Bug Description

Prometheus is unable to unmarshal `params` sent through Prometheus scrape target.
Prometheus remains in `Active` state

### To Reproduce

1. Deploy Prometheus: `juju deploy prometheus-k8s prom --channel edge --trust`
2. Deploy Prometheus scrape target: `juju deploy prometheus-scrape-target-k8s scrape --channel edge`
3. Config Prometheus scrape target:
   - `juju config prometheus-scrape-target-k8s targets=192.168.0.248:9116`
   - `juju config prometheus-scrape-target-k8s labels="job:cumulus"`
   - `juju config prometheus-scrape-target-k8s metrics_path="/snmp"`
   - `juju config scrape params='{"auth": "snmp_v3", "module": "if_mib_if_name", "target": "192.168.100.200"}'`
4. Relate Prometheus to Prometheus scrape target: `juju relate prom scrape`
5. Verify this scrape job is not included in Prometheus:
   ```shell
   $ juju ssh --container prometheus prom/0 cat /etc/prometheus/prometheus.yml                                                            
   global:
     evaluation_interval: 1m
     scrape_interval: 1m
     scrape_timeout: 10s
   rule_files:
   - /etc/prometheus/rules/juju_*.rules
   scrape_configs:
   - honor_timestamps: true
     job_name: prometheus
     metrics_path: /metrics
     relabel_configs:
     - regex: (.*)
       separator: _
       source_labels:
       - juju_model
       - juju_model_uuid
       - juju_application
       - juju_unit
       target_label: instance
     scheme: http
     scrape_interval: 5s
     scrape_timeout: 5s
     static_configs:
     - labels:
         host: localhost
         juju_application: prom
         juju_charm: prometheus-k8s
         juju_model: dmytro
         juju_model_uuid: 67755b6d-9410-46d9-8617-ee7c87d285c2
         juju_unit: prom/0
       targets:
       - prom-0.prom-endpoints.dmytro.svc.cluster.local:9090
   ```

Alternatively it is possible to use this bundle:

```yaml
bundle: kubernetes
applications:
  prom:
    charm: prometheus-k8s
    channel: latest/edge
    revision: 210
    resources:
      prometheus-image: 149
    scale: 1
    constraints: arch=amd64
    storage:
      database: kubernetes,1,1024M
    trust: true
  scrape:
    charm: prometheus-scrape-target-k8s
    channel: latest/edge
    revision: 34
    scale: 1
    options:
      labels: job:cumulus
      params: '{"auth": "snmp_v3", "module": "if_mib_if_name", "target": "192.168.100.200"}'
      targets: 192.168.0.248:9116
    constraints: arch=amd64
relations:
- - prom:metrics-endpoint
  - scrape:metrics-endpoint
```

### Environment

```
Model   Controller  Cloud/Region        Version  SLA          Timestamp
dmytro  microk8s    microk8s/localhost  3.5.2    unsupported  16:23:20-03:00

App     Version  Status  Scale  Charm                         Channel      Rev  Address        Exposed  Message
prom    2.52.0   active      1  prometheus-k8s                latest/edge  210  10.152.183.22  no       
scrape  n/a      active      1  prometheus-scrape-target-k8s  latest/edge   34  10.152.183.36  no       

Unit       Workload  Agent  Address     Ports  Message
prom/0*    active    idle   10.1.9.252         
scrape/0*  active    idle   10.1.9.217         

Integration provider     Requirer               Interface          Type     Message
prom:prometheus-peers    prom:prometheus-peers  prometheus_peers   peer     
scrape:metrics-endpoint  prom:metrics-endpoint  prometheus_scrape  regular 
```

### Relevant log output

```shell
unit-prom-0: 16:07:08.634 INFO unit.prom/0.juju-log metrics-endpoint:3: reqs=ResourceRequirements(claims=None, limits={}, requests={'cpu': '0.25', 'memory': '200Mi'}), templated=ResourceRequirements(claims=None, limits=None, requests={'cpu': '250m', 'memory': '200Mi'}), actual=ResourceRequirements(claims=None, limits=None, requests={'cpu': '250m', 'memory': '200Mi'})
unit-prom-0: 16:07:08.672 DEBUG unit.prom/0.juju-log metrics-endpoint:3: No alertmanagers available
unit-prom-0: 16:07:08.704 ERROR unit.prom/0.juju-log metrics-endpoint:3: Validating scrape jobs failed: b'time="2024-07-17T19:07:08Z" level=fatal msg="parsing YAML file /tmp/tmpe9yyw2pz: yaml: unmarshal errors:\\n  line 4: cannot unmarshal !!str `snmp_v3` into []string\\n  line 5: cannot unmarshal !!str `if_mib_...` into []string\\n  line 6: cannot unmarshal !!str `192.168...` into []string"\n'
unit-prom-0: 16:07:08.757 INFO unit.prom/0.juju-log metrics-endpoint:3: Pushed new configuration
```

### Additional context

_No response_

sinapah:
## Issue

Fixes #633.

Based on changes in https://github.com/canonical/prometheus-k8s-operator/pull/817.

## Solution

Similar to https://github.com/canonical/prometheus-k8s-operator/pull/817, we:
1. Write validation errors from Cos Tool to local app data (only the leader can do that). This was already implemented.
2. When `_configure()` runs, we fetch scrape jobs. If there are any scrape jobs with validation errors, we set blocked.
3. If the relation changes and the validation errors go away, we clear local app data.

## Testing Instructions

The charm and the changes implemented in this PR are available for testing on `dev/edge/pr849`.

### Before
To observe the "before" behaviour, first deploy the bundle below:
```yaml
bundle: kubernetes
applications:
  prom:
    charm: prometheus-k8s
    channel: dev/edge
    revision: 312
    resources:
      prometheus-image: 155
    scale: 1
    constraints: arch=amd64
    storage:
      database: kubernetes,1,1024M
    trust: true
  scrape:
    charm: prometheus-scrape-target-k8s
    channel: dev/edge
    revision: 42
    scale: 1
    options:
      labels: dns_name:k8s-cluster-ai-1.maas,kubernetes_node:k8s-cluster-ai-1
      metrics_path: /metrics
      params: ""
      scheme: http
      targets: 10.0.0.0:3000
    constraints: arch=amd64
    trust: true
relations:
- - scrape:metrics-endpoint
  - prom:metrics-endpoint

```
The bundle above results in a valid Prometheus scrape job. You can confirm this by curling the Prometheus targets endpoint:
```
curl 10.1.0.15:9090/api/v1/targets | jq 
...
...
...
"activeTargets": [
      {
        "discoveredLabels": {
          "__address__": "10.0.0.0:3000",
          "__metrics_path__": "/metrics",
          "__scheme__": "http",
          "__scrape_interval__": "1m",
          "__scrape_timeout__": "10s",
          "dns_name": "k8s-cluster-ai-1.maas",
          "job": "juju_test2_6081db3_scrape_external_jobs",
          "kubernetes_node": "k8s-cluster-ai-1"
        },
        "labels": {
          "dns_name": "k8s-cluster-ai-1.maas",
          "instance": "10.0.0.0:3000",
          "job": "juju_test2_6081db3_scrape_external_jobs",
          "kubernetes_node": "k8s-cluster-ai-1"
        },
        "scrapePool": "juju_test2_6081db3_scrape_external_jobs",
        "scrapeUrl": "http://10.0.0.0:3000/metrics",
        "globalUrl": "http://10.0.0.0:3000/metrics",
        "lastError": "",
        "lastScrape": "0001-01-01T00:00:00Z",
        "lastScrapeDuration": 0,
        "health": "unknown",
        "scrapeInterval": "1m",
        "scrapeTimeout": "10s"
      },
...
...
...

```
To simulate what an invalid scrape job, follow the steps below.
1. `juju config scrape params='{"auth": "snmp_v3", "module": "if_mib_if_name", "target": "192.168.100.200"}'. This config option provided is invalid because the `params` config option expects the values in the dictionary to be a list of strings, not strings themselves.
4. This results in a validation error that ends up getting written to the databag by Prometheus:
```
juju show-unit scrape/0
...
...
...
relation-id: 5
    endpoint: metrics-endpoint
    related-endpoint: metrics-endpoint
    application-data:
      event: '{"scrape_job_errors": "Command ''[''/var/lib/juju/agents/unit-prom-0/charm/cos-tool-amd64'',
        ''validate-config'', ''/tmp/tmpd24mqjw1.yaml'']'' returned non-zero exit status
        1."}'
    related-units:
      prom/0:
...
...
...
```
5. Because the validation of the scrape job fails, this job should not be returned when curling Prom's targets endpoint.
6. If you now clear the `params` config option, the scrape job succeeds and shows up in the list of scrape targets.
7. However, the previous validation error written to relation data does not get removed.
```
jshu scrape/0
...
...
...
- relation-id: 5
    endpoint: metrics-endpoint
    related-endpoint: metrics-endpoint
    application-data:
      event: '{"scrape_job_errors": "Command ''[''/var/lib/juju/agents/unit-prom-0/charm/cos-tool-amd64'',
        ''validate-config'', ''/tmp/tmpd24mqjw1.yaml'']'' returned non-zero exit status
        1."}'
    related-units:
      prom/0:
...
...
```
### After
To observe the after behaviour, refresh the Prom charm: `juju refresh prom --channel dev/edge/pr849`.
1. `juju config params='{"auth": "snmp_v3", "module": "if_mib_if_name", "target": "192.168.100.200"}'`. This is invalid and should result in Prometheus being blocked.
2. The status message should say "Invalid scrape jobs. See debug-log"
3. There should be a validation error message in the relation data Scrape sees:
```
application-data:
      event: '{"scrape_job_errors": "Command ''[''/var/lib/juju/agents/unit-prom-0/charm/cos-tool-amd64'',
        ''validate-config'', ''/tmp/tmphpzy36ht.yaml'']'' returned non-zero exit status
        1."}'
    related-units:
      prom/0:

```
4. Update the `params` config option to something valid such as: `juju config scrape params='{"auth": ["snmp_v3"], "module": ["if_mib_if_name", "if_mib"]}'`
5. Prometheus should become active.
8. The validation errors should be cleared:
```
- relation-id: 5
    endpoint: metrics-endpoint
    related-endpoint: metrics-endpoint
    application-data:
      event: '{}'
    related-units:
      prom/0:
        in-scope: true

```
9. The scrape job should be visible from the Prometheus POV.

## Upgrade Notes

Once this PR merges, the `prometheus_scrape` lib will have to be bumped across all charms that use it. When this happens in Opentelemetry Collector K8s and Opentelemetry Collector, we will need to ensure those two charms also set blocked status if there are any validaton errors in local app data.

We need the same functionality of writing validation errors to local app data in COS Agent. The issue is that COS Agent does not validate anything using COS Tool.

### DIFF
--- a/lib/charms/prometheus_k8s/v0/prometheus_scrape.py
+++ b/lib/charms/prometheus_k8s/v0/prometheus_scrape.py
@@ -362,7 +362,7 @@ LIBAPI = 0
 
 # Increment this PATCH version before using `charmcraft publish-lib` or reset
 # to 0 if you are raising the major API version
-LIBPATCH = 62
+LIBPATCH = 63
 
 # Version 0.0.53 needed for cosl.rules.generic_alert_groups
 PYDEPS = ["cosl>=0.0.53"]
@@ -1055,11 +1055,16 @@ class MetricsEndpointConsumer(Object):
                 try:
                     _validate_scrape_jobs(static_scrape_jobs)
                 except subprocess.CalledProcessError as e:
+                    logger.error(f"Invalid scrape job file: {e}")
                     if self._charm.unit.is_leader():
                         data = json.loads(relation.data[self._charm.app].get("event", "{}"))
                         data["scrape_job_errors"] = str(e)
                         relation.data[self._charm.app]["event"] = json.dumps(data)
                 else:
+                    if self._charm.unit.is_leader():
+                        data = json.loads(relation.data[self._charm.app].get("event", "{}"))
+                        data.pop("scrape_job_errors", None)
+                        relation.data[self._charm.app]["event"] = json.dumps(data)
                     scrape_jobs.extend(static_scrape_jobs)
 
         scrape_jobs = _dedupe_job_names(scrape_jobs)
--- a/src/charm.py
+++ b/src/charm.py
@@ -132,6 +132,7 @@ class CompositeStatus(TypedDict):
     k8s_patch: Tuple[str, str]
     config: Tuple[str, str]
     alert_rules: Tuple[str, str]
+    scrape_jobs: Tuple[str, str]
 
 
 def to_tuple(status: StatusBase) -> Tuple[str, str]:
@@ -172,6 +173,7 @@ class PrometheusCharm(CharmBase):
                 k8s_patch=to_tuple(ActiveStatus()),
                 config=to_tuple(ActiveStatus()),
                 alert_rules=to_tuple(ActiveStatus()),
+                scrape_jobs=to_tuple(ActiveStatus()),
             )
         )
 
@@ -857,6 +859,29 @@ class PrometheusCharm(CharmBase):
 
         return False
 
+    def _has_scrape_job_errors(self) -> bool:
+        """Check if any metrics-endpoint relation reported scrape job validation errors."""
+        for relation in self.model.relations.get(DEFAULT_METRICS_RELATION_NAME, []):
+            app_data = relation.data.get(self.app)
+            if not app_data:
+                continue
+
+            event_raw = app_data.get("event", "{}")
+            try:
+                event_data = json.loads(event_raw)
+            except (json.JSONDecodeError, TypeError):
+                continue
+
+            if event_data.get("scrape_job_errors"):
+                logger.error(
+                    "Scrape job validation error on relation %s: %s",
+                    relation.id,
+                    event_data["scrape_job_errors"],
+                )
+                return True
+
+        return False
+
     def _push_alert_rules(self, alerts):
         """Pushes alert rules from a rules file to the prometheus container.
 
@@ -1137,6 +1162,13 @@ class PrometheusCharm(CharmBase):
             processed_job, processed_certs = self._process_tls_config(job)
             certs = {**certs, **processed_certs}
             prometheus_config["scrape_configs"].append(processed_job)  # type: ignore
+
+        if self._has_scrape_job_errors():
+            msg = "Invalid scrape jobs. See debug-log"
+            logger.error(msg)
+            self._stored.status["scrape_jobs"] = to_tuple(BlockedStatus(msg))
+        else:
+            self._stored.status["scrape_jobs"] = to_tuple(ActiveStatus())
 
         web_config = self._web_config()
 
--- a/tests/unit/test_scrape_job_filtering.py
+++ b/tests/unit/test_scrape_job_filtering.py
@@ -4,11 +4,13 @@
 import dataclasses
 import json
 
-import yaml
+import ops
 from ops.model import ActiveStatus, BlockedStatus
 from scenario import Relation, State
 
 from charm import to_status
+
+ops.testing.SIMULATE_CAN_CONNECT = True  # pyright: ignore
 
 
 # To test that invalid scrape jobs are filtered out and the charm goes into blocked
@@ -57,17 +59,6 @@ INVALID_SCRAPE_JOB_RELATION = Relation(
 )
 
 
-def _scrape_job_names(context, state_out):
-    """Read the scrape job names written to the prometheus config on disk."""
-    fs = state_out.get_container("prometheus").get_filesystem(context)
-    prometheus_config = fs.joinpath("etc", "prometheus", "prometheus.yml")
-    if not prometheus_config.exists():
-        return set()
-
-    config = yaml.safe_load(prometheus_config.read_text())
-    return {job.get("job_name") for job in config.get("scrape_configs", [])}
-
-
 def _scrape_jobs_status(state_out):
     """Return only the scrape-jobs status, ignoring unrelated unit statuses."""
     charm_stored_state = next(
@@ -89,8 +80,7 @@ def test_valid_scrape_job_relation_remains_active(context, prometheus_container)
     # WHEN the relation changed event is processed
     state_out = context.run(context.on.relation_changed(VALID_SCRAPE_JOB_RELATION), state_in)
 
-    # THEN the valid job is written and the scrape-jobs status remains active
-    assert any("valid-job" in name for name in _scrape_job_names(context, state_out))
+    # THEN the scrape-jobs status remains active
     assert isinstance(_scrape_jobs_status(state_out), ActiveStatus)
 
 
@@ -105,8 +95,7 @@ def test_invalid_scrape_job_relation_blocks(context, prometheus_container):
     # WHEN the relation changed event is processed
     state_out = context.run(context.on.relation_changed(INVALID_SCRAPE_JOB_RELATION), state_in)
 
-    # THEN the invalid job is filtered out and the scrape-jobs status is blocked
-    assert not any("invalid-job" in name for name in _scrape_job_names(context, state_out))
+    # THEN the scrape-jobs status is blocked
     assert isinstance(_scrape_jobs_status(state_out), BlockedStatus)
 
 
@@ -168,6 +157,6 @@ def test_invalid_scrape_job_relation_becoming_valid_recovers_to_active(
         dataclasses.replace(blocked_state, relations=[now_valid_relation]),
     )
 
-    # THEN the previous invalid status is cleared and valid jobs are written
-    assert any("valid-job" in name for name in _scrape_job_names(context, recovered_state))
+    # THEN the previous invalid status is cleared
     assert isinstance(_scrape_jobs_status(recovered_state), ActiveStatus)
+
--- a/tests/unit/test_scrape_job_filtering.py
+++ b/tests/unit/test_scrape_job_filtering.py
@@ -1,0 +1,173 @@
+# Copyright 2025 Canonical Ltd.
+# See LICENSE file for licensing details.
+
+import dataclasses
+import json
+
+import yaml
+from ops.model import ActiveStatus, BlockedStatus
+from scenario import Relation, State
+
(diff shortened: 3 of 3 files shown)


lib/charms/prometheus_k8s/v0/prometheus_scrape.py
```diff
@@ -1055,11 +1055,16 @@ def jobs(self) -> list:
                 try:
                     _validate_scrape_jobs(static_scrape_jobs)
                 except subprocess.CalledProcessError as e:
+                    logger.error(f"Invalid scrape job file: {e}")
```
sed-i, line 1058: Should be truncate `{e}` in case it's too long?

lib/charms/prometheus_k8s/v0/prometheus_scrape.py
```diff
@@ -1055,11 +1055,16 @@ def jobs(self) -> list:
...
                     if self._charm.unit.is_leader():
                         data = json.loads(relation.data[self._charm.app].get("event", "{}"))
                         data["scrape_job_errors"] = str(e)
                         relation.data[self._charm.app]["event"] = json.dumps(data)
                 else:
+                    if self._charm.unit.is_leader():
```
sed-i, line 1064: Not sure if this is the right place for it, but should we add here a FIXME or TODO about the fact we need a leader guard?

src/charm.py
```diff
@@ -857,6 +859,29 @@ def _has_alert_rule_errors(self) -> bool:
 
         return False
 
+    def _has_scrape_job_errors(self) -> bool:
+        """Check if any metrics-endpoint relation reported scrape job validation errors."""
+        for relation in self.model.relations.get(DEFAULT_METRICS_RELATION_NAME, []):
```
sed-i, line 864: Iiuc, `DEFAULT_METRICS_RELATION_NAME` should not be used this way because it is intended as an internal default arg. We should have probably prefixed it with `_`.
I.e. users can provide an alternative name and then this would break. We need to obtain the actual relation name.
But maybe I'm missing something.

tests/unit/test_scrape_job_filtering.py
```diff
@@ -0,0 +1,173 @@
...
+        leader=True,
+        relations=[VALID_SCRAPE_JOB_RELATION],
+        containers=[prometheus_container],
+    )
+
+    # WHEN the relation changed event is processed
+    state_out = context.run(context.on.relation_changed(VALID_SCRAPE_JOB_RELATION), state_in)
+
+    # THEN the valid job is written and the scrape-jobs status remains active
+    assert any("valid-job" in name for name in _scrape_job_names(context, state_out))
```
sed-i: - Should we use `_has_scrape_job_errors` here instead?
- And separately test `_has_scrape_job_errors`?

sed-i (commented):
Looks good! A couple peripheral questions.

lib/charms/prometheus_k8s/v0/prometheus_scrape.py
sinapah: In most cases yes, but I think in this case it's unnecessary. If and when a subprocess error is returned it looks like:
```
"Command ''[''/var/lib/juju/agents/unit-prom-0/charm/cos-tool-amd64'',
        ''validate-config'', ''/tmp/tmphpzy36ht.yaml'']'' returned non-zero exit status
        1."}'
```

This is because `_validate_scrape_jobs` runs `subprocess` with `check=True`.
https://github.com/canonical/prometheus-k8s-operator/blob/9d57daca214461616c15f455a6870ea75482485c/lib/charms/prometheus_k8s/v0/prometheus_scrape.py#L1387-L1393

Which will always return the output in format "Command "["x", "y", "z"] returned non-zero exit status 1. "

It can never be much longer than that.

lib/charms/prometheus_k8s/v0/prometheus_scrape.py
sinapah: I think the leader guard here is not a concern (right?), but rather it's inside Mimir where we set blocked that I think we should add a comment. A comment that says something like (in Mimir's charm code, not lib code):
"Since non-leader units can't read local app data, when the coordinator is scaled up, only the leader blocks in the case of validation errors".

src/charm.py
sinapah: Good point. IIUC, this is not a concern for this charm because Prometheus instantiates the MetricsEndpointConsumer with the default relation name (always). This would be only a problem if someone else writes another charm which is somehow also a MetricsEndpointConsumer but insists on using a different rel name.

This should be okay (?).

tests/unit/test_scrape_job_filtering.py
sinapah: Yes, I think the assertion that checks the scrape jobs written to the mock FS is probably unnecessary, but I think we probably don't need `_has_scrape_job_errors` tested. The status set is an indication in itself that `_has_scrape_job_errors` has worked correctly.

lib/charms/prometheus_k8s/v0/prometheus_scrape.py
sed-i: Oh I see. I thought we're getting the failure details from cos-tool. In that case there isn't much value in this log line, which is qualitatively equivalent to "config validation failed".

sed-i (approved):

PR #862, opened 2026-09-04
fix!: WAL compression config

### LINKED ISSUE #861: Prometheus WAL compression is always enabled
### Bug Description

Prometheus 2.20.0 (compression [became default](https://prometheus.io/docs/prometheus/latest/storage/#operational-aspects)) 2020-07-22. In the charm code, we add the `storage.tsdb.wal-compression` to the pebble plan based on the Juju config being set to `true`. We need to instead set `--no-storage.tsdb.wal-compression` if the user chooses to disable it. Alternatively, we can remove this Juju config entirely since it hasn't done anything forever.

### To Reproduce

❯ jsshc prometheus prometheus-k8s/0 bash        
root@prometheus-k8s-0:/# pebble plan
services:
    prometheus:
        summary: prometheus daemon
        startup: enabled
        override: replace
        command: /bin/prometheus --config.file=/etc/prometheus/prometheus.yml --storage.tsdb.path=/var/lib/prometheus --web.enable-lifecycle --web.external-url=http://prometheus-k8s-0.prometheus-k8s-endpoints.prom.svc.cluster.local:9090 --web.route-prefix=/ --web.enable-remote-write-receiver --log.level=info --storage.tsdb.retention.time=15d --storage.tsdb.retention.size=0.8GB

jc prometheus-k8s metrics_wal_compression=true

root@prometheus-k8s-0:/# pebble plan
services:
    prometheus:
        summary: prometheus daemon
        startup: enabled
        override: replace
        command: /bin/prometheus --config.file=/etc/prometheus/prometheus.yml --storage.tsdb.path=/var/lib/prometheus --web.enable-lifecycle --web.external-url=http://prometheus-k8s-0.prometheus-k8s-endpoints.prom.svc.cluster.local:9090 --web.route-prefix=/ --web.enable-remote-write-receiver --log.level=info --storage.tsdb.wal-compression --storage.tsdb.retention.time=15d --storage.tsdb.retention.size=0.8GB

### Environment

Prometheus K8s charm in isolation

### Relevant log output

```shell
.
```

### Additional context

_No response_

MichaelThamm:
## Issue

Fixes https://github.com/canonical/prometheus-k8s-operator/issues/861

## Solution

Flip the default value of the `metrics_wal_compression` Juju config to match the upstream default. Since this is now the opposite, users will need to take care to 

## Testing Instructions

- Deploy prom
- `charmcraft pack` from this branch
- check the value of the config

## Upgrade Notes

The config option was

### DIFF
--- a/src/charm.py
+++ b/src/charm.py
@@ -887,6 +887,8 @@ class PrometheusCharm(CharmBase):
 
         if config.get("metrics_wal_compression"):
             args.append("--storage.tsdb.wal-compression")
+        else:
+            args.append("--no-storage.tsdb.wal-compression")
 
         if self._exemplars:
             args.append("--enable-feature=exemplar-storage")
--- a/charmcraft.yaml
+++ b/charmcraft.yaml
@@ -222,11 +222,15 @@ config:
       default: "80%"
     metrics_wal_compression:
       description: |
-        This flag enables compression of the write-ahead log (WAL).
+        Whether to compress the write-ahead log (WAL).
         Depending on your data, you can expect the WAL size to be
         halved with little extra cpu load.
+        Enabled by default, matching the Prometheus workload default. Disabling this roughly
+        doubles the size of the `wal` directory, which counts towards the storage the workload
+        needs but is **not** reclaimed by `maximum_retention_size` (only persistent blocks are).
+        Disabling it therefore increases the risk of exhausting the `database` storage.
       type: boolean
-      default: false
+      default: true
     evaluation_interval:
       description: |
         How frequently rules will be evaluated.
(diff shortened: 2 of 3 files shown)


src/charm.py
```diff
@@ -841,7 +841,11 @@ def _generate_command(self) -> str:
 
         args.append(f"--log.level={self.log_level}")
 
-        if config.get("metrics_wal_compression"):
+        # Keep compression enabled if an old or incomplete config map does not contain this
+        # option. Only an explicit `false` configuration value may disable compression.
```
MichaelThamm:
```suggestion
```

PR #864, opened 2026-09-08
feat(lib): Remote write rule compression

MichaelThamm:
## Issue

We want compression to avoid hitting databag limits.

## Solution

Related:
- https://github.com/canonical/charmlibs/pull/675/
- https://github.com/canonical/opentelemetry-collector-k8s-operator/pull/370
- https://github.com/canonical/mimir-operators/pull/467
- https://github.com/canonical/opentelemetry-collector-operator/pull/403

Negotiated compression, so no combination of lib versions loses alert rules. The Provider app databag advertises what it can read, consumer picks the best match:
```bash
receive-remote-write (provider app data):  alert_rules_encodings: ["lzma", "json"]
send-remote-write    (consumer app data):  alert_rules: <lzma+base64>   # or plain JSON
```
| Consumer | Provider | Result |
|--|--|--|
| new	| new | compressed |
| new | old (advertises nothing) | plain JSON, exactly as today |
| old | new | plain JSON, read fine (reader accepts both) |

Consumer: `_alert_rules_encoding(relation)` inspects the remote app databag; `_push_alerts_to_relation_databag` encodes accordingly (prometheus_remote_write.py:656). Also pushes alerts on relation_changed, so a relation that started out plain upgrades to compressed once the flag lands.

Provider: `_publish_alert_rules_encodings` (leader-guarded) called from `update_endpoint`, `relation_changed`, plus new `leader_elected`/`upgrade_charm` observers so existing relations get the flag after a provider upgrade.

## Context

Provider rollback is the one uncovered skew: a downgraded provider can't erase its own `alert_rules_encodings` key, so the consumer keeps compressing. Recovery is a consumer re-publish after the key is removed manually or a remove integration op.

### TODO
- [x] Test this end-to-end with the packed charms from the PRs
- [x] Consider Sina's logic for alert rule validation -- does this require decompression/compression prior?
	- https://github.com/canonical/prometheus-k8s-operator/pull/817/changes
- [ ] Backport the charm features

## Testing Instructions

With the tandem PRs in Otelcol and Mimir, we should test the compression end-to-end.

### Upgrade Mimir, then Otelcol
```bash
jd opentelemetry-collector-k8s --trust --channel "dev/edge" otelcol
atelier module add https://github.com/canonical/mimir-operators.git  # dev/edge
jrel otelcol mimir:receive-remote-write
jshu mimir/0 | yq '.[]."relation-info"[] | select(.endpoint == "receive-remote-write").application-data.alert_rules' | jq
# {"groups": [{...}]}
juju refresh mimir --path ./mimir-coordinator-k8s_ubuntu@26.04-amd64.charm \
  --resource nginx-image=ghcr.io/canonical/nginx:1.27.5 --resource nginx-prometheus-exporter-image=ubuntu/nginx-prometheus-exporter@sha256:28fc4e2e0a8e32d08f213889ac5eb613d52f64da47ae5007930adfc688fbbe5c
jshu mimir/0 | yq '.[]."relation-info"[] | select(.endpoint == "receive-remote-write").application-data.alert_rules' | jq
# {"groups": [{...}]}
juju refresh otelcol --path ./opentelemetry-collector-k8s_ubuntu@26.04-amd64.charm \
  --resource opentelemetry-collector-image=ubuntu/opentelemetry-collector@sha256:6df3cad7e93ef008014e85caf66765a69859f4507f1356fdb968f17e68b5766c
jshu mimir/0 | yq '.[]."relation-info"[] | select(.endpoint == "receive-remote-write").application-data'                 
# alert_rules: /Td6WFoAAAT...
```
```bash
❯ jst
Model  Controller  Cloud/Region  Version  SLA          Timestamp
prw    k8s         k8s           3.6.23   unsupported  16:01:29-04:00

App                  Version  Status  Scale  Charm                        Channel   Rev  Address         Exposed  Message
mimir                         active      1  mimir-coordinator-k8s                    1  10.152.183.18   no       
mimir-backend        2.17.10  active      1  mimir-worker-k8s             dev/edge  108  10.152.183.53   no       backend ready.
mimir-read           2.17.10  active      1  mimir-worker-k8s             dev/edge  108  10.152.183.143  no       read ready.
mimir-s3-integrator           active      1  s3-integrator                2/stable  544  10.152.183.139  no       
mimir-write          2.17.10  active      1  mimir-worker-k8s             dev/edge  108  10.152.183.90   no       write ready.
otelcol              0.130.1  active      1  opentelemetry-collector-k8s              0  10.152.183.114  no       

Unit                    Workload  Agent  Address     Ports  Message
mimir-backend/0*        active    idle   10.1.0.240         backend ready.
mimir-read/0*           active    idle   10.1.0.58          read ready.
mimir-s3-integrator/0*  active    idle   10.1.0.80          
mimir-write/0*          active    idle   10.1.0.10          write ready.
mimir/0*                active    idle   10.1.0.225         
otelcol/0*              active    idle   10.1.0.126         

Integration provider                Requirer                          Interface                Type     Message
mimir-s3-integrator:s3-credentials  mimir:s3                          s3                       regular  
mimir-s3-integrator:status-peers    mimir-s3-integrator:status-peers  status_peers             peer     
mimir:mimir-cluster                 mimir-backend:mimir-cluster       mimir_cluster            regular  
mimir:mimir-cluster                 mimir-read:mimir-cluster          mimir_cluster            regular  
mimir:mimir-cluster                 mimir-write:mimir-cluster         mimir_cluster            regular  
mimir:mimir-peers                   mimir:mimir-peers                 mimir_peers              peer     
mimir:receive-remote-write          otelcol:send-remote-write         prometheus_remote_write  regular  
otelcol:peers                       otelcol:peers                     otelcol_replica          peer
```

### Upgrade Otelcol, then Mimir

```bash
# same steps as above, deploying Otelcol and Mimir
juju refresh otelcol --path ./opentelemetry-collector-k8s_ubuntu@26.04-amd64.charm \
  --resource opentelemetry-collector-image=ubuntu/opentelemetry-collector@sha256:6df3cad7e93ef008014e85caf66765a69859f4507f1356fdb968f17e68b5766c
jshu mimir/0 | yq '.[]."relation-info"[] | select(.endpoint == "receive-remote-write").application-data.alert_rules' | jq
# {"groups": [{...}]}
juju refresh mimir --path ./mimir-coordinator-k8s_ubuntu@26.04-amd64.charm \
  --resource nginx-image=ghcr.io/canonical/nginx:1.27.5 --resource nginx-prometheus-exporter-image=ubuntu/nginx-prometheus-exporter@sha256:28fc4e2e0a8e32d08f213889ac5eb613d52f64da47ae5007930adfc688fbbe5c
jshu mimir/0 | yq '.[]."relation-info"[] | select(.endpoint == "receive-remote-write").application-data'                 
# alert_rules: /Td6WFoAAAT...
```


lib/charms/prometheus_k8s/v1/prometheus_remote_write.py
MichaelThamm, line 1: Potentially worth it to move `ALERT_RULES_KEY`, `SUPPORTED_ALERT_RULES_ENCODINGS`, etc. to a dataclass or Pydantic model if we wanted to be precise; or not since the OTLP lib succeeds this.

lib/charms/prometheus_k8s/v1/prometheus_remote_write.py
```diff
@@ -44,7 +62,7 @@
 
 # Increment this PATCH version before using `charmcraft publish-lib` or reset
 # to 0 if you are raising the major API version
-LIBPATCH = 18
+LIBPATCH = 19
```
sed-i, line 66: If we do not increment LIBAPI, then this change would spread everywhere automatically via CI.

Should be bump LIBAPI to "limit" this behaviour to otelcol to mimir/prom?

I forget if intermediate charms like otelcol we do any calculations/modifications on alert rules. If yes, it would save us on decompressing all the incoming data.

Wdyt?

lib/charms/prometheus_k8s/v1/prometheus_remote_write.py
```diff
@@ -58,6 +76,72 @@
 
 DEFAULT_ALERT_RULES_RELATIVE_PATH = "./src/prometheus_alert_rules"
 
+ALERT_RULES_KEY = "alert_rules"
```
sed-i: Should we type-hint with [Final](https://typing.python.org/en/latest/spec/qualifiers.html#uppercase-final)?

lib/charms/prometheus_k8s/v1/prometheus_remote_write.py
```diff
@@ -459,7 +547,10 @@ def __init__(
         self.framework.observe(on_relation.relation_changed, self._handle_endpoints_changed)
         self.framework.observe(on_relation.relation_departed, self._handle_endpoints_changed)
         self.framework.observe(on_relation.relation_broken, self._on_relation_broken)
-        self.framework.observe(on_relation.relation_joined, self._push_alerts_on_relation_joined)
+        self.framework.observe(on_relation.relation_joined, self._push_alerts_on_relation_event)
+        # The provider advertises the alert rules encodings it supports over relation data,
+        # so alerts are (re)pushed on relation-changed to pick up the negotiated encoding.
+        self.framework.observe(on_relation.relation_changed, self._push_alerts_on_relation_event)
```
sed-i, line 617: Could you please explain this change?
Is the original code an unnoticed bug?
I.e why do we need to `_push_alerts_on_relation_event` on both joined and changed?

lib/charms/prometheus_k8s/v1/prometheus_remote_write.py
```diff
@@ -490,13 +581,52 @@ def _handle_endpoints_changed(self, event: RelationEvent) -> None:
...
+            relation: the relation whose remote application databag to inspect.
+
+        Returns:
+            One of `SUPPORTED_ALERT_RULES_ENCODINGS`.
+        """
+        if not relation.app:
+            return JSON_ENCODING
+
+        if (remote_databag := relation.data.get(relation.app)) is None:
+            return JSON_ENCODING
```
sed-i: Should this func take as arg `relation.data.get(relation.app)` instead of Relation? That way you won't need to deal with edge cases related to lifecycle I imagine.

lib/charms/prometheus_k8s/v1/prometheus_remote_write.py
```diff
@@ -490,13 +581,52 @@ def _handle_endpoints_changed(self, event: RelationEvent) -> None:
...
+
+        try:
+            advertised = json.loads(remote_databag.get(ALERT_RULES_ENCODINGS_KEY, "[]"))
+        except json.JSONDecodeError:
+            logger.warning(
+                "Could not parse the '%s' advertised over relation %s; "
+                "falling back to plain JSON alert rules.",
```
sed-i: The exception and the warning message sound contradicting from first glance - if we failed to json.loads due to JSONDecodeError, why are we returning JSON_ENCODING?

lib/charms/prometheus_k8s/v1/prometheus_remote_write.py
```diff
@@ -766,8 +909,31 @@ def _on_consumers_changed(self, event: RelationEvent) -> None:
...
+
+        Args:
+            relation: The relation whose data to update.
+        """
+        if not self._charm.unit.is_leader():
+            # Only the leader unit can write to app data.
+            return
+
+        relation.data[self._charm.app][ALERT_RULES_ENCODINGS_KEY] = json.dumps(
+            list(SUPPORTED_ALERT_RULES_ENCODINGS)
```
sed-i: flaplint?

lib/charms/prometheus_k8s/v1/prometheus_remote_write.py
```diff
@@ -835,7 +1005,20 @@ def alerts(self) -> dict:
...
 
-            alert_rules = json.loads(relation.data[relation.app].get("alert_rules", "{}"))
+            try:
+                alert_rules = _decode_alert_rules(
+                    relation.data[relation.app].get(ALERT_RULES_KEY, "{}")
+                )
+            except Exception as e:
+                # Never let unreadable remote data break the provider: a consumer could
+                # be writing rules in a format this version of the library predates.
+                logger.error(
```
sed-i: Should we log the error outside the for-loop to avoid log-spamming?

tests/unit/test_remote_write_compression.py
sed-i, line 1: Great coverage!

sed-i (commented):
> Recovery is a consumer re-publish after the key is removed manually or a remove integration op.

Should we document in the PR description the manual call to relation-set that fixes a downgrade?

lib/charms/prometheus_k8s/v1/prometheus_remote_write.py
mmkay: I think that we want this to spread - especially that we ensure backwards compatibility by first trying to read json, then try to uncompress.

MichaelThamm: I agree with Mateusz here

lib/charms/prometheus_k8s/v1/prometheus_remote_write.py
MichaelThamm: Agreed, I will update this to log the list of relations, rather than a log per relation

tests/unit/test_remote_write_compression.py
MichaelThamm: We can probably reuse the `prometheus_charm` fixture rather than building new ones.

lib/charms/prometheus_k8s/v1/prometheus_remote_write.py
MichaelThamm: At relation-joined time the provider's app databag is typically still empty, so the encoding it can read is only known once it writes it, i.e. on relation-changed. Re-pushing on relation-changed is what lets an existing, plain-JSON relation switch over to compressed rules. Writes are idempotent: if the encoding and the rules are unchanged, the databag value is identical and juju emits no further event.

lib/charms/prometheus_k8s/v1/prometheus_remote_write.py
MichaelThamm: This list is always the same; no need to flaplint here

lib/charms/prometheus_k8s/v1/prometheus_remote_write.py
```diff
@@ -58,6 +76,119 @@
...
+    try:
+        decoded = json.loads(raw)
+    except json.JSONDecodeError:
+        # Not JSON, so this must be a compressed payload.
+        decoded = raw
+
+    if isinstance(decoded, str):
+        # A compressed payload, either bare or (as pydantic based libraries write it)
+        # JSON-encoded.
+        decoded = json.loads(LZMABase64.decompress(decoded))
```
Abuelodelanada: One comment about this function:

`json.loads('"foo"')` returns `'foo'` (str), and because this is a string, it tries to execute `LZMABase64.decompress("foo")` which fails.


What do you think about this?

```suggestion
     if isinstance(decoded, str):
         # A compressed payload, either bare or (as pydantic based libraries write it)
         # JSON-encoded.
         try:
             decoded = json.loads(LZMABase64.decompress(decoded))
         except (ValueError, lzma.LZMAError) as e:
             raise ValueError(f"Could not decompress alert rules: {e}") from e
 
     if not isinstance(decoded, dict):
         raise ValueError(
             f"Alert rules must be a JSON object, not {type(decoded).__name__}"
         )
```

sinapah (approved):

Abuelodelanada (approved):

PR #871, opened 2026-09-22
fix: Scrape jobs behind ingress

### LINKED ISSUE #870: Prometheus publishes its external_url incorrectly for scrape jobs
### Bug Description

Given the most recent feature of otelcol in COS Lite:
- https://github.com/canonical/observability-stack/pull/492

we now use otelcol to scrape Prometheus, but Prometheus is publishing its target incorrectly when it has ingress and that ingress is not serving TLS. The problem is that Prometheus determines what target to set based on its own TLS config, not the TLS config of the ingress that is infront of it. For example:
- `curl -k 10.212.157.241/cos-lite-prometheus-0/metrics`  # works
- `curl -k 10.212.157.241:443/cos-lite-prometheus-0/metrics`  # fails

for the scrape config in otelcol:
```yaml
      - job_name: juju_cos-lite_f1278dbd_prometheus_prometheus_scrape-0
        metrics_path: /cos-lite-prometheus-0/metrics
        scheme: https
        static_configs:
          - targets:
              - 10.212.157.241:443
        tls_config:
          ca_file: /etc/otelcol/certs/otel_juju_cos_lite_f1278dbd_prometheus_prometheus_scrape_0_ca.pem
          insecure_skip_verify: false
```

specifically, [this logic](https://github.com/canonical/prometheus-k8s-operator/blob/4a38ca7c1c03a79a4a73f4074822f39f402c15ad/src/charm.py#L400) is what bites us, since it checks Prometheus' own TLS config to determine what port to set in the target, but it should instead check Traefik's `scheme` from the ingress requirer.

### To Reproduce

1. `atelier module add https://github.com/canonical/observability-stack.git --module terraform/cos-lite`  # plan and apply
2. `jsshc otelcol otelcol/leader pebble logs -f `  # to see the warning logs for failed to scrape
3. `jsshc otelcol otelcol/leader cat /etc/otelcol/config.yaml | yq '.receivers'` to see the scrape job for Prometheus

### Environment

COS Lite with ingress

### Relevant log output

```shell
2026-09-22T18:06:17.110Z [otelcol] 2026-09-22T18:06:17.110Z	warn	internal/transaction.go:150	Failed to scrape Prometheus endpoint	{"resource": {"juju_application": "otelcol", "juju_charm": "opentelemetry-collector-k8s", "juju_model": "cos-lite", "juju_model_uuid": "f1278dbd-11c8-4561-8ff5-cd7328437985", "juju_unit": "otelcol/0", "loki.format": "logfmt", "loki.resource.labels": "juju_application, juju_charm, juju_model, juju_model_uuid, juju_unit", "service.instance.id": "otelcol/0", "service.name": "otelcol-internal", "service.version": "0.130.1"}, "otelcol.component.id": "prometheus/metrics-endpoint/otelcol/0", "otelcol.component.kind": "receiver", "otelcol.signal": "metrics", "scrape_timestamp": 1790100377107, "target_labels": "{__name__=\"up\", instance=\"cos-lite_f1278dbd-11c8-4561-8ff5-cd7328437985_prometheus_prometheus/0\", job=\"juju_cos-lite_f1278dbd_prometheus_prometheus_scrape-0\", juju_application=\"prometheus\", juju_charm=\"prometheus-k8s\", juju_model=\"cos-lite\", juju_model_uuid=\"f1278dbd-11c8-4561-8ff5-cd7328437985\", juju_unit=\"prometheus/0\"}"}
```

### Additional context

_No response_

MichaelThamm:
## Issue

Fixes https://github.com/canonical/prometheus-k8s-operator/issues/870. When ingressed, Prometheus advertised its scraper's target based on its own TLS config and the ingress URL, so otelcol got the wrong scheme/port (e.g. `https://<traefik-ip>:443`) and self-monitoring scrapes failed whenever the traffic path went through Traefik.

## Solution

Prometheus now advertises an in-cluster scrape target (its own FQDN, workload port, and own scheme) instead of the ingress URL, matching sibling charms (alertmanager, loki, grafana).
- `self-scraping-job` targets `<fqdn>:9090`, with scheme/tls config derived from Prometheus' own certs — correct in all four ingress/TLS combinations.
- No `external_url` is published over `self-metrics-endpoint`; the databag no longer routes scrapers through Traefik.
- The ingress URL is still used for the UI (`--web.external-url`).

## Testing Instructions

Parametrized unit test covering {no ingress, http, https} × {prom http, https}, asserting the target, scheme, `tls_config`, and that the ingress URL affects only the UI. Fails on `main`; all unit tests pass.


src/charm.py
```diff
@@ -225,14 +225,12 @@ def __init__(self, *args):
             relation_name="alertmanager",
         )
 
+        # No `external_url`: self-monitoring is in-cluster, not through the ingress.
```
sinapah: For context, this statement is also supported by https://github.com/canonical/cos-coordinated-workers/issues/168#issuecomment-4841170708

tests/unit/test_self_scrape_job.py
sinapah: FWIW, this test seems an excellent candidate for feature files 😆

sinapah (approved):
Should we also add an integration test that asserts metrics correctly reach Prometheus across all TLS/ingress scenarios when using Otelcol?
https://github.com/canonical/grafana-k8s-operator/blob/main/tests/integration/test_ingress.py

MichaelThamm:
From a discussion with the team, we will continue to publish the ingress address and fix the port provided based on the ingress scheme. When we have the feature to check in-model vs. cross-model, then we will be able to conditionally switch to scraping Prom's pod FQDN within COS.

